PR #22, opened 2022-08-26
Ctw 313 embeddable boost z index value of zus drawer in healthie

melocule:
Deleted image as this repo will be open source (Luthfi)

### DIFF
--- a/src/components/core/drawer.tsx
+++ b/src/components/core/drawer.tsx
@@ -40,7 +40,7 @@ export function Drawer({
           leaveTo="ctw-opacity-0"
           afterLeave={onAfterClosed}
         >
-          <div className="ctw-fixed ctw-inset-0  ctw-transition-opacity ctw-z-10000">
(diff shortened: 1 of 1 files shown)


pmadria3 (dismissed):
LGTM

melocule:
Thanks for catching this! Added the prefix

luthfib (changes requested):
`ctw-z-[10000]` You are also missing the brackets. Since tailwind doesnt natively have a 10000 value, you have to use the custom modifier which is the brackets

luthfib (changes requested):
Actually I just tested this locally and this does not work. You might want to change your approach and verify that it is working. I would suggest mocking an element that has a higher z-index and ensuring that the drawer still shows up above it. Also make sure you are editing the right div in the drawer component

melocule:
Found the div that z-index is supposed to apply to, tested it against a black box with a z-index set to 9999 (which shows up on top of the drawer when its index is set to 10,001)
image

luthfib (approved):
LGTM

PR #36, opened 2022-08-29
Ctw 320/pass session info

luthfib:
- Allow headers to be passed into component

### DIFF
--- a/src/components/core/ctw-provider.tsx
+++ b/src/components/core/ctw-provider.tsx
@@ -94,7 +94,6 @@ async function checkOrRefreshAuth(
   headers: HeadersInit
 ): Promise<CTWToken> {
   if (!token || Date.now() >= token.expiresAt + EXPIRY_PADDING_MS) {
-    console.log("headers here", headers);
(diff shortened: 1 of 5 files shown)


src/components/core/ctw-provider.tsx
```diff
@@ -10,26 +11,28 @@ type CTWToken = {
...
-type AuthTokenURLSpecified = { authToken?: never; authTokenURL: string; };
+type AuthTokenURLSpecified = { authToken?: never; authTokenURL: string };
 
-type CTWProviderProps = { 
-  children: React.ReactNode,
+type CTWProviderProps = {
+  children: React.ReactNode;
   env: Env;
   theme?: any;
+  headers?: HeadersInit;
```
pmadria3, line 35: why is this optional? We should probably make it required

luthfib: I think it should be optional because it wont be required in all instances

pmadria3 (approved):
LGTM

PR #38, opened 2022-08-30
CTW312/transition-from-UPID-to-HealthieID

pmadria3:
This PR utilizes either HealthieID or UPID and allows to be able to obtain information by using a patientProvider for conditions and in the future other resources.

Screen Shot 2022-09-01 at 4 04 58 PM


src/components/content/conditions.tsx
```diff
@@ -40,18 +41,20 @@ export function Conditions({ className, patientUPID }: ConditionsProps) {
           <div className="ctw-space-y-3">
             <div className="ctw-title ctw-ml-3">Confirmed</div>
             <ConditionsTable
-              patientUPID={patientUPID}
+              patientID={patientID}
               isConfirmed
               includeInactive={includeInactive}
+              system={system}
```
luthfib, line 47: Do we use system here?  I dont see where it is used, but correct me if im wrong

src/fhir/search-helpers.ts
```diff
@@ -91,17 +91,17 @@ export async function searchLensRecords<T extends ResourceTypeString>(
...
 ): SearchParams {
-  // No search param needed when not searching for a patientUPID.
-  if (!patientUPID) {
+  // No search param needed when not searching for a patientID.
+  if (!patientID) {
     return {};
   }
 
-  const identifier = `${SYSTEM_ZUS_UNIVERSAL_ID}|${patientUPID}`;
+  const identifier = `${SYSTEM_HEALTHIE_ID}|${patientID}`;
```
luthfib, line 104: Make this be able to pass in dont hardcode this in here

luthfib (changes requested):

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+type IDProviderProps = {
+  children: React.ReactNode;
+} & ThirdPartyID;
+
+export const CTWPatientContext = React.createContext<CTWID>({
+  patientID: "",
+  systemURL: "",
+});
+
+export function IDProvider({ children, ...ctwState }: IDProviderProps) {
```
luthfib, line 19: Please rename this to match the rest of the file

luthfib (dismissed):

src/fhir/conditions.ts
```diff
@@ -39,15 +40,17 @@ export async function getConfirmedConditions(
...
   conditionFilters: ConditionFilters = {}
 ) {
   try {
     const { resources: conditions } = await searchLensRecords(
       "Condition",
       fhirClient,
       {
-        patientUPID,
+        patientID,
+        systemURL: systemZus,
```
luthfib, line 53: same here? Not sure why we have to rename it if we are just passing it down, maybe rename the arg?

src/components/content/conditions.tsx
```diff
@@ -56,12 +59,30 @@ export function Conditions({ className, patientUPID }: ConditionsProps) {
...
 
       const fhirClient = await getCTWFhirClient();
+      const patientFilter = {};
+
+      const { patientUPID, systemZus } = await getUPIDfromPatientID(
+        fhirClient,
+        patientID,
+        systemURL,
+        patientFilter
+      );
```
luthfib, line 69: It would be nice to do this in the provider, instead of having to do it in every component that we need it in. Im thinking something along the lines of.... 

` const { patientID, systemURL, patientUPID, system, patientDataIsLoading } =
    usePatientContext();`

Then we can trigger of a patientDataIsLoading
`
 if (!patientDataIsLoading) {
      load();
    }
`
and re trigger the useEffect with by adding `patientDataIsLoading` to the dep array. 

Inside of your provider code you could do something like....

`
useEffect(() => {
    async function load() {
      const fhirClient = await getCTWFhirClient();
      const response = await getUPIDfromPatientID(
        fhirClient,
        ctwState.patientID,
        ctwState.systemURL
      );
      setUPIDData(response);
      setIsLoading(false);
    }
    load();
  }, []);
`

This is just a potential nice to have, doesnt necessarily have to stop the pr and this isn't fully tested code or anything, but I think it would be nice because then we dont have calls everywhere translating patientID's to UPID

luthfib (changes requested):

src/components/content/conditions.tsx
pmadria3: Confer with @WesleyKapow on this

luthfib (dismissed):
Also dont forget to add a changeset.... `npx changeset`. Add a meaningful message

src/components/content/conditions.tsx
WesleyKapow: Agreed. This should be handled by the `PatientProvider`.

Really the only thing `PatientProvider` needs to return to us is a way to get a UPID.

`const { getPatientUPID } = usePatientContext()`

This conditions component would then do `const patientUPID = await getPatientUPID()`. And the provider would either 1. quickly return the UPID it already fetched, or 2. fetch the UPID.

Loading state can be handled by the conditions component. Also no need for `systemURL` as this is a UPID and we can use the hardcoded system URL constant for that.

WesleyKapow: This will also open up the door for us in the future to make `PatientProvider` take either:
A. patientID & systemURL
B. patientUPID

So in the CTW app case, we'd already have a UPID from the URL, so we would do <PatientProvider patientUPID={valueFromURLParams}>` and then `getPatientUPID()` would return that one instead of ever needing to fetch one.

WesleyKapow (changes requested):

src/components/content/conditions.tsx
```diff
@@ -45,6 +46,8 @@ export function Conditions({ className, patientUPID }: ConditionsProps) {
   const [includeInactive, setIncludeInactive] = useState(true);
   const { getCTWFhirClient } = useCTW();
 
+  const { getPatientUPID, patientID, systemURL } = usePatient();
```
WesleyKapow, line 49: Why do we need `patientID` and `systemURL`? Can we not:
1. Do everything off of UPID.
2. Pass UPID around and use `SYSTEM_ZUS_UNIVERSAL_ID` when constructing the query.

src/fhir/search-helpers.ts
```diff
@@ -90,34 +125,31 @@ export async function searchLensRecords<T extends ResourceTypeString>(
...
-): SearchParams {
-  // No search param needed when not searching for a patientUPID.
-  if (!patientUPID) {
-    return {};
-  }
+export async function getUPIDfromPatientID(
+  fhirClient: Client,
+  patientID: string,
+  systemURL: string,
+  patientFilters?: SearchParams
```
WesleyKapow, line 132: Do we need `patientFilters` here?

src/fhir/search-helpers.ts
```diff
@@ -27,24 +28,58 @@ export type SearchReturn<T extends ResourceTypeString> = {
...
     searchParams: {
       ...params,
-      ...patientSearchParams(resourceType, patientUPID as string),
+      ...patientSearchParams(
+        resourceType,
+        patientID as string,
+        systemURL as string
+      ),
     },
   })) as fhir4.Bundle;
```
WesleyKapow, line 84: I think you can change this back to be based on `patientUPID` and not `patientID/systemURL`. My hope is we can keep things simple by ALWAYS fetching data by `patientUPID` EXCEPT when calling `getUPIDfromPatientID` which should NOT use the other search helpers. Instead it should search directly using the fhir client and find the first patient that has the matching system/patientID.

WesleyKapow (changes requested):

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,50 @@
...
+type ThirdPartyID = { patientID: string; systemURL: string };
+
+type PatientProviderProps = {
+  children: React.ReactNode;
+} & ThirdPartyID;
+
+export const CTWPatientContext = React.createContext<ThirdPartyID>({
+  patientID: "",
+  systemURL: "",
+});
```
WesleyKapow, line 16: Construct initial state with `undefined` like I did in CTWProvider.

```ts
React.createContect<ThirdPartyID | undefined>(undefined)
```

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { useCTW } from "./ctw-provider";
+
+import { getUPIDfromPatientID } from "@/fhir/search-helpers";
+
+type ThirdPartyID = { patientID: string; systemURL: string };
+
+type PatientProviderProps = {
+  children: React.ReactNode;
+} & ThirdPartyID;
```
WesleyKapow, line 11: So... we're gonna want this to work for EITHER:
A. Consuming app specifies `patientID` and `systemURL` OR
B. Consuming app specifies `patientUPID`

Look at CTWProvider for how this is setup. You'll want a state that has things as optional and a props that has `& (ThirdPartyID | {patientUPID})`

& (ThirdPartySpecified | PatientUPIDSpecified)

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,50 @@
...
+      {children}
+    </CTWPatientContext.Provider>
+  );
+}
+
+export function usePatient() {
+  const context = useContext(CTWPatientContext);
+  const { getCTWFhirClient } = useCTW();
+
+  async function getPatientUPID() {
```
WesleyKapow, line 40: Make this function work for the case where `patientUPID` is passed in as a prop.. i.e check if we have `patientUPID` and if so, return it.

Also,  look at CTWProvider's `handleAuth`. What we want is for only the first call to this to have to fetch the UPID. And actually, we know we have to fetch it.. maybe go ahead and fetch it write away in `PatientProvider`, store in some `useState` stae. You'll want to store the promise of the fetch so that consumers of `getPatientUPID()` can still `await` until it is done. But this should simplify things so that you immediately try to get the UPID and cache it in the state.

WesleyKapow (dismissed):

src/fhir/search-helpers.ts
```diff
@@ -27,24 +28,54 @@ export type SearchReturn<T extends ResourceTypeString> = {
   resources: ResourceType<T>[];
 };
 
+// Returns the needed search params to filter resources down to those
+// pertaining to our patientID and system url.
+function patientSearchParams(
```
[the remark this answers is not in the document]
pmadria3: I agree - Eslint was complaining so moved it to stop but then just reverted it back to original

src/fhir/search-helpers.ts
```diff
@@ -33,8 +33,8 @@ export type SearchReturn<T extends ResourceTypeString> = {
 // Returns {bundle: fhir4.Bundle, total: number, resources: ResourceType[]}.
 // Usage: {bundle, total, resources: tasks} =
 //          searchAllRecords("Task", fhirClient, searchParams)
-// NOTE: "patientUPID" is a special searchParam that will correctly
-//       filter down to the resources pertaining to that patient UPID.
+// NOTE: "patientID" is a special searchParam that will correctly
+//       filter down to the resources pertaining to that patient ID.
 export async function searchAllRecords<T extends ResourceTypeString>(
```
WesleyKapow, line 38: Please look at the diff you made to this function and the comments. Probably want to revert them?

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { useContext, useEffect, useState } from "react";
```
WesleyKapow, line 1: Remove uses of `React.*` with the values directly.. so `ReactNode`, `createContext`, `useMemo`.

That way it is clear what things from `React` this file depends on.

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useCTW } from "./ctw-provider";
+
+import { getUPIDfromPatientID } from "@/fhir/search-helpers";
+
+type ThirdPartyID = {
+  patientUPID?: never;
+  patientID: string;
+  systemURL: string;
+};
+type CtwUPID = { patientUPID: string; patientID?: never; systemURL?: never };
```
WesleyKapow, line 12: CtwUPID? How about `PatientUPIDSpecified` (aligns with naming used in CTWProvider).

src/components/core/patient-provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <CTWPatientContext.Provider value={providerState}>
+      {children}
+    </CTWPatientContext.Provider>
+  );
+}
+
+export function usePatient() {
+  const { patientUPIDPromise } = useContext(CTWPatientContext);
+  return { patientUPIDPromise };
```
WesleyKapow, line 66: Hmmm can be simplified to just `return useContext(CTWPatientContext)`.

WesleyKapow (changes requested):

src/fhir/search-helpers.ts
```diff
@@ -53,18 +53,6 @@ export async function searchAllRecords<T extends ResourceTypeString>(
...
-  resourceType: T,
-  fhirClient: Client,
-  searchParams?: SearchParams
-): Promise<SearchReturn<T>> {
-  return searchAllRecords(resourceType, fhirClient, {
-    ...searchParams,
-    "_tag:not": LENS_TAGS.join(","),
-  });
-}
-
```
WesleyKapow, line 67: Why remove this? It may be un-used a.t.m but I expect we'll want it.

WesleyKapow (changes requested):

.changeset/ninety-dolphins-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+There's a new PatientProvider that allows the user to provide either a PatientUPID or a combination of PatientID/SystemUrl to be utilized for fetching conditions about the patient.
```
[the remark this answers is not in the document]
pmadria3: I left the conditions in there but added a note that we will utilize it for future resources

.changeset/ninety-dolphins-shout.md
```diff
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-There's a new PatientProvider that allows the user to provide either a PatientUPID or a combination of PatientID/SystemUrl to be utilized for fetching conditions about the patient.
+There's a new PatientProvider that allows the user to provide either a PatientUPID or a combination of PatientID/SystemUrl to be utilized for fetching conditions about the patient. Will be used to fetch other resources as well as they get built out.
```
WesleyKapow, line 5: The confusion here is that it has nothing to do with fetching conditions or other resources. All it does is provide a UPID, possibly by looking it up. What other components do with that UPID is their business.

WesleyKapow (approved):

.changeset/ninety-dolphins-shout.md
pmadria3: I see - went ahead and fixed that to be specific to patientProvider.

WesleyKapow (approved):

PR #39, opened 2022-08-30
Ctw 314/filter builder managed conditions from not reviewed

luthfib:
- Removes confirmed conditions from not reviewed conditions by using ICD10 as lookup between each. 
- Only lingering question is whether condition needs to do a reference lookup... I dont think we do because I think it always has a code I think with medications sometimes it was in the contained but I dont think this is the case here, but please correct me if im wrong. 
- Fixes random eslint warnings


pmadria3 (approved):
LGTM

PR #44, opened 2022-09-02
CTW-319 embeddable add ability to pass list of styles to themes

melocule:
CTW-319
Added `Theme` type, which sets our CSS variables to new values. No need to sync.

### Test
image
Pasted Graphic 2

Without the style set:
Pasted Graphic 4


src/components/core/ctw-provider.tsx
```diff
@@ -67,10 +69,13 @@ function CTWProvider({ children, ...ctwState }: CTWProviderProps) {
     [ctwState, handleAuth, token]
   );
 
+  // Casts Style as CSSProperties so that it can be passed as a style.
   return (
-    <CTWStateContext.Provider value={providerState}>
-      {children}
-    </CTWStateContext.Provider>
+    <div style={style as React.CSSProperties}>
```
WesleyKapow, line 74: Exported style should already be cast to what is needed. That way any consumer importing it already has the right type (and so multiple consumers won't have to cast it each time).

WesleyKapow (changes requested):
I've got a few concerns with this current approach:
1. Generated files should clearly be generated. In this case the file exists within src/styles and should have a big fat comment at the top saying DO NOT EDIT, THIS IS GENERATED, and a pointer for how to re-generate it.
2. `--ctw--primary-light` and the like are kinda ugly theme names. We don't need the prefix for these when passing it into the provider. Instead, we should have `primaryLight`.
3. I think we can tweak how `CSSColorPalette` is setup to avoid the need for the `sync-style` script. How about this, change the object in `tailwind.config.cjs` to be a flat map of colors.

```ts
export const CSSColorPalette = {
  colors: {
    transparent: "transparent",
    white: "#fff",
    black: "#000",
    primaryLighter: "#FAF5FF",
    primaryLight: "#F3E8FF",
    primaryMain: "#4990E2",
    primaryDark: "#577FE6",
    iconDefault: "#374151",
  },
};
```

Then our type can be:
```ts
export type Theme = Partial<Record<keyof typeof CSSColorPalette['colors'], string>>
```

That should yield a type called `Theme` that looks like:

```ts
type Theme = {
    transparent?: string | undefined;
    white?: string | undefined;
    black?: string | undefined;
    primaryLighter?: string | undefined;
    ...
}
```

You'll then need to at runtime process the passed in object to construct the correct `--ctw--` CSS variables. You can use `kebabCase` from lodash to help here -- https://lodash.com/docs/4.17.15#kebabCase

melocule:
Fresh attempt to import the Theme object from tailwind.config.cjs:
Still no live type checking, the .cjs file requires the .d.ts to declare it as a module. With or without specifying the type of the Theme object in the .d.ts file, live type checking doesn't work in App.tsx. It would work if the type were set literally in the .d.ts file, but that leads to the syncing issue again.

src/components/core/ctw-provider.tsx
```diff
@@ -35,15 +31,20 @@ type AuthTokenURLSpecified = { authToken?: never; authTokenURL: string };
 type CTWProviderProps = {
   children: React.ReactNode;
   env: Env;
-  theme?: Theme;
+  style?: Style;
+  theme?: any;
```
WesleyKapow: Why `theme` and `style`? Let's go with just `theme` and change the type to be called `Theme` as well.

tailwind.config.cjs
```diff
@@ -1,43 +1,9 @@
...
-  for (let i = 0; i < config.length; i++) {
-    Object.assign(flattenedColors, config[i]);
-  }
-
-  return { colors: flattenedColors };
-};
+const TailwindTheme = require("./tailwind-gen.theme.cjs").TailwindTheme;
+const CLASS_PREFIX = require("./tailwind-gen.theme.cjs").CLASS_PREFIX;
+const addCSSVarReference =
+  require("./tailwind-gen.theme.cjs").addCSSVarReference;
```
WesleyKapow: Require the file once and use destructuring to get the three things you want from it.

src/styles/tailwind.theme.ts
```diff
@@ -0,0 +1,128 @@
...
+function nameCSSVar(name: string): string {
+  return `--${CLASS_PREFIX}-${name}`;
+}
+
+function createCSSVar(name: string, defaultVal: string): string {
+  return `var(${nameCSSVar(name)}, ${defaultVal})`;
+}
+
+// Takes a theme and turns it into a CSSProperties that sets CSS Variable values
+export function mapToCSSVar(colorConfig: Style): any {
```
WesleyKapow, line 67: Anyway to DRY this function up with the `addCSSVarReference` function?

src/styles/tailwind.theme.ts
```diff
@@ -0,0 +1,128 @@
...
+      black: "#111827",
+    },
+  },
+};
+
+type Subset<K> = {
+  [attr in keyof K]?: K[attr] extends object ? Subset<K[attr]> : K[attr];
+};
+
+export type Style = Subset<typeof TailwindTheme["colors"]>;
```
WesleyKapow: This is awesome! Really really awesome!

.gitignore
```diff
@@ -1,9 +1,8 @@
...
 src/styles/tailwind-gen.css
+tailwind-gen.theme.cjs
 scripts/prefixer/tmp/
 node_modules
 .env
-.DS_Store
-tailwind.theme.js
-tailwind.theme.cjs
\ No newline at end of file
+.DS_Store
```
WesleyKapow: The `mv` command shouldn't fail but just in case, add the `tailwind.theme.js` file as well.

WesleyKapow (changes requested):
Lookin' REALLY good!!

tailwind.config.cjs
```diff
@@ -1,11 +1,9 @@
 const lineClampPlugin = require("@tailwindcss/line-clamp");
 const defaultTheme = require("tailwindcss/defaultTheme");
-const TailwindTheme = require("./tailwind-gen.theme.cjs").TailwindTheme;
-const CLASS_PREFIX = require("./tailwind-gen.theme.cjs").CLASS_PREFIX;
-const nameCSSVar = require("./tailwind-gen.theme.cjs").nameCSSVar;
+const ThemeUtil = require("./tailwind-gen.theme.cjs");
```
WesleyKapow, line 3: Better to use destructuring to make it clear what things you are using from the imported file.

src/styles/tailwind.theme.ts
```diff
@@ -0,0 +1,83 @@
...
+  [attr in keyof K]?: K[attr] extends object ? Subset<K[attr]> : K[attr];
+};
+
+// Helps name a CSS variable based on the class prefix.
+export function nameCSSVar(name: string): string {
+  return `--${CLASS_PREFIX}-${name}`;
+}
+
+// Takes a theme and turns it into a CSSProperties that sets CSS Variables.
+export function mapToCSSVar(colorConfig: Style): any {
```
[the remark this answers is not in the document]
melocule: Revised this function when I removed calls to createCSSVar, this should be fixed at this commit too
image

WesleyKapow (approved):
LGTM

melocule:
Tysm for your help Wesley!

PR #49, opened 2022-09-08
CTW-270/include more active statuses

WesleyKapow:
Setup more active statuses for condition query. Also remove filter from lens as we will always fetch whatever lens has.


luthfib (dismissed):
LGTM, note you didnt include a changeset here.

WesleyKapow:
@luthfib ready for re-review -- added changeset!

PR #52, opened 2022-09-09
Attempting to enable CI for changeset PRs

WesleyKapow:
Trying to follow https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/

I believe:
1. Adding `workflow_dispatch` to the validate workflow should at least make it available to be manually run.
2. Trying to get the release workflow to trigger the validate workflow.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,5 +1,6 @@
 name: Validation
 on:
+  workflow_dispatch:
   push:
     branches:
       - "main"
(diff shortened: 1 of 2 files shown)


.github/workflows/validate.yml
```diff
@@ -1,5 +1,6 @@
 name: Validation
 on:
+  workflow_dispatch:
   push:
```
luthfib, line 4: I wonder if the colon is going to cause an invalid syntax

luthfib (approved):

.github/workflows/validate.yml
luthfib: Actually oddlly enough thats correct

WesleyKapow: Shouldn't, that's the way yml is.. since there's nothing tabbed in underneath it, it just has zero config associated with it.

PR #54, opened 2022-09-13
CTW-293 Storybook

melocule:
Added a UI component explorer for testing.

Includes tests for the following components:
- Conditions (does real fetches)
- ConditionsTableBase
- DataList
- Drawer
- Spinner
- Table

In future: Mock data fetches


scripts/prefixer/prefixer.ts
```diff
@@ -114,7 +124,7 @@ function getReplaceRegEx(file: string, toPrefix: string): RegExp {
 function getStringsToPrefix(): Set<string> {
   const map = JSON.parse(readFileSync(mapFile, "utf-8")).selectors;
   const stringsToPrefix = new Set<string>();
-  map.forEach((oldClassName: string) => {
+  for (let oldClassName in map) {
```
luthfib, line 127: Why did we change this?

src/stories/resources/conditions.ts
```diff
@@ -0,0 +1,9666 @@
+import { TableColumn } from "@/components/core/table/table";
+import { ConditionModel } from "@/models/conditions";
+
+export const CONDITION_COLUMNS: TableColumn<ConditionModel>[] = [
```
luthfib, line 4: I think CONDITION_COLUMNS probably shouldn't go in here. Honestly probably could go in the Table.stories file dunno. This file is largely dominated by the mock data... how many conditions are we mocking?

melocule:
To leave as pending until next alpha version (7.0.0-alpha.34) of Storybook that includes [this PR](https://github.com/storybookjs/storybook/pull/19123).

scripts/prefixer/prefixer.ts
melocule: Shouldn't make a difference but it's because when ESLint/Prettier were fighting each other, one of them was complaining about me using map instead of a for-loop

src/stories/resources/conditions.ts
melocule: I'm mocking 4 conditions here, two confirmed and two unconfirmed. I'll throw this constant into the Table.stories file to keep the resources folder referring to fhir resources!

PR #55, opened 2022-09-14
CTW-173/embed conditions

WesleyKapow:
Numerous tweaks to get embedding conditions working in CTW app. I worked on this while copying the `dist` folder over to my CTW App's `node_modules` directory so that I can test things out. Likely more testing and tweaks needed after this gets pushed to NPM.

* Switch to use `.cjs` extension for cjs module files. This way they do not get picked up, and we instead correctly pick up the `es.js` files.
* Attempt to setup better reference for style.css.
* Fix how typings are specified.
* Switch app to use dev environment.
* Export patient provider component.
* Tweak a few styles.
* Sort conditions by name.
* Work around dropdown menu position issue.

### DIFF
--- a/src/styles/tailwind.css
+++ b/src/styles/tailwind.css
@@ -138,8 +138,3 @@
 *[class*="ctw-"]:after {
   box-sizing: border-box;
 }
-
-*[class*="ctw-"] button,
-input {
-  margin: 0;
-}
--- a/.changeset/nervous-cameras-beam.md
+++ b/.changeset/nervous-cameras-beam.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Fixup library packaging, tweak few styles, and make PatientProvider exported.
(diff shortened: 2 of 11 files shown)


src/styles/tailwind.css
```diff
@@ -138,8 +138,3 @@
 *[class*="ctw-"]:after {
   box-sizing: border-box;
 }
-
-*[class*="ctw-"] button,
```
luthfib, line 142: Was this not used? I forget why we put this, but maybe we don’t need it anymore?

luthfib (approved):
Looks good to me, just one comment on that ctw selector

src/styles/tailwind.css
luthfib: Actually pretty sure that is for the toggle. We could instead apply that class directly to the toggle but either way I think this breaks the how the toggle is used

WesleyKapow: I added `ctw-m-0` to the toggle.

I removed this because it is a large hammer. It was also bugged in that it do this for ALL `input` elements (not just ones underneath `ctw-`.

luthfib: ahh gotcha, didnt see that. Cool

PR #56, opened 2022-09-15
Ctw 283/edit add conditions

luthfib:
Creating a draft pull request, not yet ready for review but will mark it as such when ready. 

Still have to: 

- [ ] Fix error when posting to conditions (When posting the ID provided is said to be incorrect)
- [ ] Fix typescript issues
- [ ] Figure out how to correctly handle patientProvider. Currently it always returns a patientModel, but we probably need to change to handle the case where we pass a UPID
- [ ] Possible refactors to cleanup how code is structured (Lots of moving parts, might be able to simplify it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@heroicons/react": "^1.0.6",
     "date-fns": "^2.29.1",
     "fhir-kit-client": "^1.9.1",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "zod": "^3.19.0"
   }
 }
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -167,7 +167,7 @@ export function Conditions({ className }: ConditionsProps) {
 
       <DrawerFormWithFields
         title="Add Condition"
-        action="createMedicationStatement"
+        action="createCondition"
         data={getConditionFormData(newCondition)}
         schema={conditionSchema}
         isOpen={addConditionIsOpen}
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -4,9 +4,11 @@ import {
   getLensConditions,
 } from "@/fhir/conditions";
 import { ConditionModel } from "@/models/conditions";
+import { conditionSchema, getConditionFormData } from "@/utils/helpers";
 import cx from "classnames";
 import { useEffect, useState } from "react";
 import { useCTW } from "../core/ctw-provider";
+import { DrawerFormWithFields } from "../core/forms/drawer-form-with-fields";
 import { ToggleControl } from "../core/toggle-control";
 import { ConditionFormDrawer } from "./condition-form-drawer";
 import { ConditionsTableBase } from "./conditions-table-base";
@@ -98,6 +100,16 @@ export function Conditions({ className, patientUPID }: ConditionsProps) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [patientUPID, includeInactive]);
 
+  const newMedicationStatement = new ConditionModel({
+    resourceType: "Condition",
+    subject: { reference: `Patient/${patient.id}`, display: patient.display },
+    informationSource: {
+      type: "Organization",
+      reference: `Organization/${patient.organization?.id}`,
+      display: patient.organization?.name,
+    },
+  });
+
   return (
     <div
       className={cx(
@@ -156,6 +168,15 @@ export function Conditions({ className, patientUPID }: ConditionsProps) {
           </div>
         </div>
       </div>
+
+      <DrawerFormWithFields
+        title="Add Condition"
+        action="createMedicationStatement"
+        data={getConditionFormData(newMedicationStatement)}
+        schema={conditionSchema}
+        isOpen={addConditionIsOpen}
+        onClose={() => setAddConditionIsOpen(false)}
+      />
 
       <ConditionFormDrawer
         isOpen={addConditionIsOpen}
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -8,6 +8,7 @@ import {
   getLensConditions,
 } from "@/fhir/conditions";
 import { ConditionModel } from "@/models/conditions";
+import { PatientModel } from "@/models/patients";
 import cx from "classnames";
 import { useEffect, useState } from "react";
 import { useCTW } from "../core/ctw-provider";
@@ -20,37 +21,25 @@ export type ConditionsProps = {
   className?: string;
 };
 
+const DEFAULT_MSG = "No conditions found";
 const DEFAULT_ERR_MSG =
   "There was an error fetching conditions for this patient. Refresh the page or contact your organization's technical support if this issue persists.";
 
 export function Conditions({ className }: ConditionsProps) {
   const [addConditionIsOpen, setAddConditionIsOpen] = useState(false);
-
-  const [confirmedConditions, setConfirmedConditions] = useState<
-    ConditionModel[]
-  >([]);
-  const [confirmedConditionsMessage, setConfirmedConditionsMessage] = useState(
-    "No conditions found"
-  );
-  const [confirmedConditionsIsLoading, setConfirmedConditionsIsLoading] =
-    useState(true);
-
-  const [notReviewedConditions, setNotReviewedConditions] = useState<
-    ConditionModel[]
-  >([]);
-  const [notReviewedConditionsIsLoading, setNotConfirmedConditionsIsLoading] =
-    useState(true);
-  const [notReviewedConditionsMessage, setNotReviewedConditionsMessage] =
-    useState("No conditions found");
-
+  const [confirmed, setConfirmed] = useState<ConditionModel[]>([]);
+  const [confirmedMessage, setConfirmedMessage] = useState(DEFAULT_MSG);
+  const [confirmedIsLoading, setConfirmedIsLoading] = useState(true);
+  const [notReviewed, setNotReviewed] = useState<ConditionModel[]>([]);
+  const [notReviewedIsLoading, setNotReviewedIsLoading] = useState(true);
+  const [notReviewedMessage, setNotReviewedMessage] = useState(DEFAULT_MSG);
   const [includeInactive, setIncludeInactive] = useState(true);
-  const { getCTWFhirClient } = useCTW();
+  const [patient, setPatient] = useState<PatientModel>();
 
+  const { getCTWFhirClient } = useCTW();
   const { patientUPIDPromise } = usePatient();
 
   const handleFormChange = () => setIncludeInactive(!includeInactive);
-
-  console.log("notReviewedConditions", notReviewedConditions);
 
   useEffect(() => {
     async function load() {
@@ -61,7 +50,9 @@ export function Conditions({ className }: ConditionsProps) {
         : {};
 
       const fhirClient = await getCTWFhirClient();
-      const patientUPID = await patientUPIDPromise;
+      const patientData = await patientUPIDPromise;
+      const patientUPID = patientData.UPID;
+      setPatient(patient);
 
       // use AllSettled instead of all as we want confirmed to still if lens fails
       const [confirmedConditionInfo, notReviewedConditionInfo] =
@@ -70,13 +61,13 @@ export function Conditions({ className }: ConditionsProps) {
           getLensConditions(fhirClient, patientUPID),
         ]);
 
-      setNotConfirmedConditionsIsLoading(false);
-      setConfirmedConditionsIsLoading(false);
+      setNotReviewedIsLoading(false);
+      setConfirmedIsLoading(false);
 
       /* notReviewedConditons depends confirmedConditions so that we can correctly filter out 
          conditions that appear in confirmedConditions from notReviewedConditons */
       if (confirmedConditionInfo.status === "fulfilled") {
-        setConfirmedConditions(
+        setConfirmed(
           confirmedConditionInfo.value.map((c) => new ConditionModel(c))
         );
         const ICD10ConfirmedCodes = confirmedConditionInfo.value.map(
@@ -88,18 +79,18 @@ export function Conditions({ className }: ConditionsProps) {
             notReviewedConditionInfo.value.filter(
               (c) => !ICD10ConfirmedCodes.includes(new ConditionModel(c).icd10)
             );
-          setNotReviewedConditions(
+          setNotReviewed(
             notReviewedConditionsFiltered.map((c) => new ConditionModel(c))
           );
         } else {
-          setNotReviewedConditions([]);
-          setNotReviewedConditionsMessage(DEFAULT_ERR_MSG);
+          setNotReviewed([]);
+          setNotReviewedMessage(DEFAULT_ERR_MSG);
         }
       } else {
-        setConfirmedConditions([]);
-        setConfirmedConditionsMessage(DEFAULT_ERR_MSG);
-        setNotReviewedConditions([]);
-        setNotReviewedConditionsMessage(DEFAULT_ERR_MSG);
+        setConfirmed([]);
+        setConfirmedMessage(DEFAULT_ERR_MSG);
+        setNotReviewed([]);
+        setNotReviewedMessage(DEFAULT_ERR_MSG);
       }
     }
     load();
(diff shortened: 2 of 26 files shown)


src/components/core/patient-provider.tsx
```diff
@@ -42,29 +42,25 @@ export function PatientProvider({
...
 
   useEffect(() => {
-    async function getPatientUPID() {
-      if (patientUPID) {
-        return patientUPID;
-      }
+    async function getPatient() {
+      // if (patientUPID) {
+      //   return patientUPID;
+      // }
```
WesleyKapow, line 52: BLOCKER: Handle UPID!

src/fhir/errors.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const fhirErrorResponse = (title: string, e: FhirError) => {
+  const { data, status } = e.response;
+  let errorStr = "";
+  if (isOperationOutcome(data)) {
+    errorStr = listErrorIssues(data);
+  }
+
+  return { ...{ title }, ...{ status, statusText: errorStr } };
```
WesleyKapow, line 29: Clean this up `{title, status statusText: errorStr}`

src/fhir/errors.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export type FhirError = {
+  response: {
+    status: number;
+    data: unknown;
+  };
+  config: {
+    method: string;
+    url: string;
+    headers: { [key: string]: string };
```
WesleyKapow, line 11: `Record<string, string>` (told you I didn't use `Record` enough.. let's try and fix that for the component library code).

src/models/conditions.ts
```diff
@@ -1,12 +1,37 @@
...
+      return await fhirClient.create({
+        resourceType: "Condition",
+        body: omitEmptyArrays(this.resource) as fhir4.Condition,
+      });
+    } catch (err) {
+      if (isFhirError(err)) {
+        return err;
+      }
+      throw Error("Failed saving error");
+    }
```
WesleyKapow, line 34: Having models save was a mistake in CTW App that we started to move away from. Can you refactor this to probably a generic `save` function in `fhir/client.ts` that knows how to update/create any fhir resource.

src/models/conditions.ts
```diff
@@ -131,7 +156,23 @@ export class ConditionModel {
...
   get verificationStatus(): string {
     return codeableConceptLabel(this.resource.verificationStatus);
   }
+
+  public setSubjectID(patientID: string) {
+    this.resource.subject = {
+      type: "Patient",
+      reference: `Patient/${patientID}`,
+    };
+  }
```
WesleyKapow, line 177: Along with the above note about refactoring how we save changes, refactor this out to where we are constructing the fhir condition that we want to save.

src/styles/tailwind.css
```diff
@@ -10,6 +10,64 @@
...
+  }
+
+  input.dark:focus-visible {
+    outline-offset: -1px;
+    @apply ctw-outline-2 ctw-outline-white;
+  }
+
+  label.error {
+    @apply ctw-text-error-main;
+  }
```
WesleyKapow, line 63: BLOCKER: These are not namepaced to CTW. For example `input.error` would impact any input with an `error` class.

src/components/core/forms/actions.ts
```diff
@@ -0,0 +1,84 @@
...
+  SYSTEM_CONDITION_CLINICAL,
+  SYSTEM_CONDITION_VERIFICATION_STATUS,
+  SYSTEM_SNOMED,
+} from "@/fhir/system-urls";
+import { ConditionModel } from "@/models/conditions";
+import { getFormData } from "@/utils/form-helper";
+import Client from "fhir-kit-client";
+import { conditionSchema } from "./schemas";
+
+export const action = async (
```
WesleyKapow, line 13: I don't think we need this abstraction. Instead each use of a form should handle the success case.

src/components/core/forms/actions.ts
```diff
@@ -0,0 +1,84 @@
...
+  switch (formAction) {
+    case "createCondition":
+      return createCondition(formData, patientID, getCTWFhirClient);
+    default: {
+      throw new Error(`Unexpected action: ${formAction}`);
+    }
+  }
+};
+
+export const createCondition = async (
```
WesleyKapow, line 28: Hmm this should probably live wherever the definition of the condition schema is and the listing of the fields. Maybe `content/forms/condition.ts` as mentioned elsewhere?

src/components/core/forms/actions.ts
```diff
@@ -0,0 +1,84 @@
...
+      abatementDateTime: dateToISO(result.data.abatement),
+    }),
+    onsetDateTime: dateToISO(result.data.onset),
+    recordedDate: dateToISO(result.data.recordedDate),
+    subject: { type: "Patient", reference: `Patient/${patientID}` },
+    note: result.data.note ? [{ text: result.data.note }] : undefined,
+  };
+
+  const conditionModel = new ConditionModel(fhirCondition);
+  const response = await conditionModel.save(getCTWFhirClient);
```
WesleyKapow, line 77: As mentioned in the model, we should move away from `model.save`. Instead construct the FHIR resource as you are and in then pass that to a generic `save` method that can handle any FHIR resource.

src/components/core/forms/condition-helpers.ts
```diff
@@ -0,0 +1,94 @@
...
+  condition,
+}: {
+  condition: ConditionModel;
+}): FormEntry[] => [
+  {
+    label: "subject",
+    value: condition.subjectID,
+    field: "subjectID",
+    readonly: true,
+  },
```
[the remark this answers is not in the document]
luthfib: Good point, this was before I had added hidden functionality added but will update it.

src/components/core/forms/actions.ts
luthfib: The idea here is that we can pass a actionName to the form and based on that we can run a function. This way the drawer-form stays generic. An alternative I can think of is to pass a reference to a function

src/components/content/conditions.tsx
```diff
@@ -159,10 +200,17 @@ export function Conditions({ className }: ConditionsProps) {
         </div>
       </div>
 
-      <ConditionFormDrawer
-        isOpen={addConditionIsOpen}
-        onClose={() => setAddConditionIsOpen(false)}
-      />
+      {patient && currentSelectedData && (
```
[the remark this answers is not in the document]
luthfib: We can either include it in here or have `DrawerFormWithFields` handle the case where it is not defined.

src/components/core/forms/actions.ts
WesleyKapow: Yes, pass in a function!

Passing in a function is more explicit. It also means that when adding a new form, we don't need to update any other file. Where your current solution would be to pass in a new action name and update the switch that figures out what function to call based on the function name (very indirect!).

src/components/content/conditions.tsx
```diff
@@ -127,11 +144,23 @@ export function Conditions({ className }: ConditionsProps) {
...
-              message={confirmedConditionsMessage}
+              conditions={confirmed}
+              isLoading={confirmedIsLoading}
+              message={confirmedMessage}
               rowActions={[
-                { name: "Edit", action: () => setAddConditionIsOpen(true) },
+                {
+                  name: "Edit",
+                  action: (_, condition) => {
+                    // setAddConditionIsOpen(true);
```
pmadria3, line 154: nit: remove unused code

src/utils/form-helper.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+export function processDef(
+  def: ZodTypeAny,
+  o: any,
+  key: string,
+  value: string
+) {
+  let parsedValue: any;
+  if (def instanceof ZodString || def instanceof ZodLiteral) {
```
pmadria3, line 60: you can probably utilize a switch statement here

src/utils/form-helper.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+export function getInputProps(name: string, def: ZodTypeAny): InputPropType {
+  let type = "text";
+  let min;
+  let max;
+  let minlength;
+  let maxlength;
+  let pattern;
+  if (def instanceof ZodString) {
```
pmadria3, line 209: you can probably also utilize a switch statement here

src/utils/form-helper.ts
WesleyKapow: What would that switch statement look like?

WesleyKapow: Also note, much of this code was copied from a remix library -- @luthfib probably worth adding a comment at the top where this comes from and main modifications made.

pmadria3: do something like the following? 

Screen Shot 2022-09-19 at 1 23 17 PM

luthfib: @WesleyKapow Actually already have a comment at the top about the fact that it was ported over. I can add additionally comments on changes made.

pmadria3: Ah thanks for bringing that to light @WesleyKapow @luthfib - we can leave as is

WesleyKapow: @pmadria3 `instanceof` does not work that way :'(

pmadria3: ah yeah you're right - I did a bit more digging into `instanceof` and it won't be possible unless we declare enums and such. ([link](https://stackoverflow.com/questions/5579309/is-it-possible-to-use-the-instanceof-operator-in-a-switch-statement))

package.json
```diff
@@ -77,11 +77,13 @@
     "vitest": "^0.23.1"
   },
   "dependencies": {
+    "@headlessui-float/react": "^0.9.1",
```
MrMonsk, line 80: I remember this!

src/components/content/forms/form-field.tsx
```diff
@@ -0,0 +1,104 @@
...
+  defaultValue,
+  readonly,
+  hidden,
+  ...inputProps
+}: FormFieldProps) => {
+  // We display dates in MM/DD/YYYY format, but date input fields
+  // expect it to be in YYYY-MM-DD format.
+  const value =
+    inputProps.type === "date"
+      ? formatDateLocalToISO(defaultValue as string)
```
pmadria3, line 28: defaultValue should be a string so we can remove the as string here

pmadria3 (approved):
LGTM - one small comment but not that earth shattering if you wanna go ahead and merge it in

src/components/content/forms/form-field.tsx
pmadria3: it can be undefined and if that's fine then this should be okay

pmadria3 (approved):
LGTM

pmadria3 (dismissed):
LGTM

src/components/content/conditions.tsx
```diff
@@ -147,24 +154,23 @@ export function Conditions({ className }: ConditionsProps) {
...
-                    //   })
-                    // );
+                  action: () => {
+                    if (patient) {
+                      setDrawerIsOpen(true);
+                      setFormAction("Edit");
+                      setCurrentlySelectedData(
+                        getEditingFromLensConditionData({ condition })
+                      );
+                    }
```
WesleyKapow, line 167: Pull out to helper `handleEditCondition`. And change `addNewCondition` to `handleAddNewCondition` .. `handle*` is a pretty common pattern for names of click handlers.

src/components/content/conditions.tsx
```diff
@@ -179,24 +185,23 @@ export function Conditions({ className }: ConditionsProps) {
...
-                      })
-                    );
+                  action: () => {
+                    if (patient) {
+                      setDrawerIsOpen(true);
+                      setFormAction("Add");
+                      setCurrentlySelectedData(
+                        getAddConditionData({ condition })
+                      );
+                    }
```
WesleyKapow, line 198: `handleAddNotReviewedCondition`

src/components/content/conditions.tsx
```diff
@@ -147,24 +154,23 @@ export function Conditions({ className }: ConditionsProps) {
...
-                    //     condition,
-                    //     patientID: patient.id,
-                    //   })
-                    // );
+                  action: () => {
+                    if (patient) {
+                      setDrawerIsOpen(true);
+                      setFormAction("Edit");
+                      setCurrentlySelectedData(
+                        getEditingFromLensConditionData({ condition })
```
WesleyKapow, line 165: Editing from lens? This is editing the confirmed right?

src/components/content/conditions.tsx
```diff
@@ -206,15 +211,22 @@ export function Conditions({ className }: ConditionsProps) {
...
+          action={createCondition}
           data={currentSelectedData}
           schema={conditionSchema}
-          isOpen={addConditionIsOpen}
-          onClose={() => setAddConditionIsOpen(false)}
+          isOpen={drawerIsOpen}
+          onClose={() => setDrawerIsOpen(false)}
+        />
+      )}
+      {currentlyClickedCondition && (
```
WesleyKapow, line 225: `currentlyClickedCondition` there's no clicking a condition. Maybe change to `conditionForHistory`?

src/components/content/forms/actions.ts
```diff
@@ -0,0 +1,17 @@
+import Client from "fhir-kit-client";
+import { createCondition } from "./conditions";
+
+export const action = async (
```
WesleyKapow, line 4: Again this should not be needed. Inline actions instead of using this indirection.

src/components/content/forms/conditions.ts
```diff
@@ -38,6 +55,7 @@ export const createCondition = async (
   // Some fields will need to be set as they are required.
   const fhirCondition: fhir4.Condition = {
     resourceType: "Condition",
+    ...(result.data.id && { id: result.data.id }),
```
WesleyKapow, line 58: Is this needed? Why not just `id: result.data.id` and if it is undefined, then that's fine?

src/components/content/forms/drawer-form-with-fields.tsx
```diff
@@ -1,52 +1,72 @@
...
-export type DrawerFormWithFieldsProps = {
+export type DrawerFormWithFieldsProps<T> = {
   title: string;
-  actionName: string;
-  data: FormEntry[];
+  action: (
+    data: FormData,
+    patientID: string,
+    getCTWFhirClient: () => Promise<Client>
+  ) => Promise<ActionReturn<T>>;
```
WesleyKapow, line 25: Why does action need all of this? Why can't the form just pass in the formData to the action. The action function itself should be setup/defined in such a way that it already has whatever else it needs.

By adding `patientID` and `getCTWFhirClient` you are making this form less generic. It now is setup to be a form that triggers an action that depends on patientID and fhir client.. that may not always be true!

src/components/content/forms/form-field.tsx
```diff
@@ -65,6 +66,17 @@ export const FormField = ({
...
+    return (
+      <input
+        {...inputProps}
+        className={cx({ error }, "ctw-listbox-input ctw-w-full")}
+        defaultValue={value}
+        hidden={hidden}
+      />
+    );
+  }
+
```
WesleyKapow, line 79: This looks weird to me.

1. You don't need `className` on a hidden input.
2. Just do `hidden` and not `hidden={hidden}` since you know it is true.

src/fhir/action-helper.ts
```diff
@@ -0,0 +1,40 @@
...
+    return await fhirClient.create({
+      resourceType: resourceModel.resourceType,
+      body: omitEmptyArrays(resourceModel.resource) as typeof resourceModel,
+    });
+  } catch (err) {
+    if (isFhirError(err)) {
+      return err;
+    }
+
+    throw Error("Failed saving error");
```
WesleyKapow, line 38: Make this error better.

WesleyKapow (changes requested):

src/components/content/forms/conditions.ts
luthfib: I think you are right. I'll test. I did this because we decide to create or add based on if the id is there, but theoretically having it as undefined should work as well.

src/components/content/forms/drawer-form-with-fields.tsx
luthfib: The reason I did this is because both patientID and getCTWFhirClient come from a hook. As createCondition is not a hook I cannot make the call in there. I have to pass it down from some component, I mean maybe I could make the call call to get patient from the actual drawer?

luthfib: Or actually maybe we could pass in a reference to a function

WesleyKapow: Let's leave this as is for now. I want to get this PR in! I'll have a better feel when I get into the code myself.

But in general my hope is that the form components take an action that ONLY takes form data.

luthfib: Yeah, I agree.

WesleyKapow (approved):
LGTM

PR #60, opened 2022-09-19
Quick style fix & update README

WesleyKapow:
Had an issue importing style.css via `import ctwStyles from "@zus-health/ctw-component-library/style.css";`, even though the package.json maps it to the correct `dist` directory. Instead we'll have users include it directly off of the `dist` directory.

Update README file to provide install & setup instructions.


luthfib (approved):
LGTM

luthfib (approved):

PR #63, opened 2022-09-20
Edit Condition History to show both of them

pmadria3:
This PR adds fetching by icd10 or snomed and if there are both it displays both in the condition history drawer. First search is by icd10. 
Screen Shot 2022-09-20 at 12 46 49 PM

### DIFF
--- a/src/components/content/conditions-history.tsx
+++ b/src/components/content/conditions-history.tsx
@@ -84,15 +84,15 @@ export function ConditionHistory({
               value: condition.icd10System,
             },
             {
-              label: "SnoMed Display",
+              label: "SNOMED Display",
               value: condition.snomedDisplay,
             },
             {
-              label: "SnoMed Code",
+              label: "SNOMED Code",
               value: condition.snomedCode,
             },
             {
-              label: "SnoMed System",
+              label: "SNOMED System",
               value: condition.snomedSystem,
             },
           ],
--- a/src/components/content/conditions-history.tsx
+++ b/src/components/content/conditions-history.tsx
@@ -48,8 +48,6 @@ export function ConditionHistory({
         );
       }
 
-      console.log(filteredConditions);
-
       setConditions(filteredConditions.map((model) => setupData(model)));
       setLoading(false);
     }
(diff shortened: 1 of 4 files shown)


luthfib (changes requested):
Correct me if I am wrong but I believe we want to use both snomed code AND ICD10 code if they are present. As currently constructed ICD10 code has presedence over snomed code.

src/components/content/conditions-history.tsx
```diff
@@ -31,49 +35,140 @@ export function ConditionHistory({ icd10 }: ConditionHistoryProps) {
...
+          {
+            label: "Code",
+            value: condition.icd10Code,
+          },
+          {
+            label: "System",
+            value: condition.icd10System,
+          },
+        ],
+      };
```
luthfib, line 167: You can simplify this to something along the lines of... 

```
let data = [
        {
          label: "Verification Status",
          value: condition.verificationStatus,
        },
        {
          label: "Clinical Status",
          value: condition.clinicalStatus,
        },
        {
          label: "Recorded Date",
          value: condition.recordedDate,
        },
      ];
      const ICD10Fields = [
        {
          label: "ICD10 Display",
          value: condition.icd10Display,
        },
        {
          label: "ICD10 Code",
          value: condition.icd10Code,
        },
        {
          label: "ICD10 System",
          value: condition.icd10System,
        },
      ];
      const SNOMEDFields = [
        {
          label: "SNOMED Display",
          value: condition.snomedDisplay,
        },
        {
          label: "SNOMED Code",
          value: condition.snomedCode,
        },
        {
          label: "SNOMED System",
          value: condition.snomedSystem,
        },
      ];

      if (icd10Code) {
        data = data.concat(ICD10Fields);
      }

      if (snomedCode) {
        data = data.concat(SNOMEDFields);
      }

      return {
        id: condition.id,
        data: [...data],
      };
```

luthfib (changes requested):
Looking good, but found some areas where the code surface can be reduced

src/components/content/conditions-history.tsx
```diff
@@ -31,49 +35,140 @@ export function ConditionHistory({ icd10 }: ConditionHistoryProps) {
...
+      } else if (icd10Code) {
+        filteredConditions = models.filter(
+          (condition) => condition.icd10Code === icd10Code
+        );
+      } else {
+        filteredConditions = models.filter(
+          (condition) => condition.snomedCode === snomedCode
+        );
+      }
+
```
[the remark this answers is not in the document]
pmadria3: ah nice - used this TY

src/components/content/conditions-history.tsx
pmadria3: much cleaner - totally forgot about concat. TY.

luthfib (approved):
LGTM

PR #64, opened 2022-09-20
Update changeset for fetch

pmadria3:
Update changeset for the change to fetch the icd10 and snomed codes via filter

### DIFF
--- a/.changeset/weak-points-tickle.md
+++ b/.changeset/weak-points-tickle.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Fix the filter to fetch by icd10 and snomed and not just one or the other if both are present.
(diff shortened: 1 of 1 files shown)


luthfib (changes requested):
What fetch?

.changeset/weak-points-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Fix the fetch to fetch by icd10 and snomed and not just one or the other if both are present.
```
luthfib, line 5: What fetch? Need to be more clear since this will be in a changelog with no context

luthfib (changes requested):

luthfib (approved):

PR #67, opened 2022-09-21
Bugfix - reset the state after viewing condition

pmadria3:
This resets the state so if you view condition A and then click on condition B, it doesn't show you condition A.

### DIFF
--- a/.changeset/rich-pets-refuse.md
+++ b/.changeset/rich-pets-refuse.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Couple fixes reseting the loading state after fetching condition history for a particular condition and increasing fetch size.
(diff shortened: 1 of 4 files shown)


src/fhir/search-helpers.ts
```diff
@@ -30,6 +30,8 @@ export type SearchReturn<T extends ResourceTypeString> = {
   resources: ResourceType<T>[];
 };
 
+const MAX_COUNT = 250;
```
luthfib, line 33: Move CONSTANTS to top file

.changeset/rich-pets-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Couple fixes reseting the loading state after fetching condition history for a particular condition and increasing fetch size.
```
luthfib, line 5: How about... 
- "Resets history loading state after fetching condition history for a particular condition and increases fetch size for searchAllRecords"

luthfib (approved):

PR #70, opened 2022-09-22
Simplify tables

WesleyKapow:
Removed sticky column and supporting shadows.

### DIFF
--- a/.changeset/tough-rice-remain.md
+++ b/.changeset/tough-rice-remain.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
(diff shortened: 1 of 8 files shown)


luthfib (dismissed):
Some good ole removal of code. My only question here would be if it makes sense to include actual patient id's. Its a sample patient, but would everyone necessarily have access to that patient? Also noticed we are having a type issue that I didnt notice first time around

WesleyKapow:
@luthfib if they don't have access, they'll know. That's also why I included the comment of env/builder/patient.

luthfib (approved):

PR #72, opened 2022-09-22
Refactor styles

WesleyKapow:
**NOTE TO REVIEWER: Probably best to review single commit at a time!**

Switching to use dynamic importing with SCSS (sass). This allows each of our components to specify and import their own CSS.

Few benefits:
* The use of SCSS makes it easier for us to define component styles via selector nesting.
* The `CTWProvider` component now imports our main CSS so consuming apps no longer need to manage importing styles themselves.
* We no longer have a single growing css file and consuming apps will only pull in the styles of the components they use.

Additionally this PR removes `onRowClick` from table as it is unused.

TODO:
1. Refactor listbox styles.
2. Maybe switch to CSS modules .. turns out this is as simple as renaming the files to `foo.module.scss` and then `import classes from "foo.module.scss"` and `className={classes.foo}` that removes the need for BEM like naming.
3. Specificity troubles! Previously all of our custom CSS classes were in the `base` layer which tailwind would compile to the top of the generated CSS file. With dynamic CSS imports, you get them in the order things are imported. This means we have `main.scss` showing up above our component css in the DOM. This results in not being able to simply overwrite classes, e.g. our table uses `<div className={cx("ctw-table-container", className)}>`. If the passed in `className` uses tailwind utilities to try and overwrite `ctw-table-container` properties, it will fail since the utilities are defined ABOVE `.ctw-table-container` in the DOM. I think this is acceptable for now, we can in the meantime use `!ctw-` to make overwrites use `important`. We'll need to think more on this in the future.
4. Figure out a better way to organize borders! Without the nice defaulting (tailwind preflight), we end up having to manage `border-0` and `border-solid` ourselves.


luthfib (commented):
Looking good so far. I do love me some Sass and have had a pretty good experience with css modules, dont really want to go back to using BEM. I did find it odd that the consumers had to import the CSS styles themselves so this should make using our components even easier and more modular.

pmadria3 (approved):
Really like what I see! LGTM

luthfib (approved):
LGTM

PR #73, opened 2022-09-22
CTW-327 Followup: Condition Table Column Width Patch

melocule:
Recalculated the Condition table column width classes to total 100% to prevent confusion. This will not impact how the table is viewed.


pmadria3 (approved):
LGTM

PR #77, opened 2022-09-23
Ctw 358/refetch condition when adding or editing

luthfib:
- Adds React Query to enable refetching when adding or when refocusing on page.
- Fixes bug where date would crash if parser couldnt parse date
- Used ref to pass FHIR Client on advice of creator of react Query. A different approach was also given, in which we would write a custom hashing function for the keys that and then we would potentially hash the fhir client based on the auth header, that way we would know if the auth header had been updated and we would know that we have re calculate the key


.changeset/warm-dolls-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Refetches condition when adding new condition or adding condition from not reviewed. Additionally fixes bug when date editing condition and date is of bad format
```
WesleyKapow, line 5: Maybe typo here: "when date editing condition and date" .. probably remove first "date".

src/fhir/formatters.ts
```diff
@@ -11,9 +11,12 @@ export function formatDateISOToLocal(dateStr?: string): string | undefined {
...
   if (!dateStr) return undefined;
-
-  const date = parse(dateStr, "P", new Date());
-  return formatISO(date, { representation: "date" });
+  try {
+    const date = parse(dateStr, "P", new Date());
+    return formatISO(date, { representation: "date" });
+  } catch (e) {
+    return dateStr;
+  }
```
WesleyKapow, line 19: Do you know what dates we are passing in that fail this? Should we be calling a different helper instead of `formatDateLocalToISO`?

src/components/content/conditions.tsx
```diff
@@ -42,16 +44,37 @@ export function Conditions({ className }: ConditionsProps) {
   const [includeInactive, setIncludeInactive] = useState(true);
   const [patient, setPatient] = useState<PatientModel>();
   const [formAction, setFormAction] = useState("");
+  // We use a ref here because if we pass FHIR Client to meta normally it will use whatever
+  // you passed as meta when the query was created which means it can be come stale
```
WesleyKapow, line 48: "can become stale."

Also, this will likely need to be done everywhere we do data fetching. A comment at each use isn't great. Instead see if you can refactor any of this to a helper. Something like `usefhirClientRef()` would be great. And then a description of why we use refs would go there.

src/components/content/conditions.tsx
```diff
@@ -42,16 +44,37 @@ export function Conditions({ className }: ConditionsProps) {
...
   const [currentSelectedData, setCurrentlySelectedData] =
     useState<FormEntry[]>();
   const [conditionForHistory, setConditionForHistory] =
     useState<ConditionModel>();
+  const confirmedResponse = useQuery(
+    ["conditions", patientUPID, conditionFilter],
+    getConfirmedConditions,
+    {
+      enabled: !!patientUPID,
+      meta: fhirClientRef as unknown as Record<string, unknown>,
```
WesleyKapow, line 61: These type casts seem wrong to me.

Meta is a map from string to unknown but you are trying to pass fhirClientRef directly to it, don't do this!

Instead try `meta: {fhirClientRef}` and then use it via `meta.fhirClientRef`.

src/components/content/conditions.tsx
```diff
@@ -84,60 +107,64 @@ export function Conditions({ className }: ConditionsProps) {
...
           setNotReviewedMessage(ERROR_MSG);
         }
       }
+
+      if (confirmedResponse.error) {
+        setConfirmed([]);
+        setConfirmedMessage(ERROR_MSG);
+        setNotReviewed([]);
+        setNotReviewedMessage(ERROR_MSG);
+      }
```
WesleyKapow, line 156: Probably want to split and detect if confirmed OR reviewed have an error.

src/fhir/conditions.ts
```diff
@@ -42,18 +61,32 @@ export async function getConfirmedConditions(
...
+  const { meta, queryKey } = queryParams;
+  let fhirClient;
+
+  if (meta?.current) {
+    fhirClient = meta.current as Client;
+  }
+
+  if (!fhirClient) {
+    throw Error("Fhir Client is required");
+  }
```
WesleyKapow, line 81: Refactor / DRY

WesleyKapow (changes requested):

src/components/content/forms/conditions.ts
```diff
@@ -102,5 +103,7 @@ export const createCondition = async (
     result.success = false;
   }
 
+  queryClient.invalidateQueries(["conditions"]);
```
WesleyKapow, line 106: Do edits work correctly? As in they don't need to invalidate anything and instead we correctly show the edited condition in the table?

src/fhir/formatters.ts
luthfib: the date comes in as `201911011338`. We could create another helper to call if we see a date in that format maybe?

WesleyKapow: Well that looks like a valid dateTime to me. Can you:
1. Verify with team that we only want to show DATEs here and not also hh:mm?
2. Update this function to return dates when given dateTime strings.

src/components/content/conditions.tsx
luthfib: Do we want to display notReviewed if confirmed doesnt load? Because if confirmed doesnt load then we wouldnt want to incorrectly show lens conditions that were supposed to be filtered through the snomed, idc10 and other codes

src/components/content/forms/conditions.ts
luthfib: Invalidating should be fine, because that will retrigger a refetch and it will update the table with the new edited info. We could also theoretically edit the table and not retrigger a refetch but not sure if that should be a v2

WesleyKapow (changes requested):

src/components/content/forms/conditions.ts
WesleyKapow: `invalidateQueries` is being called in just `createCondition` right? So how does it work for edit?

src/fhir/formatters.ts
```diff
@@ -8,15 +8,30 @@ export function formatDateISOToLocal(dateStr?: string): string | undefined {
   return format(date, "P");
 }
 
+export const matchDatePattern = (dateStr: string): Date | string => {
+  const patterns = ["P", "yyyyMMddHHmm"];
+
+  // Going to try to parse all the patterns and if pattern is not recognized then will return date as is
+  for (let i = 0; i < patterns.length; i += 1) {
+    const parsedDate = parse(dateStr, patterns[i], new Date());
+    if (!(parsedDate instanceof Date && !Number.isNaN(parsedDate))) {
```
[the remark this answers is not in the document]
luthfib: No, it just returns an invalid Date Object

src/fhir/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  getCTWFhirClient()
+    .then((value) => {
+      fhirClientRef.current = value;
+    })
+    .catch((error) => {
+      throw Error("Failed to get the FHIR Client");
+    });
+
+  return { fhirClientRef };
```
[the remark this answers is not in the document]
luthfib: Good point. Uneeded

src/fhir/formatters.ts
```diff
@@ -8,15 +8,30 @@ export function formatDateISOToLocal(dateStr?: string): string | undefined {
...
+};
+
 // Formats a date from MM/DD/YYYY to YYYY-MM-DD.
 export function formatDateLocalToISO(dateStr?: string): string | undefined {
   if (!dateStr) return undefined;
-  try {
-    const date = parse(dateStr, "P", new Date());
+
+  const date = matchDatePattern(dateStr);
+  if (typeof date !== "string") {
```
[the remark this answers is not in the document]
luthfib: So we could check for `date instanceof Date` but technically that could be an invalid Date object, but theoretically our matchDatePattern function should guard against returning invalid dates

src/components/content/forms/conditions.ts
luthfib: Renamed the function for clarity, that is my bad. Renamed to... createOrEditCondition, which is what it is actually doing.

src/fhir/utils.ts
```diff
@@ -3,6 +3,8 @@ import Client from "fhir-kit-client";
 import { useRef } from "react";
 import { useCTW } from "../components/core/ctw-provider";
 
+// We use a ref here because if we pass FHIR Client to meta normally it will use whatever
+// you passed as meta when the query was created which means it can be come stale
```
WesleyKapow, line 7: This is now a bit out of place, like what is meta here? How about something like:

> A reference to fhirClient is useful for things like useQuery where we pass it in as part of the meta data. This allows useQuery to always use a fresh copy of fhirClient (whatever the reference points to).

WesleyKapow (approved):
nit about comment, otherwise LGTM

WesleyKapow (approved):

luthfib:
Looking into the error, might have been merge conflict errors

WesleyKapow (approved):

PR #78, opened 2022-09-23
CTW-248/responsive conditions

WesleyKapow:
Number of changes to prepare for responsive tables:
* Add better handling of column sizes (min-width and %)
* Fixup condition column sizes
* Add placeholder classes that can stack the table (to be used for responsive)
* Re-add back scroll shadows
* Hide table heading until data is loaded

Screen Shot 2022-09-23 at 4 21 41 PM

Screen Shot 2022-09-23 at 4 25 04 PM

### DIFF
--- a/src/components/core/table/table-colgroup.tsx
+++ b/src/components/core/table/table-colgroup.tsx
@@ -1,0 +1,22 @@
+import type { MinRecordItem, TableColumn } from "./table";
+
+export type TableColgroupProps<T extends MinRecordItem> = {
+  columns: TableColumn<T>[];
+};
+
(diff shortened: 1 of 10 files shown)


src/components/core/table/table-colgroup.tsx
```diff
@@ -0,0 +1,22 @@
...
+        key={column.title ?? index}
+        className={column.className}
+        style={{
+          minWidth: column.minWidth,
+          width: `${column.widthPercent}%`,
+        }}
+      />
+    ))}
+  </colgroup>
+);
```
luthfib, line 22: This feels nice

luthfib (approved):
LGTM

PR #81, opened 2022-09-27
Re add style.css as consumers still need to include it

WesleyKapow:
Also fixed bg of main conditions container.


pmadria3 (approved):
LGTM

PR #85, opened 2022-09-27
Don't use useLayoutEffect directly

WesleyKapow:
This fixes an issue with SSR.

> Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.

Followed https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a for work around.

### DIFF
--- a/.changeset/gentle-clouds-travel.md
+++ b/.changeset/gentle-clouds-travel.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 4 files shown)


luthfib (approved):
Neat solution. LGTM

PR #87, opened 2022-09-28
Ctw 353/update logic in condition display

luthfib:
This pr adds more codes to filter out duplicate conditions

TODO: 
- [x] Fix type issue in fhir/conditions.ts
- [x] Find more conditions to test or create conditions to test
- [x] Changeset

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,7 @@ const { VITE_SYSTEM_URL, VITE_AUTH_TOKEN, VITE_PATIENT_ID } = import.meta.env;
 
 function App() {
   return (
-    <CTWProvider env="dev" authToken={VITE_AUTH_TOKEN} theme={{}}>
+    <CTWProvider env="sandbox" authToken={VITE_AUTH_TOKEN} theme={{}}>
       <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
         <div className="App ctw-space-y-5">
           <h1>CTW Component Library</h1>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,7 @@ const { VITE_SYSTEM_URL, VITE_AUTH_TOKEN, VITE_PATIENT_ID } = import.meta.env;
 
 function App() {
   return (
-    <CTWProvider env="sandbox" authToken={VITE_AUTH_TOKEN} theme={{}}>
+    <CTWProvider env="dev" authToken={VITE_AUTH_TOKEN} theme={{}}>
       <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
         <div className="App ctw-space-y-5">
           <h1>CTW Component Library</h1>
--- a/src/fhir/conditions.ts
+++ b/src/fhir/conditions.ts
@@ -13,15 +13,16 @@ import {
   SYSTEM_ICD10,
   SYSTEM_ICD10_CM,
   SYSTEM_ICD9,
+  SYSTEM_ICD9_CM,
   SYSTEM_SNOMED,
 } from "./system-urls";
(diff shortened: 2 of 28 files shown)


src/App.tsx
```diff
@@ -7,7 +7,7 @@ const { VITE_SYSTEM_URL, VITE_AUTH_TOKEN, VITE_PATIENT_ID } = import.meta.env;
 
 function App() {
   return (
-    <CTWProvider env="dev" authToken={VITE_AUTH_TOKEN} theme={{}}>
+    <CTWProvider env="sandbox" authToken={VITE_AUTH_TOKEN} theme={{}}>
```
pmadria3, line 10: probably should use dev here

src/fhir/conditions.ts
```diff
@@ -98,3 +108,38 @@ function filterAndSort(conditions: fhir4.Condition[]) {
...
+
+  data.forEach((condition) => {
+    const conditionModel = new ConditionModel(condition);
+    ACCEPTABLE_CODES.forEach((code) => {
+      if (!(code in confirmedCodeDict)) {
+        confirmedCodeDict[code] = [];
+      }
+      if (typeof conditionModel[code] === "string") {
+        const modelCode = conditionModel[code];
+        confirmedCodeDict[code].push(modelCode as string);
```
pmadria3, line 123: nit: can just use conditionModel[code] directly instead of setting to modelCode

src/fhir/conditions.ts
```diff
@@ -98,3 +108,38 @@ function filterAndSort(conditions: fhir4.Condition[]) {
...
+export const filterDuplicateCodesFromTarget = (
+  target: fhir4.Condition[],
+  codesLookup: ConfirmedCodes
+) =>
+  target.filter((c) => {
+    const isDuplicate = ACCEPTABLE_CODES.some((code) => {
+      const conditionModel = new ConditionModel(c);
+      if (conditionModel[code]) {
+        const modelCode = conditionModel[code];
+        return codesLookup[code].includes(modelCode as string);
```
pmadria3, line 140: same nit as above

pmadria3 (dismissed):
Small nits but otherwise LGTM

pmadria3 (approved):
LGTM

WesleyKapow (changes requested):
Quick feedback -- It'd be useful if we could have an array of SYSTEM URLs of systems to use. Where all we need to do is edit this array to edit how things work. I think this means moving away from having explicit `icd9Code` and `icd10CMCode` getters on the model.

We're going to use this array in a few places. For matching and removing. And also for history lookup.

luthfib:
That makes sense. That's where my thought was with the codes array but an array of system urls would simplify it even more as we wouldnt have to add methods to the condition model

src/models/conditions.ts
```diff
@@ -71,6 +73,14 @@ export class ConditionModel {
     );
   }
 
+  get availableCodes(): fhir4.Coding[] {
```
pmadria3, line 76: change availableCodes into `allCodeSystems` or `allSystemCodes`?

src/models/conditions.ts
```diff
@@ -78,16 +73,12 @@ export class ConditionModel {
...
-    return findCoding(SYSTEM_ICD9, this.resource.code)?.code;
-  }
-
-  get icd9CMCode(): string | undefined {
-    return findCoding(SYSTEM_ICD9_CM, this.resource.code)?.code;
-  }
-
-  get icd10CMCode(): string | undefined {
-    return findCoding(SYSTEM_ICD10_CM, this.resource.code)?.code;
+  get allSystemCodings(): fhir4.Coding[] {
```
WesleyKapow, line 76: How about `knownCodings`?

WesleyKapow (approved):
Handful of naming nitpicks to make this a bit more clear, otherwise LGTM

WesleyKapow (approved):

PR #88, opened 2022-09-29
Ctw 354/manage inconsistent coding in history panel

luthfib:
Manage inconsistent coding in Conditions History Panel. Now supports SNOMED, ICD-10, ICD-10CM, ICD-9, ICD-9CM for code matching. Will now filter out summary and lens tags as well. Also now uses react query for history searches for faster performance when going back to already visited history searches.

### DIFF
--- a/src/components/content/conditions-history.tsx
+++ b/src/components/content/conditions-history.tsx
@@ -1,7 +1,8 @@
-import { getConditionHistory } from "@/fhir/conditions";
+import { ACCEPTABLE_CODES, getConditionHistory } from "@/fhir/conditions";
+import { useFhirClientRef } from "@/fhir/utils";
 import { ConditionModel } from "@/models/conditions";
+import { useQuery } from "@tanstack/react-query";
 import { useEffect, useState } from "react";
-import { useCTW } from "../core/ctw-provider";
 import {
   DataListStack,
   DataListStackEntries,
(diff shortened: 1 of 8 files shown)


src/components/content/conditions-history.tsx
```diff
@@ -12,110 +16,105 @@ import { Spinner } from "../core/spinner";
 
 const CONDITION_HISTORY_LIMIT = 10;
 
-export type ConditionHistoryProps = {
-  icd10Code?: string | undefined;
-  snomedCode?: string | undefined;
-};
+function setupData(condition: ConditionModel): DataListStackEntry {
```
WesleyKapow, line 19: Priyank will be changing this in his PR. For now can you revert this stuff to not do any changes w.r.t the codes we display in the history? I believe Priyank's PR takes care of looping over all of the codes we want to display via `knownCodes`.

src/components/content/conditions-history.tsx
```diff
@@ -12,110 +16,105 @@ import { Spinner } from "../core/spinner";
...
+  const fhirClientRef = useFhirClientRef();
   const { patientPromise } = usePatient();
+  const historyResponse = useQuery(
+    ["conditions", patientUPID, conditionTokens],
+    getConditionHistory,
+    {
+      enabled: !!patientUPID && !!fhirClientRef,
+      meta: { fhirClientRef },
+    }
+  );
```
WesleyKapow, line 91: Is it possible or reasonable to pass `condition` instead of `conditionTokens`? I'd prefer for that complexity to be captured within the request for `getConditionHistory`.

src/fhir/conditions.ts
```diff
@@ -38,9 +37,13 @@ export type ConditionFilters = {
   "clinical-status"?: ClinicalStatus | ClinicalStatus[];
 };
 
+export type QueryKeyConditionHistory = [string, string, string[]];
 export type QueryKeyPatientConditions = [string, string, ConditionFilters];
 export type QueryKeyOtherProviderConditions = [string, string];
 
+export const getConditionFilterTokens = (condition: ConditionModel) =>
+  condition.knownCodings.map((coding) => `${coding.system}|${coding.code}`);
+
```
WesleyKapow, line 46: Re capturing this complexity in `getConditionHistory`. This logic is VERY specific to fetching history. So a generic `getConditionFilterTokens` is weird (not pit of success) as it should only ever be used for one purpose.

WesleyKapow (changes requested):

WesleyKapow (approved):

PR #89, opened 2022-09-29
CTW-333: Filter out non-builder records using the builder ID in the JWT

melocule:
Components that were meant to display only records owned by the builder now address the case of shared records between builders. Rather than filtering out by non-builder tags, the query filters for the ownership tag of the specific builder.

### DIFF
--- a/src/fhir/client.ts
+++ b/src/fhir/client.ts
@@ -28,7 +28,7 @@ export type ZusJWT = {
   [key: string]: string | string[];
 };
 
-export function getJWT(fhirClient: Client): ZusJWT {
+export function getClaims(fhirClient: Client): ZusJWT {
   // Cast to access an unexposed property.
   const authHeader = (fhirClient as ClientAuth).httpClient.authHeader
     .authorization;
--- a/src/fhir/client.ts
+++ b/src/fhir/client.ts
@@ -25,8 +25,6 @@ export type ClientAuth = {
 } & Client;
 
 export type ZusJWT = {
-  SYSTEM_ZUS_BUILDER_ID: string;
-  // We only use the above properties. Others can be any key with any type.
   [key: string]: string | string[];
 };
 
--- a/src/fhir/client.ts
+++ b/src/fhir/client.ts
@@ -1,4 +1,5 @@
 import Client from "fhir-kit-client";
+import jwt_decode from "jwt-decode";
 
 import { Env } from "@/components/core/ctw-provider";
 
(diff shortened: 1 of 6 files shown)


src/fhir/search-helpers.ts
```diff
@@ -60,17 +63,17 @@ export async function searchAllRecords<T extends ResourceTypeString>(
...
   searchParams?: SearchParams
 ): Promise<SearchReturn<T>> {
-  const nonBuilderTags = [...THIRD_PARTY_TAGS, ...LENS_TAGS];
-
+  const jwt = getJWT(fhirClient);
+  const builderTag = `${SYSTEM_ZUS_OWNER}|builder/${jwt[SYSTEM_ZUS_BUILDER_ID]}`;
   return searchAllRecords(resourceType, fhirClient, {
     ...searchParams,
-    "_tag:not": nonBuilderTags.join(","),
+    _tag: builderTag,
```
luthfib, line 76: This doesnt look right to me. Might have to look around and ensure but I think you would create duplicates with this. You might be able to do something like: 

```
  _tag:not": nonBuilderTags.join(","),
  _tag: builderTag,
```

luthfib (changes requested):
Looking good but a couple of show stoppers

luthfib:
I also wonder if we want to get the auth from the FHIR client or if we want to allow the ctw provider to pass the auth token, as I think that will be a more generic approach that will decouple the FHIR client and the auth token.. thoughts @WesleyKapow

src/fhir/client.ts
```diff
@@ -8,13 +9,33 @@ export function getFhirClient(env: Env, accessToken: string) {
...
+
+  return client;
+}
+
+export type ZusJWT = {
+  SYSTEM_ZUS_BUILDER_ID: string;
+  // We only use the above properties. Others can be any key with any type.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
+};
```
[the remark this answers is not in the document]
melocule: There's a property in the JWT that's actually an array of strings, just realized I can type it as `string | string[]` though so I'm going with that!

src/fhir/search-helpers.ts
melocule: Thanks for the catch! Didn't see any resources with both, but apparently whether those tags overlaps can change with time.

WesleyKapow:
@luthfib agreed that getting it from FHIRClient is hacky. I'm just not thrilled with having to pass the accessToken around everywhere which is why I was opting for the hacky approach for now (no refactoring or plumbing of an argument needed).

Eventually we may need the `CTWProvider` to hand us something that we can pass around everywhere and hang stuff off of that. Things like `env`, `accessToken`, user details, etc.

src/fhir/client.ts
```diff
@@ -17,6 +18,28 @@ export function getFhirClient(env: Env, accessToken: string) {
...
 }
 
+export type ClientAuth = {
+  httpClient: {
+    authHeader: { authorization: string };
+  };
+} & Client;
+
+export type ZusJWT = {
+  SYSTEM_ZUS_BUILDER_ID: string;
```
WesleyKapow, line 28: Here you are saying this has a key of the literal string: `SYSTEM_ZUS_BUILDER_ID`.

src/fhir/client.ts
```diff
@@ -17,6 +18,28 @@ export function getFhirClient(env: Env, accessToken: string) {
...
+  // We only use the above properties. Others can be any key with any type.
+  [key: string]: string | string[];
+};
+
+export function getJWT(fhirClient: Client): ZusJWT {
```
WesleyKapow, line 33: s/getJWT/getClaims

The accessToken is the JWT, I believe you are decoding it and getting the claims data.

WesleyKapow (changes requested):

src/fhir/client.ts
```diff
@@ -17,6 +18,28 @@ export function getFhirClient(env: Env, accessToken: string) {
   });
 }
 
+export type ClientAuth = {
```
WesleyKapow, line 21: Need to be exported?

WesleyKapow (approved):
Side note -- avoid super-long-branch-names-that-describe-everything :P (it could have been something like `CTW-333/filter-by-owner-tag`)

WesleyKapow:
@melocule may need to resolve merge conflicts with main to get CI going.. not sure.. CI has been a little weird lately.

luthfib (approved):

PR #93, opened 2022-10-04
add logic to add recorder field

luthfib:
Still need to: 
- [x] Decide what we are going to show in the the display field
- [x] Decide what nulling a recording field looks like, or how to do that in the fhir post
- [x] Get my practitioner id to actually return a valid id, currently still get empty practitioner id despite the fact that I do have a practitioner associated with my role


src/components/content/forms/autocomplete-select-field.tsx
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    async function load() {
+      const response = await fetch(
+        "https://api.dev.zusapi.com/forms-data/terminology/conditions?display=b",
+        {
+          headers: { Authorization: `Bearer ${authToken}` },
+        }
+      );
+      const data = await response.json();
+      console.log("data", data);
```
pmadria3, line 14: remove console log

pmadria3 (approved):
LGTM

src/fhir/search-helpers.ts
```diff
@@ -71,8 +71,8 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
   searchParams?: SearchParams
 ): Promise<SearchReturn<T>> {
   const nonBuilderTags = [...THIRD_PARTY_TAGS, ...LENS_TAGS, SUMMARY_TAGS];
-  const jwt = getClaims(fhirClient);
-  const builderTag = `${SYSTEM_ZUS_OWNER}|builder/${jwt[SYSTEM_ZUS_BUILDER_ID]}`;
+  const claims = getClaims(fhirClient);
```
melocule, line 74: Nice rename! I missed this in my original PR.

src/components/content/forms/conditions.ts
```diff
@@ -53,10 +60,15 @@ export const createOrEditCondition = async (
     return result;
   }
 
+  const fhirClient = await getCTWFhirClient();
+  const practitionerId = result.data.id
+    ? (getClaims(fhirClient)[SYSTEM_PRACTITIONER_ID] as string)
```
melocule, line 65: Hey Luthfi, do you know if there's a way to avoid casting here? I thought to define some of the claims as part of the type definition, but wasn't sure since the property names are also system URLs

luthfib: We cast because it can return string | string[]. In this case we know that practioner id will always be a string

PR #97, opened 2022-10-04
Condition History Detail and Summary Spacing fix

pmadria3:
This PR fixes the spacing between detail and summary card. 
Screen Shot 2022-10-04 at 3 38 36 PM


luthfib (changes requested):
should be `ctw-space-y-3`

luthfib (approved):

PR #101, opened 2022-10-05
Ctw 392/fix patient overfetching

luthfib:
This pr: 

- Ensures that it only fetches when on the conditions component and not elsewhere. 
- We used to fetch everywhere because our patient provider is used in the top level of apps and when patientID, systemUrl or PatientUPID changed it fetched again, irregardless of whether they were in our conditions component. This is behavior, in my opinion we want to avoid. We should try to keep fetching behavior tied down to our components to avoid weird behavior like this. By switching to React Query and calling it in the component we stop that behavior, are able to cache calls to get patients, and if we reuse the same query anywhere else... say medications it will use the cache instead of making a new call.

Reasoning behind certain changes: 
- Only returning system URL and patientID from the provider. This way anywhere we usePatient we dont have to handle the case of UPID as our patient provider now handles that in the patientProvider. Only confusion could be that that it is called patientID instead of patientUPID, as it could technically be both. Im open for debate on this, and or suggestions.

### DIFF
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -117,10 +117,7 @@ export function Conditions({ className }: ConditionsProps) {
     async function load() {
       if (patientResponse.data) {
         setPatient(patientResponse.data);
-      }
-
-      if (patient?.UPID) {
-        setPatientUPID(patient.UPID);
+        setPatientUPID(patientResponse.data.UPID);
       }
 
       const tempConditionFilters: ConditionFilters = includeInactive
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -8,7 +8,6 @@ import {
   getOtherProviderConditions,
   getPatientConditions,
 } from "@/fhir/conditions";
-import { getPatient } from "@/fhir/patient";
 import { useFhirClientRef } from "@/fhir/utils";
 import { useBreakpoints } from "@/hooks/use-breakpoints";
 import { ConditionModel } from "@/models/conditions";
@@ -63,16 +62,7 @@ export function Conditions({ className }: ConditionsProps) {
     useState<FormEntry[]>();
   const [conditionForHistory, setConditionForHistory] =
     useState<ConditionModel>();
-  const { patientID, systemURL } = usePatient();
-
-  const patientResponse = useQuery(
-    ["patient", patientID, systemURL],
-    getPatient,
-    {
-      enabled: !!fhirClientRef,
-      meta: { fhirClientRef },
-    }
-  );
+  const patientResponse = usePatient();
 
   const patientRecordResponse = useQuery(
     ["conditions", patientUPID, conditionFilter],
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -53,7 +53,6 @@ export function Conditions({ className }: ConditionsProps) {
   const [OtherProviderRecordsMessage, setOtherProviderRecordsMessage] =
     useState(EMPTY_MESSAGE);
   const [patient, setPatient] = useState<PatientModel>();
-  const [patientUPID, setPatientUPID] = useState<string>("");
   const [includeInactive, setIncludeInactive] = useState(true);
   const [formAction, setFormAction] = useState("");
   const [conditionFilter, setConditionFilter] = useState<ConditionFilters>({});
@@ -65,19 +64,19 @@ export function Conditions({ className }: ConditionsProps) {
   const patientResponse = usePatient();
 
   const patientRecordResponse = useQuery(
-    ["conditions", patientUPID, conditionFilter],
+    ["conditions", patient?.UPID, conditionFilter],
     getPatientConditions,
     {
-      enabled: !!patientUPID && !!fhirClientRef,
+      enabled: !!patient && !!fhirClientRef,
       meta: { fhirClientRef },
     }
   );
 
   const OtherProviderRecordsResponse = useQuery(
-    ["conditions", patientUPID],
+    ["conditions", patient?.UPID],
     getOtherProviderConditions,
     {
-      enabled: !!patientUPID && !!fhirClientRef,
+      enabled: !!patient && !!fhirClientRef,
       meta: { fhirClientRef },
     }
   );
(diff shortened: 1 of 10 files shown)


src/components/core/patient-provider.tsx
```diff
@@ -43,27 +32,23 @@ export function PatientProvider({
...
-    setPatientPromise(getPatient());
-  }, [patientID, systemURL, patientUPID, getCTWFhirClient]);
+  }
+
+  const providerState = useMemo(
+    () => ({
+      patientID: patientUPID || patientID,
+      systemURL: patientUPID ? SYSTEM_ZUS_UNIVERSAL_ID : systemURL,
+    }),
+    [patientID, patientUPID, systemURL]
```
pmadria3, line 49: this would just be `[patientID, systemURL]` since you are setting the ID to be either patientID or patientUPID

pmadria3: oh right!

src/fhir/patient.ts
```diff
@@ -0,0 +1,26 @@
...
+    const response = await getBuilderFhirPatient(
+      fhirClient,
+      patientID,
+      systemURL
+    );
+    return response;
+  } catch (e) {
+    throw new Error(
+      `Failed fetching patient information for patient: ${patientID}`
+    );
```
WesleyKapow, line 24: Let's upgrade this error to also include the systemURL.

src/components/content/conditions.tsx
```diff
@@ -45,23 +45,35 @@ export function Conditions({ className }: ConditionsProps) {
...
   const [OtherProviderRecords, setOtherProviderRecords] = useState<
     ConditionModel[]
   >([]);
   const [OtherProviderRecordsIsLoading, setOtherProviderRecordsIsLoading] =
     useState(true);
   const [OtherProviderRecordsMessage, setOtherProviderRecordsMessage] =
     useState(EMPTY_MESSAGE);
-  const [includeInactive, setIncludeInactive] = useState(true);
   const [patient, setPatient] = useState<PatientModel>();
+  const [patientUPID, setPatientUPID] = useState<string>("");
```
WesleyKapow, line 57: Can probably refactor to `patient.UPID` instead of `patientUPID`. As in, if we have a real patient model, no need to also have patientUPID.

src/components/core/patient-provider.tsx
```diff
@@ -43,27 +32,23 @@ export function PatientProvider({
...
-        const fhirClient = await getCTWFhirClient();
-        return getBuilderFhirPatient(
-          fhirClient,
-          patientID ?? patientUPID,
-          systemURL ?? SYSTEM_ZUS_UNIVERSAL_ID
-        );
-      }
-      // This should not actually be possible.
-      throw new Error("Patient UPID or patient id/system url was not defined.");
+  // Case to catch if user doesnt provide either of the
```
WesleyKapow, line 35: s/doesnt/doesn't

And finish comment :)

WesleyKapow: Actually just remove this, the type guard should be enough here.

src/components/core/patient-provider.tsx
```diff
@@ -43,27 +32,23 @@ export function PatientProvider({
...
+      );
     }
-    setPatientPromise(getPatient());
-  }, [patientID, systemURL, patientUPID, getCTWFhirClient]);
+  }
+
+  const providerState = useMemo(
+    () => ({
+      patientID: patientUPID || patientID,
+      systemURL: patientUPID ? SYSTEM_ZUS_UNIVERSAL_ID : systemURL,
```
WesleyKapow, line 47: Typescript isn't smart enough to know that `patientID` and `systemURL` cannot be `undefined` due to the `||` and ternary. Add ` as ProviderState` to the end of this to get `providerState` treated as the correct thing.

src/components/content/conditions.tsx
```diff
@@ -45,23 +45,35 @@ export function Conditions({ className }: ConditionsProps) {
...
+  const { patientID, systemURL } = usePatient();
+
+  const patientResponse = useQuery(
+    ["patient", patientID, systemURL],
+    getPatient,
+    {
+      enabled: !!fhirClientRef,
+      meta: { fhirClientRef },
+    }
+  );
```
WesleyKapow, line 75: Refactor this into `usePatient`? Many components will need to fetch the patient. Would be nice if `usePatient` did this for us and returned us a patient model.

WesleyKapow (changes requested):
Few minor feedbacks, LGTM.

src/components/core/patient-provider.tsx
```diff
@@ -50,12 +45,22 @@ export function PatientProvider({
...
+  const { patientID, systemURL } = useContext(CTWPatientContext);
+  const patientResponse = useQuery(
+    ["patient", patientID, systemURL],
+    getPatient,
+    {
+      enabled: !!fhirClientRef,
+      meta: { fhirClientRef },
+    }
+  );
+  return patientResponse;
```
WesleyKapow, line 65: What about making `usePatient` `async` and then having it process the response and wrap it in a patient model. That way consumers can do `const patient = await userPatient()`? Does that make things cleaner?

luthfib: This is a tricky one. Because in order to be able to use the hooks inside the usePatient we need to be able to keep it as a hook. And hooks aren't really async, meaning that if we wanted to await it we would then have to be put in the useEffect but if we awaited it then we couldnt use the hooks we use in the FHIR client.

src/components/content/conditions.tsx
luthfib: Thinking about this one. The tricky part is that in the patient model patient.upid can return a string or undefined, which is true. Which should have upid for all patients though...

WesleyKapow: ^ Or better yet, log an error if it can't get the UPID.

src/components/core/patient-provider.tsx
WesleyKapow: Hmmm ok leave as is. I don't understand react query enough yet to weigh in.

src/components/content/conditions.tsx
luthfib: so patientUPID did return a string my bad. Issue was patient can be undefined. Set so that useQuery takes `patient.UPID` as suggested but had to change the param of some queries to take string or undefined and catch if it is given an undefined UPID. with enabled we are able to to ensure `enabled: !!patient && !!fhirClientRef,` that patient is always defined, but I guess react query cant infer that, or I dont know how to ensure it infers those enabled params

WesleyKapow (approved):
LGTM

PR #102, opened 2022-10-05
Primarily alphabetize by CCS Grouping

melocule:
Alphabetize conditions primarily by CCS grouping and secondarily the condition name so that users can quickly find the conditions by their category.

Example:
image


melocule:
Fixed sorting:
image

PR #104, opened 2022-10-05
CTW-370/patientid-guard

pmadria3:
This PR adds a guard for if the patientID doesn't exist in Zus. 

Screen Shot 2022-10-06 at 8 58 11 AM


src/components/core/alert.tsx
```diff
@@ -0,0 +1,15 @@
...
+      <img src={alert} alt="alert" /> Conditions Unavailable
+    </div>
+    <div className="ctw-text-caution-lightred">
+      We are unable to access Condition information for this patient.
+    </div>
+    <div className="ctw-text-caution-lightred">
+      Contact your system administrator or customer service for assistance.
+    </div>
+  </div>
+);
```
luthfib, line 15: I imagine we are going to be using this AlertDialog in other components so it should be generic and take in the message it should

.changeset/silent-squids-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Don't fetch when not on patient provider. Also fixes overfetching
```
luthfib, line 5: Fix changeset, this is not the correct description

luthfib:
Nit: branch names should be projectPrefix-ProjectNumber/description-of-project

luthfib (changes requested):
Looking good, but some changes needed. Need padding around the alert dialog. Refer to figma and compare it yours.

src/components/core/alert.tsx
luthfib: Alert Icon and the title need to be aligned. Also the paragraph should not appear right below the alert icon but to the side. Please refer to figma

luthfib (commented):
I believe we dont need to add a new icon but can exclamation-triangle from heroicons.

src/components/content/conditions.tsx
```diff
@@ -183,79 +185,83 @@ export function Conditions({ className }: ConditionsProps) {
...
+          <button
+            type="button"
+            className="ctw-btn-clear ctw-link"
+            onClick={handleAddNewCondition}
+          >
+            + Add Condition
+          </button>
+        )}
       </div>
+      {patientResponse.error && <AlertDialog />}
```
luthfib, line 198: Need a fragment around this to get rid of the error

luthfib, line 198: you can do patientResponse.isError for more succint error handling

src/components/content/conditions.tsx
```diff
@@ -183,79 +184,92 @@ export function Conditions({ className }: ConditionsProps) {
...
-          <div className="ctw-conditions-title-container">
-            <div className="ctw-title">Patient Record</div>
-            <ToggleControl
-              onFormChange={handleFormChange}
-              toggleProps={{ name: "conditions", text: "Include Inactive" }}
+      <>
+        {patientResponse.isError && (
+          <div className="ctw-p-5">
+            <AlertDialog
+              resourceType="Condition"
```
luthfib, line 201: Doesnt this need to be plural?

pmadria3:
Going with the design team icon after discussion

luthfib (changes requested):
These still do not match. Look at the aligement with the icon. Align center should the the tricky if the icon and the title are in the same flex div. Also the content shouldnt be below the alert icon

image


image

src/styles/tailwind.theme.ts
```diff
@@ -45,6 +45,8 @@ export const TailwindTheme = {
     caution: {
       main: "#F59E0B",
       light: "#FEF3C7",
+      burgundy: "#92400E",
```
luthfib, line 48: Still checking if we want to add new colors, if we do I think we could change these to be:

```
burgundy/text
lightred/body
yellow/default
```

The current names dont have too much of a semantic meaning

src/components/content/conditions.tsx
```diff
@@ -183,79 +184,92 @@ export function Conditions({ className }: ConditionsProps) {
...
       <div className="ctw-flex ctw-h-11 ctw-items-center ctw-justify-between ctw-bg-bg-light ctw-p-3">
         <div className="ctw-title">Conditions</div>
-        <button
-          type="button"
-          className="ctw-btn-clear ctw-link"
-          onClick={handleAddNewCondition}
-        >
-          + Add Condition
-        </button>
+        {!patientResponse.isError && (
```
WesleyKapow, line 187: Might be easier to have do:

```ts
if (patientResponse.isError) {
  return (<...>)
}

return ( ...) // The existing stuff.
```

src/components/core/alert.tsx
```diff
@@ -0,0 +1,19 @@
...
+export type AlertDialogProps = {
+  resourceType: string;
+  children: ReactNode;
+};
+
+export const AlertDialog = ({ resourceType, children }: AlertDialogProps) => (
+  <div className="ctw-flex ctw-items-start ctw-space-x-2 ctw-bg-caution-bg ctw-p-4">
+    <img src={alert} alt="alert" className="ctw-shrink-0" />
+    <div className="ctw-space-y-2">
+      <div className="ctw-flex ctw-text-caution-heading">
```
WesleyKapow, line 13: flex is usually not needed when there's only one child.. flex box is about how elements line up in a row or column, one child = nothing to do.

src/components/core/alert.tsx
```diff
@@ -0,0 +1,19 @@
+import { ReactNode } from "react";
+import alert from "./alert.svg";
+
+export type AlertDialogProps = {
+  resourceType: string;
+  children: ReactNode;
+};
+
+export const AlertDialog = ({ resourceType, children }: AlertDialogProps) => (
```
WesleyKapow, line 9: You've made this alert component specific to resource unavailable errors. We may want to show it for other things too.

.changeset/silent-squids-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Adds an error display to the conditions table if there is no patientID corresponding in Zus Health.
```
WesleyKapow, line 5: Maybe change to `... conditions table if patient could not be found".

WesleyKapow (approved):

PR #105, opened 2022-10-06
Cleanup

luthfib:
cleanup code and fix error where fhir client throws error intermittently. Fixed by referencing `fhirClientRef.current` instead of `fhirClientRef`

### DIFF
--- a/.changeset/silly-turkeys-speak.md
+++ b/.changeset/silly-turkeys-speak.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+General code cleanup. Remove case where FhirClient must be provided error was thrown
--- a/.changeset/silly-turkeys-speak.md
+++ b/.changeset/silly-turkeys-speak.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-General code cleanup. Remove case where FhirClient must be provided error was thrown
+General code cleanup. Remove case where FhirClient must be provided error was thrown.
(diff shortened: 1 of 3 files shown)


src/components/content/conditions.tsx
```diff
@@ -63,35 +53,43 @@ export function Conditions({ className }: ConditionsProps) {
     useState<ConditionModel>();
   const patientResponse = usePatient();
```
WesleyKapow, line 54: Any way to get this to just be `patient`? I'd really like to see things like `patient?.UPID` instead of `patientResponse.data?.UPID`

.changeset/silly-turkeys-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+General code cleanup. Remove case where FhirClient must be provided error was thrown
```
WesleyKapow, line 5: nit: Sentences end with a period.

WesleyKapow (dismissed):
Left a couple comments but LGTM.

pmadria3 (dismissed):
LGTM

src/components/content/conditions.tsx
luthfib: Since its a hook, not sure how we would await especially since hooks below depend on the state of it. Something worth exploring how we could await some of this stuff but I think that might require a discussion. I think for purposes of this pr I'd advocate to leave that out as im unsure how we would achieve that currently

pmadria3: @luthfib any luck here? If this is resolved, I can take a peek at the PR

pmadria3 (approved):
LGTM

src/components/content/conditions.tsx
WesleyKapow: OK sounds good. Hopefully we can circle back on all of this react-query and usePatient stuff to get a simpler interface going.

WesleyKapow (approved):

PR #107, opened 2022-10-06
Add sorting to condition history page

pmadria3:
This PR just updates sorting for the history panel page to now sort by date and then move the empty dates to the end. 
Screen Shot 2022-10-06 at 11 43 44 AM

### DIFF
--- a/.changeset/dull-taxis-attack.md
+++ b/.changeset/dull-taxis-attack.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Condition History page now sorts by date and the entries without a date are filtered to the end.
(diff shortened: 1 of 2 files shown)


.changeset/dull-taxis-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Condition History page now sorts by date and the entries without a date are filtered to the end.
```
WesleyKapow, line 5: Remove "page" as it is not a page.

WesleyKapow (approved):
LGTM

luthfib (approved):
LGTM. While we are there can you change 

```
const filteredConditions = historyResponse.data.conditions.map(
          (c) => new ConditionModel(c, includedResources)
        );
```

to not say `filteredConditions` anymore... how about just `conditions`?

luthfib (approved):

PR #109, opened 2022-10-06
Add Records with no date in a separate format in condition history

pmadria3:
This PR adds a separate section for records with no date. If there is no entry, then no section will show up. 

With Both Date and No Date: 
Screen Shot 2022-10-06 at 2 16 13 PM

With only Dates:
Screen Shot 2022-10-06 at 2 19 17 PM

With only no Dates: 
Screen Shot 2022-10-06 at 2 20 09 PM

### DIFF
--- a/.changeset/quiet-eyes-buy.md
+++ b/.changeset/quiet-eyes-buy.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Condition History has a header that subdivides the condition entries to sort by date otherwise group the entries with no dates.
--- a/.changeset/quiet-eyes-buy.md
+++ b/.changeset/quiet-eyes-buy.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Condition History has a header that subdivides the condition entries to sort by date otherwise group the entries with no dates.
+Add “Records with no date” section to condition history.
--- a/src/components/content/conditions-history.tsx
+++ b/src/components/content/conditions-history.tsx
@@ -133,8 +133,8 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
   function conditionHistoryDisplay() {
     if (
       conditionsWithDate.length === 0 &&
-      !loading &&
-      conditionsWithoutDate.length === 0
+      conditionsWithoutDate.length === 0 &&
+      !loading
     ) {
       return <div>No history found.</div>;
     }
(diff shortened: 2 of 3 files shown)


.changeset/quiet-eyes-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Condition History has a header that subdivides the condition entries to sort by date otherwise group the entries with no dates.
```
WesleyKapow, line 5: This is a little wordy and confusing. Pretend you are talking to the customer and keep it simple if possible.

src/components/content/conditions-history.tsx
```diff
@@ -102,21 +103,39 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
       setLoading(true);
     };
   }, [condition, patientResponse.data, historyResponse.data]);
 
   function conditionHistoryDisplay() {
-    if (conditions.length === 0 && !loading) {
+    if (
+      conditionsWithDate.length === 0 &&
+      !loading &&
+      conditionsWithoutDate.length === 0
```
WesleyKapow, line 137: nit: Put the two `=== 0 ` checks next to each other.

src/components/content/conditions-history.tsx
```diff
@@ -138,9 +158,18 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
           <div className="ctw-text-sm">{condition.ccsGrouping}</div>
         </div>
         <CollapsibleDataListStack
-          entries={conditions}
+          entries={conditionsWithDate}
           limit={CONDITION_HISTORY_LIMIT}
         />
+        <div className="ctw-space-y-2 ctw-text-base">
+          {conditionsWithoutDate.length !== 0 && (
```
luthfib, line 165: Please increase the font weight to 500

luthfib: Also please match indentation to the figma file for "Records with no date:"

src/components/content/conditions-history.tsx
```diff
@@ -138,9 +158,18 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
+            <div className="ctw-space-y-2">Records with no date:</div>
```
WesleyKapow, line 166: Spacing classes are used to add space between children elements. This only has one child and so will do nothing.

.changeset/quiet-eyes-buy.md
melocule, line 5: Should this be a patch or a minor? I read that patches are for bug fixes and minors are for backward-compatible new features

WesleyKapow: Ya I haven't really pinned down my thinking on this yet. What you read is my understanding as well. We can play this a little loose a.t.m. For most libraries they are able to release patch updates for older versions. We're not setup do that at the moment. We really have linear updates right now (less than ideal but fine for now).

pmadria3: I'll leave it as a patch for now? And then I guess moving forward if it's anything more than a bug fix maybe make it a minor? Just for some consistency.

src/components/content/conditions-history.tsx
```diff
@@ -161,14 +161,18 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
           entries={conditionsWithDate}
           limit={CONDITION_HISTORY_LIMIT}
         />
-        <div className="ctw-space-y-2 ctw-text-base">
+        <div className="ctw-space-y-2">
```
WesleyKapow, line 164: Do we want this div when we are not showing the no date section below?

src/components/content/conditions-history.tsx
```diff
@@ -138,9 +158,22 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
           <div className="ctw-text-sm">{condition.ccsGrouping}</div>
         </div>
         <CollapsibleDataListStack
-          entries={conditions}
+          entries={conditionsWithDate}
           limit={CONDITION_HISTORY_LIMIT}
         />
+        {conditionsWithoutDate.length !== 0 && (
+          <>
```
WesleyKapow, line 165: And now this fragment is not needed.

pmadria3: ah good catch

WesleyKapow (approved):

PR #110, opened 2022-10-06
Ctw 328/add practitioner fetch

luthfib:
Add display when saving the recorder field for condition.

### DIFF
--- a/src/components/content/forms/conditions.ts
+++ b/src/components/content/forms/conditions.ts
@@ -1,4 +1,6 @@
+import { createOrEditFhirResource } from "@/fhir/action-helper";
 import { getClaims } from "@/fhir/client";
+import { isFhirError } from "@/fhir/errors";
 import { dateToISO } from "@/fhir/formatters";
 import { getPractitioner } from "@/fhir/practitioner";
 import {
@@ -8,6 +10,7 @@ import {
   SYSTEM_SNOMED,
 } from "@/fhir/system-urls";
 import { ConditionModel } from "@/models/conditions";
+import { PractitionerModel } from "@/models/practitioner";
 import { getFormData } from "@/utils/form-helper";
 import { queryClient } from "@/utils/request";
 import Client from "fhir-kit-client";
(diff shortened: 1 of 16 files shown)


src/fhir/practitioner.ts
```diff
@@ -0,0 +1,21 @@
...
+      "Practitioner",
+      fhirClient,
+      {
+        _id: practitionerId,
+      }
+    );
+
+    return practitioners;
+  } catch (e) {
+    throw new Error(`Failed fetching practitioner: ${e}`);
```
pmadria3, line 19: For this error maybe we can also include the practitionerID as well

src/fhir/practitioner.ts
```diff
@@ -0,0 +1,23 @@
+import Client from "fhir-kit-client";
+import { searchBuilderRecords } from "./search-helpers";
+
+export const getPractitioner = async (
```
WesleyKapow, line 4: Change to return only one and error if 0 found.

WesleyKapow (changes requested):
Minor comments, LGTM

pmadria3 (approved):
LGTM

src/fhir/practitioner.ts
```diff
@@ -14,7 +14,11 @@ export const getPractitioner = async (
       }
     );
 
-    return practitioners;
+    if (practitioners.length === 0) {
+      throw new Error(`No practitioner found with an id of: ${practitionerId}`);
+    }
```
WesleyKapow, line 19: I've hit this tricky problem before. This `throw` will get caught in the below `catch`.

WesleyKapow (changes requested):

src/fhir/practitioner.ts
luthfib: You are right, good catch.

WesleyKapow (dismissed):
LGTM, may want to push the use of `new PractitionerModel` into the helper as well. I think it makes sense for our fhir helpers to just return the models since that's mostly what we want to be working with.

src/components/content/forms/conditions.ts
```diff
@@ -48,8 +47,7 @@ export const conditionSchema = z.object({
 });
 
 const setRecorderField = async (practitionerId: string, fhirClient: Client) => {
-  const practitioner = await getPractitioner(practitionerId, fhirClient);
-  const practitionerModel = new PractitionerModel(practitioner);
+  const practitionerModel = await getPractitioner(practitionerId, fhirClient);
   const display = practitionerModel.fullName;
```
WesleyKapow, line 51: can simplify to just calling it a `practitioner` and not model.

WesleyKapow (approved):

PR #113, opened 2022-10-07
CTW-381 Automatic Fields Conditions

melocule:
Simplify the user experience of adding a new condition by pre-populating some fields with practical default values. Additionally hides the reported date field, as the now pre-populated value of today's date should not be tampered with by the user.

### DIFF
--- a/.changeset/breezy-cobras-remember.md
+++ b/.changeset/breezy-cobras-remember.md
@@ -1,5 +1,0 @@
----
-"@zus-health/ctw-component-library": patch
----
-
-Alphabetize conditions primarily by CCS grouping and secondarily the condition name so that users can quickly find the conditions by their category.
--- a/.changeset/dull-taxis-attack.md
+++ b/.changeset/dull-taxis-attack.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Condition History now sorts by date and the entries without a date are filtered to the end.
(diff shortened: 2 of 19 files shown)


WesleyKapow (changes requested):
LGTM, what about updating `recordedDate` when editing a condition OR when adding from "other provider records"? Is that happening?

melocule:
> LGTM, what about updating `recordedDate` when editing a condition OR when adding from "other provider records"? Is that happening?

Hey thanks for bringing this up, I missed that the note at the bottom of the ticket was part of the AC. Changed it to update the recorded date whenever a condition is edited! Also confirmed that the recordedDate is up to date when adding from other provider records.

luthfib (commented):
Did you verify that this works? Not sure what dateToISO does to an already dateToISO date, because in `forms/condition.ts` we do `dateToISO(result.data.recordedDate)`. So theoretically we are doing  `dateToISO(dateToISO(new date())`

melocule:
> Did you verify that this works? Not sure what dateToISO does to an already dateToISO date, because in `forms/condition.ts` we do `dateToISO(result.data.recordedDate)`. So theoretically we are doing `dateToISO(dateToISO(new date())`

It works, but it was redundant-- wdy think of this solution? I removed the reportedDate field from the form and just set it to today's date right before sending the request.

WesleyKapow (approved):
LGTM, love the solution for recordedDate!

melocule:
Thanks Luthfi and Wesley!

PR #114, opened 2022-10-07
Fix theming issue with drawer + dropdown menu

WesleyKapow:
* Added manual headlessui-portal-root to get drawer added within CTWProvider which defines theme CSS vars.
* Added a CTWProviderRef to the useCTW context and explicitly told dropdown-menu to use that for portal placement.
* Added custom theme example to App.tsx
* Fixed a few minor theming issues with colors in a few spots.

Screen Shot 2022-10-07 at 3 24 25 PM
Screen Shot 2022-10-07 at 3 24 43 PM

Screen Shot 2022-10-07 at 3 25 19 PM

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,12 +2,31 @@ import "./App.css";
 import { Conditions } from "./components/content/conditions";
 import { CTWProvider } from "./components/core/ctw-provider";
 import { PatientProvider } from "./components/core/patient-provider";
+import { Theme } from "./styles/tailwind.theme";
 
 const { VITE_SYSTEM_URL, VITE_AUTH_TOKEN, VITE_PATIENT_ID } = import.meta.env;
 
+const theme: Theme = {
+  colors: {
+    primary: {
+      light: "#F2F6FD",
+      main: "#5C8EDC",
+      dark: "#3E6197",
+    },
+    content: {
+      black: "#1A2848",
+      light: "#4A4A4A",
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -2,12 +2,31 @@ import "./App.css";
...
+      black: "#1A2848",
+      light: "#4A4A4A",
+      lighter: "#999999",
+    },
+    bg: {
+      light: "#F1F1F1",
+    },
+  },
+};
+
```
luthfib, line 26: This is good, that way we have a visible example in our library of how it's done

src/App.tsx
```diff
@@ -2,12 +2,31 @@ import "./App.css";
...
 function App() {
   return (
-    <CTWProvider env="dev" authToken={VITE_AUTH_TOKEN} theme={{}}>
+    <CTWProvider env="dev" authToken={VITE_AUTH_TOKEN} theme={theme}>
```
luthfib, line 29: I wonder if at some point we want env to be dictated by the .env file.

luthfib (approved):
LGTM, left a minor comment that we may not even want to address, just a thought

src/App.tsx
WesleyKapow: Ya not a bad idea. The whole auth and env stuff will likely change when we setup storybook. Gonna leave as is for now.

PR #118, opened 2022-10-17
CTW-382/conditions-quick-view

pmadria3:
This PR adds an optional quick view functionality to provide a read only view for the conditions table. There is no ability to add or edit in this view.
Screen Shot 2022-10-18 at 12 26 30 PM

### DIFF
--- a/.changeset/perfect-countries-hear.md
+++ b/.changeset/perfect-countries-hear.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Adjusted the quick view for the conditions table to be more narrow. User can view the Patient Record and Other Provider Records in the quick profile view.
+Added an optional quick view functionality to provide a read only view for the conditions table.
--- a/.changeset/perfect-countries-hear.md
+++ b/.changeset/perfect-countries-hear.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Adjusted the quick view for the conditions table to be more narrow. User can view the Patient Record and Other Provider Records in the quick profile view.
--- a/.changeset/perfect-countries-hear.md
+++ b/.changeset/perfect-countries-hear.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Added an optional quick view functionality to provide a read only view for the conditions table.
+Added an optional readyOnly functionality to provide a read only view for the conditions table. If this readOnly functionality is on, the meatball menu will be hidden.
--- a/.changeset/perfect-countries-hear.md
+++ b/.changeset/perfect-countries-hear.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Added an optional readyOnly functionality to provide a read only view for the conditions table. If this readOnly functionality is on, the meatball menu will be hidden.
+Add optional readOnly property to conditions. This will hide the “…” menu and the “add conditions” link.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,7 @@ function App() {
       <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
         <div className="App ctw-space-y-5">
           <h1>CTW Component Library</h1>
-          <Conditions />
+          <Conditions quickProfile={false} />
         </div>
       </PatientProvider>
     </CTWProvider>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,7 @@ function App() {
       <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
         <div className="App ctw-space-y-5">
           <h1>CTW Component Library</h1>
-          <Conditions quickProfile={false} />
+          <Conditions />
         </div>
       </PatientProvider>
     </CTWProvider>
--- a/src/components/content/conditions-table-base.tsx
+++ b/src/components/content/conditions-table-base.tsx
@@ -1,6 +1,5 @@
 import { Table, TableBaseProps, TableColumn } from "../core/table/table";
 
-import { Breakpoints } from "@/hooks/use-breakpoints";
 import { ConditionModel } from "@/models/conditions";
 import { DotsHorizontalIcon } from "@heroicons/react/outline";
 import { DropdownMenu, MenuItems } from "../core/dropdown-menu";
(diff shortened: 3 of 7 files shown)


src/styles/tailwind.theme.ts
```diff
@@ -3,6 +3,7 @@ import { Subset } from "../utils/typescript";
 export const CLASS_PREFIX = "ctw-";
 
 export const defaultBreakpoints = {
+  esm: 320,
```
pmadria3, line 6: wasn't sure what to call this esm stands for extra small but am open to renaming this

luthfib (changes requested):
image

Patient Record should be above the inactive filter as show in the desired behavior for the quick view

pmadria3:
For the pill, design team said to not have that and forgot to remove it from the design (put it in the out-of-scope as well)

luthfib:
Additionally I don't see a prop to turn quick profile on and off, meaning that we have by default set the quick profile view on, which I dont think is the desired behavior.

WesleyKapow:
This needs to get reworked. Quick view is not something that happens due to window/container size. It is instead something that is explicitly used by healthie via a new component or prop(s) on the existing component.

src/App.tsx
```diff
@@ -30,7 +30,7 @@ function App() {
       <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
         <div className="App ctw-space-y-5">
           <h1>CTW Component Library</h1>
-          <Conditions />
+          <Conditions quickProfile={false} />
```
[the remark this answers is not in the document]
melocule: Smth like this https://stackoverflow.com/questions/37828543/how-to-pass-props-without-value-to-component

src/components/content/conditions.tsx
```diff
@@ -30,13 +30,17 @@ import {
 
 export type ConditionsProps = {
   className?: string;
+  quickProfile?: boolean;
```
melocule, line 33: Hey looks great! Just one thing: Since you're setting `quickProfile = false` in the Conditions function, `quickProfile` doesn't have to be optional here

melocule (changes requested):

src/components/content/conditions.tsx
```diff
@@ -174,32 +178,40 @@ export function Conditions({ className }: ConditionsProps) {
...
+        </div>
+      )}
       <div className="ctw-conditions-body">
         <div className="ctw-space-y-3">
-          <div className="ctw-conditions-title-container">
+          <div
+            className={
+              !breakpoints.esm
+                ? "ctw-conditions-title-container"
+                : "ctw-flex ctw-flex-col ctw-items-start ctw-space-y-2 ctw-p-3"
```
WesleyKapow, line 199: Maybe refactor to a `ctw-conditions-title-container-readonly` .. actually I'm not sure what the differences here are. What particularly is needed for this view here?

WesleyKapow (changes requested):
Please update PR description with latest screenshot!

src/styles/tailwind.theme.ts
melocule: How about `xs`? Tailwind uses this to refer to extra small in [their docs](https://tailwindcss.com/docs/screens)

src/components/content/conditions.tsx
pmadria3: I do need the question mark otherwise will need to make it a param in App.tsx

src/components/content/conditions.tsx
```diff
@@ -174,32 +178,40 @@ export function Conditions({ className }: ConditionsProps) {
...
+            + Add Condition
+          </button>
+        </div>
+      )}
       <div className="ctw-conditions-body">
         <div className="ctw-space-y-3">
-          <div className="ctw-conditions-title-container">
+          <div
+            className={
+              !breakpoints.esm
```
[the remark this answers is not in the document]
WesleyKapow: Don 't use responsive breakpoints for this, instead use `flex-wrap` and setup whatever `space-y` is needed. This should probably get added to the existing `ctw-conditions-title-container` class.

The idea is that we aren't saying "these should stack when below 320px", what we really want to say is "these should stack when there isn't enough room" and normal CSS wrapping can handle that.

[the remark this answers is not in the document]
WesleyKapow: Oh actually spacing will be a little trickier and will likely need both a `column-gap` and a `row-gap` so that X and Y spacing is only applied when needed.

src/components/content/conditions-table-base.tsx
```diff
@@ -8,12 +8,14 @@ export type ConditionsTableBaseProps = {
   className?: string;
   conditions: ConditionModel[];
   rowActions: (condition: ConditionModel) => MenuItems[];
+  hideMenu: boolean;
```
WesleyKapow, line 11: Make this optional and default to `false`.

.changeset/perfect-countries-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Added an optional quick view functionality to provide a read only view for the conditions table.
```
WesleyKapow, line 5: Renamed quick view.. change to `readOnly` and `hideMenu`?

WesleyKapow (changes requested):

.changeset/perfect-countries-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Added an optional readyOnly functionality to provide a read only view for the conditions table. If this readOnly functionality is on, the meatball menu will be hidden.
```
WesleyKapow, line 5: Makes it sound like meatball menu being hidden is the only thing result.

src/components/content/conditions.scss
```diff
@@ -20,7 +20,7 @@
   }
 
   .ctw-conditions-title-container {
-    @apply ctw-m-3 ctw-mb-0;
+    @apply ctw-m-3 ctw-mb-0 ctw-space-y-2;
```
WesleyKapow, line 23: You'll likely want column and row gaps on the `title-container itself, and not just when it is stacked.

src/components/content/conditions.tsx
```diff
@@ -193,7 +196,6 @@ export function Conditions({ className }: ConditionsProps) {
               toggleProps={{ name: "conditions", text: "Include Inactive" }}
             />
           </div>
-
```
WesleyKapow, line 196: Sometimes blank newlines are nice to separate chunks of functionality. Did you feel strongly against them?

src/components/content/conditions.tsx
```diff
@@ -217,7 +219,6 @@ export function Conditions({ className }: ConditionsProps) {
             ]}
           />
         </div>
-
```
WesleyKapow, line 220: Same here.

WesleyKapow (changes requested):

WesleyKapow (approved):

PR #120, opened 2022-10-18
Ctw 284/autocomplete conditions

luthfib:
- Adds autocomplete generic component
- Adds autocomplete wrapper for conditions
- Slight rename I tackled in table component 

Still need answered: 

- ICD-10 code is saved to the code, and ICD-10 system is specified. (I thought this would be done by lens enriching being that we fill the snomed code from the autocomplete, but have to verify)


src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const options = conditions.data?.map(
+    (condition: ConditionsAutoCompleteOption) => ({
+      value: condition.display,
+      id: condition.code,
+    })
+  );
+
+  useEffect(() => {
+    /* This handles the case where a user goes to an existing condition that has already been added but doesn't exist
+       in our search, so we clear the input so that the user does not try adding a condition that doesnt exist. */
```
luthfib, line 46: Might be able to handle this in zod, but have to play around with it

src/components/content/forms/condition-schema.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <ConditionsAutoComplete readonly={readonly} {...inputProps} />
+    ),
+  },
+  {
+    label: "subject",
+    value: condition.subjectID,
+    field: "subjectID",
+    readonly: true,
+    hidden: true,
+  },
```
WesleyKapow, line 56: This looks shared as well.

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    return debounce(handleSearchInputChange, 300);
+  }, [setQuery]);
+
+  useEffect(
+    () => () => {
+      debouncedSearchInputChange.cancel();
+    },
+    [query, debouncedSearchInputChange]
+  );
```
WesleyKapow, line 37: Add some comments and a newline before the `return`

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,107 @@
...
+    return <ComboboxOption option={{ value: "Type to search", id: "empty" }} />;
+  }
+
+  if (query.length < 2) {
+    return (
+      <ComboboxOption
+        option={{ value: "No choices to choose from", id: "empty" }}
+      />
+    );
+  }
```
WesleyKapow, line 72: What's this one doing? Doesn't look like you are guarding the API request until user types 3 chars. And if that's the intent, this message doesn't guide the user.

src/components/content/forms/condition-schema.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  conditionSystem: z.string({
+    required_error: "Condition system must be specified.",
+  }),
+  conditionCode: z.string({
+    required_error: "Condition code must be specified.",
+  }),
+  condition: z.string({
+    required_error: "Condition name must be specified.",
+  }),
```
WesleyKapow, line 103: How about `condition: {system, code, display}` instead of these three separate things. The field name would then be `condition.system`, etc. Check out zod for how to represent that.

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { authToken, env } = useCTW();
+  const [query, setQuery] = useState((inputProps.defaultValue as string) || "");
+  const [selectedCondition, setSelectedCondition] = useState<
+    ConditionsAutoCompleteOption | undefined
+  >({
+    display: "",
+    code: "",
+    system: "",
+  });
+  const conditions = useQuery(
```
WesleyKapow, line 31: Curious if we should ALWAYS use `useQuery` or if we should only reach for it in certain situations. It's nice that it can cache results but the interface is def. more complex than a native `fetch` request.

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+        name={`${inputProps.name}Code`}
+        defaultValue={selectedCondition?.code}
+        hidden
+      />
+      <input
+        name={`${inputProps.name}System`}
+        value={selectedCondition?.system}
+        hidden
+        readOnly
+      />
```
WesleyKapow, line 84: Autocomplete for codings is likely to be a common pattern. Maybe refactor this to something that takes name like `condition` and then sets up `condition.code`, `condition.display` and `condition.system`. It'd also need to maybe take `system` and take a callback to get the list of options.

WesleyKapow (changes requested):
Screenshots!

.changeset/dull-sheep-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Removes snomed field and condition name field in favor autocomplete conditions component.
```
pmadria3, line 5: How about "Replaces snomed and condition name field with an autocomplete conditions component."?

WesleyKapow: Great suggestion. And use `Replace` instead of `Replaces`...

You want to “Use the imperative mood in the subject line” (like “Add” and “Fix”)
https://zeushealth.atlassian.net/wiki/spaces/ENG/pages/116129793/Pull+Requests+Code+Reviews

src/api/conditions.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  const response = await fetch(
+    `${getFormsConditionsUrl(env)}?display=${searchTerm}`,
+    {
+      headers: { Authorization: `Bearer ${authToken}` },
+    }
+  );
+  const data = await response.json();
+  return data.conditionsList;
+};
```
pmadria3, line 21: Rename the file to autocomplete-conditions.ts? We have a bunch of conditions.ts files.

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+
+  const options = conditions.data?.map(
+    (condition: ConditionsAutoCompleteOption) => ({
+      value: condition.display,
+      id: condition.code,
+    })
+  );
+
+  // useEffect(() => {
```
pmadria3, line 44: do we need this commented code?

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  return (
+    <>
+      <ComboboxField
+        options={options || []}
+        query={query}
+        setQuery={setQuery}
+        handleSelectChange={handleSelectedConditonChange}
+        name={inputProps.name as string}
```
pmadria3, line 71: instead of force typing - would you just want to make it optional param? That way we aren't forcing undefined to be string

src/components/content/forms/form-field.tsx
```diff
@@ -19,16 +23,22 @@ export const FormField = ({
...
   hidden,
+  render,
   ...inputProps
 }: FormFieldProps) => {
   // We display dates in MM/DD/YYYY format, but date input fields
   // expect it to be in YYYY-MM-DD format.
+
   const value =
     inputProps.type === "date"
       ? formatDateLocalToISO(defaultValue as string)
```
pmadria3, line 34: don't need the as string

src/components/content/forms/form-field.tsx
```diff
@@ -19,16 +23,22 @@ export const FormField = ({
...
       : defaultValue;
 
   const getFieldComponent = () => {
+    if (render) {
```
pmadria3, line 38: NIT: general comment - we have a bunch of `if () {return}` we can probably next them in `if...else if...else...`

luthfib: We can't do that here because we have a default return at the end if none of the if statements are met. If we add else if then that would never be met. We could add if, else if, else if, else, but personally I find this easier to read.

pmadria3: that's fair - just left it as a nit : )

pmadria3 (dismissed):
LGTM

src/api/autocomplete-conditions.ts
```diff
@@ -0,0 +1,40 @@
...
+    `${getFormsConditionsUrl(env)}?display=${searchTerm}`,
+    {
+      headers: { Authorization: `Bearer ${authToken}` },
+    }
+  );
+  const data = await response.json();
+  return data.conditionsList;
+};
+
+export const setAutoCompleteConditions = async (
```
WesleyKapow, line 7: Better to keep functions a bit more explicit and "pure" where they don't modify things themselves. So instead of having to pass `setConditions` to this, what if instead this returned the list of condition options, something like:

`const conditions = await fetchConditionOptions(...)`

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,106 @@
+import { Combobox } from "@headlessui/react";
+import { debounce } from "lodash";
+import React, { useEffect, useMemo } from "react";
```
WesleyKapow, line 3: Replace uses of `React.foo` with `foo` directly. That way we are explicit about what we are importing and using from react.

src/api/autocomplete-conditions.ts
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { Env } from "..";
+import { getFormsConditionsUrl } from "./urls";
+
+export type QueryKeyAutoCompleteConditions = [string, string, Env, string];
```
WesleyKapow, line 5: This used?

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export type ComboboxFieldProps = {
+  options: ComboxboxFieldOption[];
+  query: string;
+  setQuery: React.Dispatch<React.SetStateAction<string>>;
+  handleSelectChange: (eventValue: string) => void;
+  readonly: boolean | undefined;
+};
+
+export const ComboboxField = ({
```
WesleyKapow, line 15: What if the interface to this was instead an `onSearchChange` that debounced calls. Conditions autocomplete can then setup a `handleOnSearch` to fetch the options and update the `options` passed to the combobox. So no need for `query` or `setQuery`.

`handleSelectChange` is also hopefully not needed as the combobox should setup the `input` needed. It should behave similar to our other field components by taking in a `name` and possibly other things like `disabled`.

Also, react naming convention is to use `onFoo` and `handleOnFoo` for events. So the property would be `onSelectChange` or simply `onChange` (since that is the standard for field components) and then you'd do `onChange={handleOnChange}` in the consumer.

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    return debounce(handleSearchInputChange, 300);
+  }, [setQuery]);
+
+  useEffect(
+    () => () => {
+      debouncedSearchInputChange.cancel();
+    },
+    [query, debouncedSearchInputChange]
+  );
```
WesleyKapow, line 36: This needed? Why do we need to cancel?

src/components/content/forms/combobox-field.tsx
```diff
@@ -0,0 +1,106 @@
...
+    </Combobox>
+  );
+};
+
+type RenderCorrectOptionsProps = {
+  options: ComboxboxFieldOption[];
+  query: string;
+};
+
+const RenderCorrectOptions = ({
```
WesleyKapow, line 57: "Render correct options" is a weird name for a component. How about just `ComboboxOptions`? I hope it renders, and I hope it renders correctly :P.

WesleyKapow (changes requested):

src/components/content/forms/conditions.ts
```diff
@@ -122,7 +101,8 @@ export const createOrEditCondition = async (
     result.success = false;
   }
 
-  queryClient.invalidateQueries(["conditions"]);
+  queryClient.invalidateQueries([PATIENT_CONDITIONS_KEY]);
+  queryClient.invalidateQueries([OTHER_PROVIDER_CONDITIONS_KEY]);
```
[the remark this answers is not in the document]
luthfib: We could have an explicit check whether it was updated or changed, but yes since this function createorEdits resources, it could theoretically cause that list to change

src/utils/form-helper.ts
```diff
@@ -137,7 +148,6 @@ export function getParamsInternal<T>(
       if (!Array.isArray(errors[key])) {
         errors[key] = [errors[key]];
       }
-      errors[key].push(message);
```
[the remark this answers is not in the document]
luthfib: This was appending messages when using ZodArray type. I couldnt think of a good use case when we'd want to display multiple error messages for one field, I think we'd always want to display one error message for each field. To support nested Objects types this was also getting appended so you would get something like `required required required` for the error messages.

src/components/content/forms/form-field.tsx
```diff
@@ -69,6 +79,18 @@ export const FormField = ({
...
+  if (render && hidden) {
+    return (
+      <>
+        {render(readonly, {
+          ...inputProps,
+          defaultValue,
+          hidden,
+        })}
+      </>
+    );
```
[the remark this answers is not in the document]
luthfib: I'll have to play around with the types. The issue didn't know if render always returned a React Element

src/api/autocomplete-conditions.ts
luthfib: The issue here is that `useEffects` are synchronous. So I couldn't await this function

src/components/content/forms/combobox-field.tsx
luthfib: This is when the component unmounts we want to make sure we dont call anything afterwards.

src/api/autocomplete-conditions.ts
luthfib: Maybe I could create a setter inside autocomplete and keep the fetching pure...?

src/components/content/forms/combobox-field.tsx
WesleyKapow: hmmm what's the risk here? And this is setup to cancel every time `query` changes?

src/components/content/forms/form-field.tsx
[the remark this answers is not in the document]
WesleyKapow: Ya see if you can make it so!

src/components/content/forms/conditions.ts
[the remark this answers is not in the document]
WesleyKapow: hmmm I'm not 100% but I suspect this isn't needed, but since I'm not sure I agree, let's leave it as is!

src/api/autocomplete-conditions.ts
WesleyKapow: You can create an async function within your `useEffect` function. This is a common pattern to work around this issue.

```ts
useEffect(() => {
  async function foo() {
    const conditions = await fetchConditionOptions();
  }
  
  // And then call it.
  foo();
}, [...])
```

luthfib: Yup, that's what I ended up doing.

src/components/content/forms/combobox-field.tsx
luthfib: I'll remove it

pmadria3:
So when I search something different and then search for pain, I get something different and /or previous results. Might need a delay or a reset of the state

Screen Shot 2022-10-21 at 3 32 23 PM

src/components/content/forms/form-field.tsx
```diff
@@ -81,9 +99,9 @@ export const FormField = ({
           <LockClosedIcon className="ctw-absolute ctw-right-3 ctw-top-1/2 ctw-h-4 ctw-w-4 ctw--translate-y-1/2 ctw-transform ctw-fill-content-lighter" />
         )}
         {error && (
-          <div className="ctw-pointer-events-none ctw-absolute ctw-inset-y-0 ctw-right-0 ctw-flex ctw-items-center ctw-pr-3">
+          <div className="ctw-pointer-events-none ctw-absolute ctw-inset-y-0 ctw-right-0 ctw-top-2 ctw-flex ctw-h-min ctw-items-center ctw-pr-3">
             <ExclamationCircleIcon
-              className="ctw-h-5 ctw-w-5 ctw-text-error-main"
+              className="ctw-h-5 ctw-w-5 ctw-text-error-main  "
```
pmadria3, line 104: extra space

pmadria3 (dismissed):
LGTM

src/components/core/form/combobox-field.tsx
```diff
@@ -26,16 +26,19 @@ export const ComboboxField = <T,>({
...
 
     return debounce(handleSearchInputChange, 300);
-  }, [onSearchChange, searchTerm]);
+  }, []);
+
+  useEffect(() => {
+    if (searchTerm.length > 1) {
+      onSearchChange(searchTerm);
+    }
+  }, [searchTerm, onSearchChange]);
```
WesleyKapow, line 41: Can this be rolled into `handleSearchInputChange`?

src/components/core/form/combobox-field.tsx
```diff
@@ -0,0 +1,122 @@
...
+        onChange={(e) => {
+          debouncedSearchInputChange(e);
+        }}
+        placeholder="Type to search"
+      />
+
+      <input
+        hidden
+        name={name}
+        value={JSON.stringify(defaultValue || inputValue)}
```
WesleyKapow, line 62: Do you have to stringify it :(? Little awkward when the value is not an object.

And the saving works ok when this is stringified? We don't have to then parse it?

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -20,52 +23,36 @@ export const ConditionsAutoComplete = ({
...
-  const [query, setQuery] = useState((inputProps.defaultValue as string) || "");
-  const [selectedCoding, setSelectedCoding] = useState<fhir4.Coding>();
   const [conditions, setConditions] = useState<fhir4.Coding[]>();
   const [options, setOptions] = useState<ComboxboxFieldOption[]>();
 
-  useEffect(() => {
-    if (query.length > 1) {
-      setAutoCompleteConditions(authToken, env, query, setConditions);
-    }
+  const onSearchChange = useCallback(
```
WesleyKapow, line 29: nit: React naming convention is to call this `handleSearchChange`

src/components/content/forms/conditions-autocomplete.tsx
```diff
@@ -20,52 +23,36 @@ export const ConditionsAutoComplete = ({
...
-        (item: fhir4.Coding) => item.display === eventValue
-      )[0];
-      setSelectedCoding(currentCondition);
-    }
-
-    if (conditions?.length) {
-      setQuery(eventValue);
-    }
-  };
+  }, [authToken, env, conditions]);
```
WesleyKapow, line 45: Why the use of `useEffect` and not roll `setOptions` into `onSearchChange`.. probably don't even need the `conditions` state. Just options that get updated when search changes.

WesleyKapow (changes requested):
Love the changes, looking really good!

Request: Please don't move files after a PR has begun, or if you do make sure to do it in a single commit and make it clear that all you are doing is moving the file (that's fine too!). Otherwise it makes it hard to see "changes since last review" as the file is 100% new and just all green in the diff.

src/components/core/form/combobox-field.tsx
luthfib: We do have to stringify it because input values cannot be objects. Our helpers take care of parsing it back to objects.

luthfib: Maybe we should check if it is an object before stringifying it?

WesleyKapow: Yes, unless you are always parsing it on the other side.

Stringifying a string makes it a string with quotes!
```
JSON.stringify("rar")
'"rar"'
```

WesleyKapow (approved):
LGTM, one comment around use of `stringify`.

WesleyKapow (approved):

PR #121, opened 2022-10-18
CTW-369 Delete Conditions from Patient Record

melocule:
This PR allows users to delete a condition resource on a patient's profile via a new meatball menu option next to each record on the table that pulls up a confirmation dialogue. 
In our codebase this adds a modal core component, a modal component to confirm deleting a given resource, and a  "dangerous" styling option to meatball menu options to make the color of the text stand out.

**Tests**
1. Here's a [video](https://user-images.githubusercontent.com/33408946/196495284-8c3ee40e-2189-4c8e-80a4-9cebb43644b7.mov) of what deleting looks like from start to end. This is the [test condition](https://fhirplace.dev.zusapi.com/#/Condition/f0128094-e097-4a62-a7eb-36d77d64e112) I used.
2. An error is shown if the delete fails:
Remove Condition


luthfib:
Did you confirm design that for the cancel we want an actual button vs text as shown in the design?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "testing.automaticallyOpenPeekView": "never"
+}
```
luthfib, line 3: What is this for?

scripts/prefixer/prefixer.ts
```diff
@@ -1,138 +1,27 @@
...
+      // Replaces non-prefixed appearances of this class name that aren't part of a larger word. Be careful, this can replace things you didn't want to.
+      new RegExp(
+        `(?<!import .*)` + // Not an import
+          `(?<!\\w|\\w-|\\.)` + // Not part of another name, incl not already prefixed (part 1), not a property/function
+          `(?<!//.*)` + // Not part of a comment
+          `(?<=".*)` + // Preceded somewhere on that line by a quotation mark
+          `${selector}`, // Thing to prefix
+        "g"
+      ),
+      replacement
```
luthfib, line 21: Are these changes intended? If changing things otherwise not outlined in the ticket it is best to either... do a separate pr or call them out in pr description

src/components/content/conditions.tsx
```diff
@@ -27,19 +27,21 @@ import {
...
+import { ModalConfirmDelete } from "./modal-confirm-delete";
 
 export type ConditionsProps = {
   className?: string;
 };
 
 const EMPTY_MESSAGE = "No conditions found";
 const ERROR_MSG =
-  "There was an error fetching conditions for this patient. Refresh the page or contact your organization's technical support if this issue persists.";
+  "There was an ctw-error fetching conditions for this patient. Refresh the page or contact your organization's technical support if this issue persists.";
```
luthfib, line 38: Why ctw error? Not sure we want this. The error is not necessarily a ctw error.. could be an error with how our customers setup the integration.

src/components/content/conditions.tsx
```diff
@@ -53,6 +55,8 @@ export function Conditions({ className }: ConditionsProps) {
     useState<FormEntry[]>();
   const [conditionForHistory, setConditionForHistory] =
     useState<ConditionModel>();
+  const [conditionForDelete, setConditionForDelete] =
+    useState<ConditionModel>();
```
luthfib, line 59: I think we are creating extra variables here without needing to. I think we could create a new variable, maybe `currentlySelectedCondition` that can be used by any drawer.. history, delete modal, etc

luthfib (changes requested):
Looking good, left some comments

.changeset/itchy-pens-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
```
luthfib, line 4: Not sure we need `Clients don't need to change their code`. If we were doing a breaking change we would do it via communicate using versioning.

luthfib:
The confirmation text doesn't seem to match the text in the Figma

src/models/resource.ts
```diff
@@ -0,0 +1,5 @@
+export interface ResourceModel {
+  id: string | undefined;
+  resourceType: string;
+  resource: fhir4.Condition;
+}
```
WesleyKapow, line 5: Why the need for `ResourceModel`? And this is not generic as you have `fhir4.Condition` for the resource.

src/fhir/action-helper.ts
```diff
@@ -1,19 +1,14 @@
...
-}
-
-type CreateOrEditData<T extends FhirResourceBase> = {
+type CreateOrEditData<T extends ResourceModel> = {
   resourceModel: T;
   fhirClient: Client;
 };
 
-export async function createOrEditFhirResource<T extends FhirResourceBase>({
+export async function createOrEditFhirResource<T extends ResourceModel>({
```
WesleyKapow, line 11: Should we instead be editing/saving FHIR resources and not our models?

scripts/prefixer/renamings.json
```diff
@@ -0,0 +1,1535 @@
+{
```
WesleyKapow, line 1: We don't need a map of renamings, we only need the list of class names to rename. They will always get `ctw-` prefixed.

.vscode/settings.json
WesleyKapow, line 3: gitignore the `.vscode` directory

.changeset/itchy-pens-repeat.md
```diff
@@ -0,0 +1,5 @@
...
+To allow users to remove conditions from a patient's profile, a "Delete" meatball menu option was added to each record of the condition table that invokes a confirmation modal to delete the FHIR resource. Clients don't need to change their code.
```
WesleyKapow, line 5: See if you can make this less wordy. Avoid "meatball menu" as that is not a very user friendly description.

> Add "Delete" menu option ...

src/components/content/conditions.tsx
```diff
@@ -27,19 +27,21 @@ import {
...
 const EMPTY_MESSAGE = "No conditions found";
 const ERROR_MSG =
-  "There was an error fetching conditions for this patient. Refresh the page or contact your organization's technical support if this issue persists.";
+  "There was an ctw-error fetching conditions for this patient. Refresh the page or contact your organization's technical support if this issue persists.";
 
 export function Conditions({ className }: ConditionsProps) {
   const containerRef = useRef<HTMLDivElement>(null);
   const breakpoints = useBreakpoints(containerRef);
   const [drawerIsOpen, setDrawerIsOpen] = useState(false);
+  const [showModal, setShowModal] = useState(false);
```
WesleyKapow, line 44: What if we have more than one modal. Make this more specific.

src/components/content/conditions.tsx
```diff
@@ -245,6 +261,13 @@ export function Conditions({ className }: ConditionsProps) {
                   setConditionForHistory(condition);
                 },
               },
+              {
+                name: "Delete",
+                className: "dangerous",
+                action: () => {
+                  handleConditionDelete(condition);
+                },
+              },
```
WesleyKapow, line 270: I don't think we can delete summary/lens resources.

src/components/core/dropdown-menu.scss
```diff
@@ -5,6 +5,9 @@
 .ctw-dropdown-menu-item {
   @apply ctw-flex ctw-w-full ctw-cursor-pointer ctw-p-2 ctw-text-primary-dark ctw-outline-none;
 
+  &.dangerous {
+    @apply ctw-text-error-main;
+  }
```
WesleyKapow, line 10: nit: newline below this

WesleyKapow (dismissed):

.changeset/itchy-pens-repeat.md
pmadria3: How about "Allow users to remove conditions from a patient's profile via the delete option in meatball menu. This will invoke a confirmation dialog to confirm the deletion of the FHIR resource."?

.vscode/settings.json
melocule: Huhh I'm actually not sure why this was added, I'll change it back

src/models/resource.ts
melocule: This was actually the FhirResourceBase interface that was defined in another file, I tried moving it to /models because I thought it made sense for it to be there

src/fhir/action-helper.ts
melocule: I think the reason why we had it edit the model and not the resource itself is so we have the "display" property that maps to different properties for different FHIR resource types

WesleyKapow: How is `display` used here?

I think we should change this to save the fhir resource. It's a little misleading that it takes a model and actually saves the resource.

src/components/content/conditions.tsx
```diff
@@ -179,32 +177,32 @@ export function Conditions({ className }: ConditionsProps) {
   return (
     <div
       ref={containerRef}
-      className={cx("ctw-conditions", className, {
-        "ctw-conditions-stacked": breakpoints.sm,
+      className={cx("conditionFoconditions", className, {
```
luthfib: Is this class name supposed to be `conditionFoconditions` I think maybe you changed this accidently?

luthfib (changes requested):
I think you might have used the prefixer incorrectly in a few classes. Take a look at the classnames you changed

src/models/resource.ts
melocule: Removed!

melocule:
New styling to fit the new figma
Screen Shot 2022-10-24 at 1 11 21 PM
Screen Shot 2022-10-24 at 1 03 30 PM

src/components/content/conditions-history.tsx
```diff
@@ -77,7 +78,7 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
           historyResponse.data.bundle
         );
         const conditionModels = historyResponse.data.conditions.map(
-          (c) => new ConditionModel(c, includedResources)
+          (c: Condition) => new ConditionModel(c, includedResources)
```
pmadria3, line 81: you can get rid of Condition

src/fhir/conditions.ts
```diff
@@ -175,7 +175,7 @@ export const filterConditionsWithConfirmedCodes = (
   confirmedCodes: fhir4.Coding[]
 ) =>
   target.filter((c) => {
-    const conditionModel = new ConditionModel(c);
+    const conditionModel: ConditionModel = new ConditionModel(c);
```
pmadria3, line 178: can change to `const ConditionModel`

pmadria3 (commented):
Looking good, left a couple comments. Also see some unresolved conversations

pmadria3 (approved):

PR #133, opened 2022-10-19
Removal of unnecessary function

pmadria3:
Quick PR to remove the unnecessary function
Screen Shot 2022-10-19 at 12 33 16 PM


WesleyKapow (approved):
Awesome, and you verified it works correctly still!?

pmadria3:
yep I also just added screenshot just for recordkeeping

PR #135, opened 2022-10-19
Ctw 431/add tag patient filtering

WesleyKapow:
* Minor code cleanup to `useBreakpoints` (remove incorrect "class").
* Fix condition history bug (see commit message)
* Add optional `tags` prop to `PatientProvider` .. this required a little refactor in how we handle searchParams and some plumbing of `tags` down to the search queries.


luthfib (approved):

pmadria3 (approved):
LGTM

PR #137, opened 2022-10-19
Add aria-required attribute to the form fields

pmadria3:
This PR removes the fill out the field tooltip and adds aria-required attribute to the form fields that are required.
Screen Shot 2022-10-19 at 2 06 19 PM

Also: fixed hidemenu to reflect readonly for the meatball menu when it is present in App.tsx

### DIFF
--- a/.changeset/four-shirts-hide.md
+++ b/.changeset/four-shirts-hide.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Remove required attributes from input fields and add an aria-required attribute to the form fields when adding/editing conditions.
+Remove required attributes from input fields and add aria-required attribute to those input fields when adding/editing conditions.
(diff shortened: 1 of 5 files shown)


.changeset/four-shirts-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
```
luthfib, line 4: Maybe call out the fact that we removed required attributes?

luthfib (approved):
LGTM, left one comment

.changeset/four-shirts-hide.md
pmadria3: grr changeset

luthfib (dismissed):

src/utils/form-helper.ts
```diff
@@ -241,7 +242,10 @@ export function getInputProps(
     options,
   };
 
-  if (!(def instanceof ZodOptional)) inputProps.required = true;
+  if (!(def instanceof ZodOptional)) {
+    inputProps.required = false;
```
WesleyKapow, line 246: Is this line needed?

PR #142, opened 2022-10-20
CTW-432/zus account builder

WesleyKapow:
* Add optional `builderId` prop to `CTWProvider`. When set, we'll use that builderId for any filtering AND we'll set the `Zus-Account` header for FHIR requests.
* Refactor `CTWProvider` and search code to pass around a new `CTWRequestContext`. This gives us one place to add things that any of our requests need. For now that consists of `env, builderId, authToken, fhirClient`.
* Add new `useQueryWithPatient` helper which will handle setting up the `patient` model and the `requestContext`. Refactor existing uses of `useQuery` to use that instead.
* Add a default `staleTime` of `30 seconds` to all `useQuery` queries. This means that we will use the cached result for queries based on their `keys`.
* Add a `staleTime` of `5 minutes` for patient queries.


src/components/core/ctw-provider.tsx
```diff
@@ -89,12 +96,19 @@ function useCTW() {
...
 
-  const getCTWFhirClient = useCallback(async () => {
+  const getRequestContext = useCallback(async () => {
     const authToken = await context.actions.handleAuth();
-    return getFhirClient(context.env, authToken);
+    const requestContext: CTWRequestContext = {
+      env: context.env,
+      authToken,
+      builderId: context.builderId ?? claimsBuilderId(authToken),
+      fhirClient: getFhirClient(context.env, authToken, context.builderId),
```
luthfib, line 105: This is nice. I already had to use env and authToken in my pr so good to have an abstraction for this

luthfib (approved):
LGTM, lots of quality of life changes

.changeset/seven-wasps-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Add optional builderId prop to CTWProvider. When set, we'll use that builderId for any filtering AND we'll set the Zus-Account header for FHIR requests.
```
pmadria3, line 5: SuperNit: slight edit- When set, we can use that builderID for any filtering AND set the Zus-Account header for FHIR requests.

pmadria3 (approved):
LGTM, minus a super nit which you can fix or ignore

PR #144, opened 2022-10-21
Hex to RGBA converter 

pmadria3:
CTW-379

This PR adds a hex to rgba converter that allows to convert the values and also declare opacity. The black is now replaced with CTW theme.

Before:
image

After: 
image

Note: will need to wait till autocomplete conditions is merged to merge this PR.


src/styles/tailwind.theme.ts
```diff
@@ -83,12 +83,26 @@ export function mapToCSSVar(colorConfig: ColorTheme): Record<string, string> {
...
+// Converts hex values to RGB values.
+function hexToRGB(hex: string, opacity?: number): string {
+  const r = parseInt(hex.slice(1, 3), 16);
+  const g = parseInt(hex.slice(3, 5), 16);
+  const b = parseInt(hex.slice(5, 7), 16);
+  if (opacity) {
+    return `rgba(${r}, ${g}, ${b}, ${opacity})`;
+  }
+  return `rgb(${r}, ${g}, ${b})`;
+}
```
luthfib, line 108: I think we can just return `rgb(${r}, ${g}, ${b})`. Since this is happening in the converter and users are going to specify hex I dont see us needing to allow option to specify rgba. Also would be kind nice if we supported rgba values, so if a user specifies rgba we just return rgba instead of converting.

luthfib (commented):
Looking good, just a couple of changes

src/styles/tailwind.theme.ts
```diff
@@ -83,12 +83,23 @@ export function mapToCSSVar(colorConfig: ColorTheme): Record<string, string> {
...
+
+// Converts hex values to RGB values if it's not already in RGB format.
+function hexToRGB(colorMapping: string): string {
+  if (colorMapping.includes("#")) {
+    const r = parseInt(colorMapping.slice(1, 3), 16);
+    const g = parseInt(colorMapping.slice(3, 5), 16);
+    const b = parseInt(colorMapping.slice(5, 7), 16);
+    return `rgb(${r}, ${g}, ${b})`;
+  }
+  return colorMapping;
```
WiktorKania, line 104:
```suggestion
  if (colorMapping[0] !== "#") return colorMapping;
  const r = parseInt(colorMapping.slice(1, 3), 16);
  const g = parseInt(colorMapping.slice(3, 5), 16);
  const b = parseInt(colorMapping.slice(5, 7), 16);
  return `rgb(${r}, ${g}, ${b})`;
```

WiktorKania (approved):
Looks good, left one suggestion.
Keep in mind that this converter won't work for shorthand values like `#FFF`.
Also, it doesn't allow you to specify the alpha channel so technically you can't declare opacity. If we want to declare opacity then hex values will do just fine too, you can add another octet at the end which will specify the opcity.
Approving this since I don't really know the requirements

src/styles/tailwind.theme.ts
WesleyKapow: and yet another option: `colorMapping.startsWith('#')`

src/components/content/forms/drawer-form-with-fields.tsx
```diff
@@ -63,11 +62,7 @@ export const DrawerFormWithFields = <T,>({
                 className="ctw-space-y-1.5 ctw-text-sm ctw-font-medium ctw-text-content-black"
               >
                 <div className="ctw-flex ctw-justify-between">
-                  <label
-                    className={cx({ "ctw-error": error }, "leading-tight")}
-                  >
-                    {label}
-                  </label>
+                  <label>{label}</label>
```
WesleyKapow, line 65: Why the change to styles here? I think we want to keep the classes as they were.

WesleyKapow (changes requested):

src/styles/tailwind.theme.ts
pmadria3: Yeah I agree - originally I had opacity but removed it but added it back and made it optional.

src/components/content/forms/drawer-form-with-fields.tsx
pmadria3: Design team updated some of the error handling and thus the change was small I just implemented it in this PR. The third screenshot captures the change.

WesleyKapow: Is there a ticket for that? Can you link me to slack discussion or figma for this?

pmadria3: Removed this per discussion

WesleyKapow (approved):

PR #148, opened 2022-10-24
Update Form page with minor fixes 

pmadria3:
CTW-456

This PR helps to resolve a few issues: 

Text style is consistent with the Figma  
Update label to not be highlighted 
Remove tooltip from onset field 

Screen Shot 2022-10-24 at 1 26 02 PM

Set abatement/onset to constrain dates to past dates or present

Screen Shot 2022-10-24 at 1 26 50 PM


WesleyKapow (approved):
LGTM

PR #150, opened 2022-10-24
Ctw 357/choose best code

luthfib:
- Adds preferred coding function to get the coding best on our definition of best coding, favoring enrichment. 
- Adds default value for add new provider and edit condition. 
- Fixes issue where when adding condition the error message wasnt 100% clear. 
- Remove uneeded code


luthfib:
My bad on the package.lock.json

src/fhir/conditions.ts
```diff
@@ -24,6 +24,14 @@ import {
   SYSTEM_SNOMED,
 } from "./system-urls";
 
+export const BEST_CODE_PREFERENCE_ORDER = [
```
WesleyKapow, line 27: This is specific to conditions.

How about `CONDITION_CODE_PREFERENCE_ORDER`?

src/fhir/codeable-concept.ts
```diff
@@ -21,3 +23,35 @@ export function findCoding(
...
+  system: string,
+  concept?: fhir4.CodeableConcept
+): fhir4.Coding | undefined {
+  return find(concept?.coding, {
+    system,
+    extension: [{ url: SYSTEM_ENRICHMENT }],
+  });
+}
+
+export const findCodingByOrderOfPreference = (
```
WesleyKapow, line 37: Maybe this should take the preferences so it isn't specific to conditions.

src/fhir/codeable-concept.ts
```diff
@@ -21,3 +23,35 @@ export function findCoding(
...
+  concept: fhir4.CodeableConcept | undefined
+): fhir4.Coding | undefined => {
+  let codeSystem;
```
WesleyKapow, line 40: 1. Don't need this, instead of `break` you can `return`.
2. `codeSystem` should be `coding`.

src/fhir/codeable-concept.ts
```diff
@@ -21,3 +23,35 @@ export function findCoding(
...
+
+  for (let i = 0; i < BEST_CODE_PREFERENCE_ORDER.length; i += 1) {
+    const code = BEST_CODE_PREFERENCE_ORDER[i];
+    if (code.checkForEnrichment) {
+      codeSystem = findCodingWithEnrichment(code.system, concept);
+      if (codeSystem) {
+        break;
+      }
+    }
```
WesleyKapow, line 49: You need an else here, otherwise it will check for snomed with enrichment and then snomed without where ticket has all enrichment systems first.

You'll also need to add snomed and icd10 without enrichment in the right order to `BEST_CODE_PREFERENCE_ORDER`

src/components/content/forms/condition-schema.tsx
```diff
@@ -31,26 +36,14 @@ export const getEditingPatientConditionData = ({
...
-  },
   {
     label: "Condition",
     field: "condition",
     value: condition.display,
     readonly: true,
     render: (readonly: boolean | undefined, inputProps) => (
       <ConditionsAutoComplete
-        defaultCoding={condition.snomedCoding}
+        defaultCoding={condition.codings}
```
WesleyKapow, line 46: `codings`? Is this more than one? Seems weird for the autcomplete component. We want edit to display the preferred but not modify ANY of the `codings` right?

src/models/conditions.ts
```diff
@@ -71,8 +75,19 @@ export class ConditionModel {
     return this.resource.code;
   }
 
+  get preferredCoding(): fhir4.Coding | undefined {
+    return findCodingByOrderOfPreference(this.resource.code);
+  }
+
+  get preferredSystem(): string | undefined {
```
WesleyKapow, line 82: This used?

WesleyKapow (changes requested):

src/fhir/codeable-concept.ts
luthfib: Yup, good catch on the else. I dont think we need to add snomed and icd10 without the enrichment to the array. Because if we dont find it in the enrichment find then we can just do a find in that same loop

WesleyKapow: Check again, I think you will need snomed without enrichment.

It should go:
1. Snomed with enrichment
2. ICD10 with enerichment
3. Snomed without enrichment
4. ICD10 without enrichment
5. ICD10_CM without
6. ...

So whatever gives us that!

src/components/content/forms/condition-schema.tsx
luthfib: We do this because we want to keep all of the codings when editing

src/fhir/conditions.ts
```diff
@@ -8,6 +8,8 @@ import {
 } from "@/utils/query-keys";
 import { SearchParams } from "fhir-kit-client";
 import { sortBy } from "lodash";
+import { CodePreference } from "./codeable-concept";
+
```
pmadria3, line 12: remove extra space

pmadria3 (approved):
LGTM

WesleyKapow (approved):

PR #151, opened 2022-10-24
Map level 2 codes to level 1

pmadria3:
[CTW-383](https://zeushealth.atlassian.net/browse/CTW-383)

This PR maps level 2 codes to level 1. Adding a new condition only gives you the requested dropdown changes:

Clinical status - active/inactive 

Verification Status - Confirmed, Unconfirmed, Refuted, Entered in error

After hitting save and then editing the condition you are able to only see the updated mappings, the following value is also written to the new changes.

Before Mapping:
Screen Shot 2022-10-24 at 3 06 52 PM

After Mapping:
Screen Shot 2022-10-24 at 3 06 26 PM

### DIFF
--- a/src/components/content/forms/condition-schema.tsx
+++ b/src/components/content/forms/condition-schema.tsx
@@ -70,12 +70,12 @@ const sharedFields = (condition: ConditionModel) => [
   },
   {
     label: "Clinical Status",
-    value: condition.clinicalStatus,
+    value: levelTwotoOneMapping(condition.clinicalStatus),
     field: "clinicalStatus",
   },
   {
     label: "Verification Status",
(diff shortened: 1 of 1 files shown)


src/components/content/forms/condition-schema.tsx
```diff
@@ -145,3 +136,20 @@ export const conditionAddSchema = z.object({
     }),
   }),
 });
+
+function levelTwoToOneMapping(value: string): string {
+  console.log("value", value);
```
luthfib, line 141: remove console.log

WesleyKapow (commented):
LGTM, outside of console log.

WesleyKapow (approved):

PR #152, opened 2022-10-25
Ctw 357/history system dedupe

luthfib:
Dedupes history records based on system.


src/models/conditions.ts
```diff
@@ -171,6 +183,13 @@ export class ConditionModel {
     return undefined;
   }
 
+  get preferredCoding(): fhir4.Coding | undefined {
+    return findCodingByOrderOfPreference(
+      CONDITION_CODE_PREFERENCE_ORDER,
+      this.resource.code
+    );
+  }
+
```
luthfib, line 192: I simply moved the location of this to be alphabetical

WesleyKapow (approved):
LGTM

Question though, could we use this new `knownCodingsWithEnrichmentPReference` everywhere we currently use `knownCodings`? And if so, can we instead replace `knownCodings` with this new one and keep the simpler name?

WesleyKapow (approved):

PR #153, opened 2022-10-25
Ctw 412/word wrap tables

luthfib:
- Add word wrap to all cells in tables to avoid content from spilling over to other cells.

### DIFF
--- a/src/components/core/table/table-data-cell.tsx
+++ b/src/components/core/table/table-data-cell.tsx
@@ -23,7 +23,8 @@ export const TableDataCell = <T extends MinRecordItem>({
       key={column.title ?? index}
       className={cx(
         { "ctw-font-medium ctw-text-content-black": index === 0 },
-        column.className
+        column.className,
+        "break-all"
       )}
     >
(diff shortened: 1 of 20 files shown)


src/components/core/table/table-data-cell.tsx
```diff
@@ -23,7 +23,8 @@ export const TableDataCell = <T extends MinRecordItem>({
       key={column.title ?? index}
       className={cx(
         { "ctw-font-medium ctw-text-content-black": index === 0 },
-        column.className
+        column.className,
+        "break-all"
```
WesleyKapow, line 27: Should it be `ctw-break-all`?

WesleyKapow (changes requested):
Please add before/after screenshots.

pmadria3 (dismissed):

WesleyKapow (approved):
LGTM, assuming autoprefixing works.

Also, you may want an "after" screenshot for both safari and chrome as they behave differently (chrome is nicer).

PR #154, opened 2022-10-25
Cleanup console errors

luthfib:
- Cleanup duplicate key error we were getting in the console in edit and add conditions.
- Remove error where react query key was returning undefined by adding an enabled field that can be passed to useQueryWithPatient. Per react query docs: 

image


src/fhir/conditions.ts
```diff
@@ -125,13 +125,11 @@ export function useOtherProviderConditions() {
   );
 }
 
-export function useConditionHistory(condition?: ConditionModel) {
+export function useConditionHistory(condition: ConditionModel) {
```
WesleyKapow, line 128: I think this still needs to optionally take `undefined`.

WesleyKapow (changes requested):

WesleyKapow (approved):

PR #162, opened 2022-10-26
Ctw-361/edit field requirements

pmadria3:
[CTW-361](https://zeushealth.atlassian.net/jira/software/projects/CTW/boards/49?selectedIssue=CTW-361)

This PR updates the field requirements and read only content for the Edit/Add Condition forms.

Edit Conditions: 
Screen Shot 2022-10-26 at 1 56 09 PM

Add Conditions: 
Screen Shot 2022-10-26 at 1 58 22 PM

Additionally: 
- Removed unused conditions-drawer-form file


src/components/content/forms/drawer-form-with-fields.tsx
```diff
@@ -19,13 +19,19 @@ export type FormEntry = {
 
 export type DrawerFormWithFieldsProps<T> = {
   title: string;
+  header?: ReactNode;
+  formType?: FormActionTypes;
```
WesleyKapow, line 23: Why do I need formType here?

src/components/content/forms/drawer-form-with-fields.tsx
```diff
@@ -43,56 +49,64 @@ export const DrawerFormWithFields = <T,>({
       {...drawerFormProps}
     >
       {(submitting, errors) => (
-        <div className="ctw-space-y-6">
-          {data.map(
-            ({ label, field, value, lines, readonly, hidden, render }) => {
-              const error = errors?.[field];
+        <div>
+          {formType === "Edit" && <div>{header}</div>}
```
WesleyKapow, line 53: See if just doing this will work for when it is defined and when it is not defined: `{header}`

src/components/content/conditions.tsx
```diff
@@ -299,6 +302,20 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
         <DrawerFormWithFields
           patientID={patientResponse.data.id}
           title={`${formAction} Condition`}
+          formType={formAction}
+          header={
+            <div className="ctw-py-2">
+              <div>
```
WesleyKapow, line 308: Combine these two divs

src/components/content/conditions.tsx
```diff
@@ -84,6 +86,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       setFormAction("Edit");
       setSchema(conditionEditSchema);
       setCurrentlySelectedData(getEditingPatientConditionData({ condition }));
+      setConditionHeader(condition);
     }
```
WesleyKapow, line 90: Do we need both of these states? Can we instead have a `selectedCondition` state?

src/components/content/conditions.tsx
```diff
@@ -299,6 +302,20 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
         <DrawerFormWithFields
           patientID={patientResponse.data.id}
           title={`${formAction} Condition`}
+          formType={formAction}
+          header={
```
WesleyKapow, line 306: Is this the same content as used by the condition history? May want to DRY it up into a `condition-header` component.

src/components/content/forms/condition-schema.tsx
```diff
@@ -36,19 +36,6 @@ export const getEditingPatientConditionData = ({
     readonly: true,
     hidden: true,
   },
-  {
```
WesleyKapow, line 39: Will editing a condition lose the codings now that we aren't setting up this field?

src/components/content/forms/drawer-form-with-fields.tsx
```diff
@@ -43,56 +49,64 @@ export const DrawerFormWithFields = <T,>({
...
+          <div className="ctw-space-y-6 ctw-py-4">
```
WesleyKapow, line 54: Why the `py-4`? Is that to create space between this and the header? It will add space above and below!

Favor using `space-y` classes on a parent div.

WesleyKapow (changes requested):

src/components/content/forms/drawer-form-with-fields.tsx
pmadria3: so for this and the `formType` comment - I do need this because I need to know which `formType` it is (`add` vs `edit`) otherwise if I remove this and just do `{header}`, the title shows up in the add Form drawer as well.

pmadria3: Taking a closer look, it seems to be that the previous action doesn't seem to reset and thus need to reset the state/clear cache after drawer closes

WesleyKapow: Have the consumer of this only pass in `header` for edit.

src/components/content/conditions-history.tsx
```diff
@@ -133,13 +134,8 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
 
     return (
-      <div className="ctw-space-y-6">
-        <div>
-          <div className="ctw-text-2xl">
-            {condition.display} ({condition.preferredCoding?.code})
-          </div>
-          <div className="ctw-text-sm">{condition.ccsGrouping}</div>
-        </div>
+      <>
```
WesleyKapow, line 137: Why removing the div and the space?

WesleyKapow (changes requested):

src/components/content/conditions-history.tsx
```diff
@@ -135,20 +135,24 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
-        <CollapsibleDataListStack
-          entries={conditionsWithDate}
-          limit={CONDITION_HISTORY_LIMIT}
-        />
-        {conditionsWithoutDate.length !== 0 && (
-          <div className="ctw-space-y-2">
-            <div className="ctw-font-medium">Records with no date:</div>
+        <div className="ctw-space-y-6">
+          <ConditionHeader condition={condition} />
+          <div>
```
WesleyKapow, line 140: Is this div needed?

WesleyKapow (dismissed):
One minor nit but otherwise LGTM

src/components/content/forms/form-field.tsx
```diff
@@ -81,14 +81,6 @@ export const FormField = ({
     );
   };
 
-  if (render && hidden) {
```
WesleyKapow, line 84: Why removing this? I don't fully know this code but this seems unrelated to your work.

WesleyKapow (changes requested):

WesleyKapow (dismissed):
LGTM

WesleyKapow (approved):

PR #168, opened 2022-10-27
CTW-374 ODS and Multi-Field Form Error Display

melocule:
CTW-374
Shows in-line errors for form issues involving multiple fields, as well as errors associated with the request as opposed to a field (e.g. ODS error, network error).

**Example A.** Multi-field errors, in-line with support for multiple per field
image

**Example B.** Single request error (network went offline)
EDIT CONDITION

**Example C.** Multiple request errors (artificial example of an issue response from ODS)
EDIT CONDITION


src/components/content/conditions-history.tsx
```diff
@@ -78,7 +79,7 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
           historyResponse.data.bundle
         );
         const conditionModels = historyResponse.data.conditions.map(
-          (c) => new ConditionModel(c, includedResources)
+          (c: Condition) => new ConditionModel(c, includedResources)
```
WesleyKapow, line 82: Was this needed? `historyResponse.data.conditions` should already be `fhir4.Condition` type so this shouldn't be needed.

We don't need to explicitly add types when TS already knows them. E.g. `['foo', 'bar'].map((item: string) => {}` does not need the `: string` typing as that is already a given due to what you are mapping over.

src/fhir/conditions.ts
```diff
@@ -178,7 +178,7 @@ export const filterConditionsWithConfirmedCodes = (
   confirmedCodes: fhir4.Coding[]
 ) =>
   target.filter((c) => {
-    const conditionModel = new ConditionModel(c);
+    const conditionModel: ConditionModel = new ConditionModel(c);
```
WesleyKapow, line 181: This is superfluous, it'd be like doing `const foo: string = "hello world"`. Only when we want to be careful that the right side is in fact the type we expect do we need to do this, like:
```ts
const data: LoaderData = {...}
```
Since `{...}` could be any object, only by adding in the explicit `: LoaderData` do we enforce its type/shape.

src/fhir/action-helper.ts
```diff
@@ -16,16 +15,10 @@ export async function createOrEditFhirResource(
       });
     }
     return await fhirClient.create({
-      resourceType: resource.resourceType,
+      resourceType: "foo",
```
WesleyKapow, line 18: This don't look right O_o

src/components/core/main.scss
```diff
@@ -41,9 +41,9 @@
     padding: 0;
   }
 
-  /* Make sure the floating dropdown get full-width of reference. */
+  /* Align the error icon */
   .ctw-listbox-button + div {
-    width: 100%;
+    width: 8%;
   }
```
WesleyKapow, line 47: See if you can get the style right without this. My thinking is that `8%` seems awfully specific. The icon should be positioned in a more general manner (like "be centered and 4px from the right")

src/components/content/forms/form-field.tsx
```diff
@@ -93,6 +96,7 @@ export const FormField = ({
     return <input {...inputProps} defaultValue={value} hidden={hidden} />;
   }
 
+  // styling of the error icon before: ctw-pointer-events-none ctw-absolute ctw-inset-y-0 ctw-right-0 ctw-top-2 ctw-flex ctw-h-min ctw-items-center ctw-pr-3
```
WesleyKapow, line 99: remove

src/components/core/alert.tsx
```diff
@@ -11,20 +11,22 @@ export type AlertProps = {
 };
 
 export const ErrorAlert = ({ header, children, className }: AlertProps) => (
-  <div className={cx("ctw-alert-bg ctw-bg-error-bg", className)}>
-    <ExclamationIcon className="ctw-h-5 ctw-text-error-icon" />
+  <div className={cx("ctw-alert-bg ctw-rounded-md ctw-bg-error-bg", className)}>
+    <XCircleIcon className="ctw-h-5 ctw-flex-none ctw-text-error-icon" />
```
WesleyKapow, line 15: I think we want exclamation here too. The X circle looks like something you can close.

src/models/operation-outcome-issue.ts
```diff
@@ -0,0 +1,27 @@
+import { codeableConceptLabel } from "@/fhir/codeable-concept";
+
+export type OperationOutcomeIssueSeverity =
+  | "error"
+  | "fatal"
+  | "warning"
+  | "information";
```
WesleyKapow, line 7: `fhir4.OperationOutcomeIssue.severity` already has this. You can use  `fhir4.OperationOutcomeIssue['severity']`. We also don't currently force all functions to have explicit return types so TS will pick this up automatically in your `get severity()` function.

src/components/content/forms/condition-schema.tsx
```diff
@@ -118,6 +118,30 @@ export const conditionAddSchema = z.object({
...
+    condition.onset &&
+    condition.abatement < condition.onset
+  ) {
+    ctx.addIssue({
+      code: Zod.ZodIssueCode.custom,
+      message: "Abatement cannot happen before onset.",
+      path: ["abatement"],
+    });
+  }
+};
```
WesleyKapow, line 143: I was hoping this could be on the schema itself using `refine` or `superRefine`. Did you try that?

WesleyKapow (changes requested):
Didn't review everything yet but wanted to stop here. See my note in condition schema about using `refine` or `superRefine` on the schema itself.

src/components/content/forms/condition-schema.tsx
melocule: Yeah I tried that initially, but .refine and .superRefine return a Zod.Effects instead of a Zod.Object which messes things up

src/components/core/alert.tsx
melocule: Hey! I took this from something Jesse sent me, let me just confirm this edit with him

src/components/core/main.scss
melocule: Hmm just discussed with Priyank about this, the select box is peculiar because the "v" icon is added automatically outside of our code and there doesn't seem to be a way to move it. We might need to keep this special styling and have the error icon just stay on the left of the selector. 

Any chance we can huddle on this tomorrow (if you want to take a look)?

src/components/content/forms/types.ts
```diff
@@ -1,3 +1,21 @@
...
+  | { success: false; data: undefined; errors: { [key: string]: string[] } };
+
+export type MedicationFormData = {
+  note?: string | undefined;
+  display?: string | undefined;
+  dosage?: string | undefined;
+  subjectID: string;
+  dateAsserted: Date;
+  rxNormCode: string;
+  status:
```
WesleyKapow, line 12: Can this be defined via fhir4?

WesleyKapow (approved):
Looks good! Left a few nitpicks.

src/components/content/forms/condition-schema.tsx
```diff
@@ -96,27 +96,53 @@ const sharedSchema = {
     "entered-in-error",
   ]),
   note: z.string().optional(),
+});
```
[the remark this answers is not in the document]
melocule: This makes extend not work unfortunately

WesleyKapow (approved):

PR #171, opened 2022-10-28
[CTW-399] Medication history component

WiktorKania:
Ticket: [CTW-399](https://zeushealth.atlassian.net/browse/CTW-399)

To do:
- apply styles from the [figma mockups](https://www.figma.com/file/BX8WkUMyecKimerE4VjdkB/Medication-Overview?node-id=682%3A6088)
- figure out how to access certain data
- move the `MedicationHistory` component to a different place (currently placed inside `MedicationDrawer` for convenience)


src/components/content/medication-history.tsx
```diff
@@ -1,72 +1,133 @@
...
-    //   const models = result.medications.map(
-    //     (medication) =>
-    //       new MedicationModel(medication, result.includedResources)
-    //   );
-    //   setMedications(models.map((model) => setupData(model)));
-    //   setLoading(false);
-    // }
+    (async () => {
+      const requestContext = await getRequestContext();
+      const res = await useMedicationHistory(
```
mrosata, line 32: The reason eslint is probably complaining here is that it treats any function that starts with `use*` as a react hook. Looking at the implementation of `useMedicationHistory` it's not a hook, maybe calling it `queryMedicationHistory` would be more apt (and possibly clear this eslint error).

mrosata: There's a function called `useQueryWithPatient` that can be used to make a hook that would use the `queryMedicationHistory` function (and you wouldn't need to get the patient or requestContext here in this function.

```ts
export function useQueryGetPatientMedicationHistory(
  rxNorm = ""
): UseQueryResult<SOME_TYPE_HERE, unknown> {
  return useQueryWithPatient(
    QUERY_KEY_MEDICATION_HISTORY,
    [rxNorm],
    queryMedicationHistory
  );
}
```
using that in a react component would abstract getting the patient and the request context, you wouldn't need to worry about those things in your code. You would just need to change the signature of `queryMedicationHistory` _(the function currently called useMedicationHistory)_ to take `(requestContext, patient, [rxNorm])`
```ts
export async function queryMedicationHistory(
  requestContext: CTWRequestContext
  patient: PatientModel,
  [rxNorm]: string[]
) {
  const patientUPID = patient.UPID
```
The only confusing part here is probably the `[rxNorm]` being in an array. It's just how the extra query keys are passed through from `useQueryWithPatient` atm. A working example may help here https://github.com/zeus-health/ctw-component-library/blob/103fbc406be7c8409826517b8f5af69df4ec205a/src/fhir/medications.ts#L62-L68

eric-gunther (approved):

PR #173, opened 2022-10-28
CTW-495 - fix pagination and add HOC

mrosata:


WesleyKapow:
Please add description to PR. What is being fixed? Why the HOC?

mrosata:
Was unable to type the HOC correctly. The type for the columns prop passed to `<Table column />` is meant to "extend MinTableRecord", but within the `withPagination` function it changed to require that it was "MinTableRecord" and neither I or Wesley could fix the TS validation.

This bug however was fixed in https://github.com/zeus-health/ctw-component-library/pull/174 so closing this PR.

mrosata:
Was unable to type the HOC correctly. The type for the columns prop passed to <Table column /> is meant to "extend MinTableRecord", but within the withPagination function it changed to require that it was "MinTableRecord" and neither I or Wesley could fix the TS validation.

This bug however was fixed in https://github.com/zeus-health/ctw-component-library/pull/174 so closing this PR.

PR #174, opened 2022-10-31
CTW-293/add storybook 7

WesleyKapow:
Add storybook 7 with stories for the following components:
* Table
* Drawer
* Spinner
* Data-List

Additionally fixed a few bugs around tables and pagination:
* Pagination now belongs to the table container so that any passed in `className` to table gets applied to pagination. This fixes a bug where passing in `ctw-w-[300px]` would reduce table width but not pagination width.
* Hide "Show All" button when already displaying all records (and display count is greater than number of records).
* Hide pagination while table `isLoading`.

# Screenshots

Screen Shot 2022-10-31 at 10 58 39 AM
Screen Shot 2022-10-31 at 10 59 03 AM


.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
mrosata, line 1: is this needed for a dependency that storybook required?

WesleyKapow: I had some problems with npm and storybook. I think storybook added this for me but I'll remove it and see if things break.

mrosata (approved):

.npmrc
WesleyKapow: Removed!

src/components/core/table/table.stories.tsx
```diff
@@ -65,19 +65,19 @@ const records: Record[] = [
...
-    records: records,
-    columns: columns,
+    records,
+    columns,
   },
 };
 
 const manyRecords = [...records];
-for (let i = records.length + 1; i <= 35; i++) {
+for (let i = records.length + 1; i <= 35; i += 1) {
```
mrosata, line 74: I just ran into this same eslint today too, threw me off

mrosata (approved):

src/components/core/table/table.stories.tsx
WesleyKapow: Ya.. I need to look it up again, seems silly to me.

PR #175, opened 2022-10-31
CTW-480 Adding fields to med details and updating logic for breaking out other providers meds

eric-gunther:
This PR adds a handful of fields required in our design for the medication details/history drawer. Excluding prescriber for the moment as it has some added complexity with the lens team.

Also refactoring our logic for splitting out "provider medications" and "other provider medications" and adding tests for some of the trickier logic.

image

**Design**
image


src/components/content/medications/provider-meds-table.tsx
```diff
@@ -35,8 +34,16 @@ export function ProviderMedsTable({
 
   useEffect(() => {
     if (!builderMedications) return;
-    setMedicationModels(sort(builderMedications, get(sortColumn), sortOrder));
-  }, [builderMedications, sortColumn, sortOrder]);
+    setMedicationModels(
+      sort(
+        showInactive
+          ? builderMedications
+          : builderMedications.filter((bm) => bm.status === "Active"),
```
eric-gunther, line 41: We had previously been making 2 calls to ODS to retrieve all of the builder medications but that was unnecessary since one of the calls was returning all of the builder meds and we can simply filter on the client side.

src/fhir/medication.ts
```diff
@@ -59,9 +60,23 @@ export function getRxNormCode(
     includedResources
   );
 
-  return find(codeableConcept?.coding, {
-    system: SYSTEM_RXNORM,
-  })?.code;
+  const excludedExtensions = ["ActiveIngredient", "BrandName"];
```
eric-gunther, line 63: This is a little hack but currently it's the best way of finding the "best" RxNorm code on a given med. The previous logic just using `find` was picking up the first one arbitrarily, I believe grabbing the active ingredient or brand name RxNorm code sometimes.

src/fhir/system-urls.ts
```diff
@@ -43,3 +43,14 @@ export const SYSTEM_ZUS_UNIVERSAL_ID =
...
+export const LENS_EXTENSION_MEDICATION_LAST_FILL_DATE =
+  "https://zusapi.com/lens/extension/medicationLastFillDate";
+export const LENS_EXTENSION_MEDICATION_LAST_PRESCRIBED_DATE =
+  "https://zusapi.com/lens/extension/medicationLastPrescribedDate";
+export const LENS_EXTENSION_MEDICATION_QUANTITY =
+  "https://zusapi.com/lens/extension/medicationQuantity";
+export const LENS_EXTENSION_MEDICATION_DAYS_SUPPLY =
+  "https://zusapi.com/lens/extension/medicationDaysSupply";
+export const LENS_EXTENSION_MEDICATION_REFILLS =
+  "https://zusapi.com/lens/extension/medicationRefills";
```
eric-gunther, line 56: Agreed w/ @e-m-albright on lens about these extension URLs.

src/fhir/medication.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        {
+          code: testRxNormCode,
+          system: SYSTEM_RXNORM,
+        },
+      ],
+    };
+
+    expect(getIdentifyingRxNormCode(baseMed)).toEqual(testRxNormCode);
+  });
+
```
mrosata, line 28: mutating an out of scope object in a test can cause issues. Here it's done in every test so it wouldn't cause problems but it's better use a pattern w/o side-effects. Could do one of these:
1. Clean up mutations between each test using something like `beforeEach`
2. Could use `...` operator
```ts

  test("gets RxNorm code from codeable concept", () => {
    const testMed = {
      ...baseMed,
      medicationCodeableConcept: {
      coding: [
        {
          code: testRxNormCode,
          system: SYSTEM_RXNORM,
        }],
      }
    };

```  
2. Use `set` function from "lodash/fp" (regular lodash set would be mutative)
```ts
const createTestMed = set("medicationCodeableConcept[0]", {
  resourceType: "MedicationStatement",
  status: "active",
  subject: {
    reference: "test",
  },
  medicationCodeableConcept: [],
} as MedicationStatement);

// Then to create a test med in tests
const testMed = createTestMed(
  code: testRxNormCode,
  system: SYSTEM_RXNORM,
});
```

src/fhir/medications.test.ts
```diff
@@ -0,0 +1,71 @@
...
+          ],
+        },
+      },
+    ];
+
+    const { builderMedications, otherProviderMedications } =
+      splitSummarizedMedications(summarizedMeds, builderMeds);
+
+    expect(builderMedications.length).toBe(1);
+    expect(otherProviderMedications.length).toBe(1);
```
mrosata, line 62: just so failures don't read "expected undefined to be 1" or "expected "2 to be 1"
```suggestion
    expect(builderMedications).toHaveLength(1);
    expect(otherProviderMedications).toHaveLength(1);
```

src/models/medication-statement.ts
```diff
@@ -127,7 +131,7 @@ export class MedicationStatementModel {
   }
 
   get status(): string {
-    return this.resource.status;
+    return capitalize(this.resource.status);
```
mrosata, line 134: if this is for styling the output of the html there is a `ctw-uppercase` className, but if it's for something other than display purposes this is fine.

src/components/content/medications-table-base.tsx
```diff
@@ -36,6 +36,30 @@ export const MedicationsTableBase = ({
         </>
       ),
     },
+    {
+      title: "Dispensed",
+      render: (medication: MedicationStatementModel) => (
```
mrosata, line 41: Don't need to do this now, but realized today myself that we don't need types here:
```suggestion
      render: (medication) => (
```

src/components/content/medications/other-provider-meds-table.tsx
```diff
@@ -1,21 +1,18 @@
...
-import { MedicationDrawer } from "@/components/content/medication-drawer";
+import { get } from "lodash/fp";
+import { useEffect, useState } from "react";
 
 type OtherProviderMedsTableProps = {
   className?: string;
   sortColumn?: keyof MedicationStatementModel;
   sortOrder?: "asc" | "desc";
-  // should inactive meds be shown?
-  showInactive?: boolean;
```
mrosata, line 14: What's the reason behind removing this from this table but not from the `provider-meds-table.tsx`? It might make more sense to leave it and then just always set it to true in the `patient-medications` so the two table components work the same?

mrosata (approved):

PR #183, opened 2022-11-01
Show verification status

melocule:
For the conditions table, the status column now shows both the clinical status and the verification status for faster viewing. 

Additionally renames the "Latest Status" column to "Status" and capitalizes the first letter of the status.

CTW Component Library

### DIFF
--- a/.changeset/rude-balloons-hug.md
+++ b/.changeset/rude-balloons-hug.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+For the conditions table, the status column now shows both the clinical status and the verification status for faster viewing.
--- a/src/components/content/conditions-table-base.tsx
+++ b/src/components/content/conditions-table-base.tsx
@@ -32,8 +32,16 @@ export function ConditionsTableBase({
       minWidth: 192,
     },
     {
-      title: "Latest Status",
-      dataIndex: "clinicalStatus",
+      title: "Status",
+      render: (row) => (
+        <div className="ctw-capitalize">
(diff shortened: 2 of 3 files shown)


.changeset/rude-balloons-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+For the conditions table, the status column now shows both the clinical status and the verification status for faster viewing.
```
pmadria3, line 5: How about "Update the status column to show both clinical status and verification status."

pmadria3:
Also, I had a ticket to update Status -> Latest Status, are we sure to revert that change?

pmadria3 (approved):
LGTM

src/components/content/conditions-table-base.tsx
```diff
@@ -32,8 +32,15 @@ export function ConditionsTableBase({
...
-      title: "Latest Status",
-      dataIndex: "clinicalStatus",
+      title: "Status",
+      render: (condition) => (
+        <div className="ctw-capitalize">
+          <p className="ctw-m-0">{condition.clinicalStatus}</p>
+          <p className="ctw-m-0 ctw-text-content-lighter">
+            {condition.verificationStatus}
+          </p>
+        </div>
```
WesleyKapow, line 42: 1. `<p>`'s can be problematic as they come with their own styles. Use `div` and that way you won't need `ctw-m-0`.
2. The design actually has the clinical status as darker (like condition name) and the verification status the same style as the other columns.

PR #185, opened 2022-11-02
CTW Misc Fixes

melocule:
- Fixes some red underlines on the dev side.
- Fixes the coloring of the fonts of the verification column to match the Figma.
- Uses divs instead of p's to prevent unwanted styling.
- The custom theme object was removed so that devs see the colors used in the tailwind config (which match the Figma).
image


WesleyKapow (approved):
LGTM

PR #188, opened 2022-11-03
[WIP][CTW-151] Add medications autocomplete

ada-cienciala:
Ticket: [CTW-151](https://zeushealth.atlassian.net/browse/CTW-151)

TODO:
- patient ID is undefined and the request is failing
- dosage select is not reset when medName is deleted
- dosage select should have empty value "-" before selecting an option
- `onMedicationNameChange` and `onMedicationDosageChange` could be merged to one handler and medName/dosage could be differentiated inside
- simplify `src/components/content/forms/form-field.tsx:54` (typescript onChange types problems)
- fix warning in the console (about uncontrolled combobox)
- changeset

### DIFF
--- a/src/components/content/forms/dosages-select.tsx
+++ b/src/components/content/forms/dosages-select.tsx
@@ -43,12 +43,12 @@ export const DosageSelect = ({
 
   const handleChange = (value: string) => {
     const dosage = options.find((el) => el.text === value);
-    if (dosage && onChange) onChange(dosage);
+    if (dosage) onChange?.(dosage);
   };
 
   return (
     <FormField
-      options={options.map((item) => item.text)}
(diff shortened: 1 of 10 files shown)


src/components/content/forms/dosages-select.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }, [getRequestContext, medName]);
+
+  const handleChange = (value: string) => {
+    const dosage = options.find((el) => el.text === value);
+    if (dosage) onChange?.(dosage);
+  };
+
+  return (
+    <FormField
+      options={["Select", ...(options?.map?.((item) => item.text) || [])]}
```
WiktorKania, line 51: Having `-` as the default value is not possible to set since we're using `startCase` from the lodash library which strips down these characters

PR #197, opened 2022-11-04
Node engines use gt 16 lt 19

mrosata:
Updated the node versions allowed and then ran `npm install` while on 18.11.0 to verify no issues on the engine check for us. Also verified it yelled if we used an engine outside the range.


WesleyKapow (dismissed):
LGTM

WesleyKapow (approved):

PR #199, opened 2022-11-07
CTW- 533 De-dupe med history

eric-gunther:
Adding de-duping logic to medication history so that we only show 1 record per date. This logic takes the first record per date, so it's possible we'll be choosing a record that has less useful information than another, but we can address that as it comes up during troubleshooting. This seems like a reasonable first step, especially as we saw so many dupes in early production tests that the history was hard to make sense of.

Screenshot with dupes
Pasted Graphic 1

Screenshot after change
albuterol sulfate 0 042  (1 25 MG3

### DIFF
--- a/.changeset/eleven-eels-try.md
+++ b/.changeset/eleven-eels-try.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Adding de-duping logic for medication history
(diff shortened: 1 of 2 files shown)


luthfib (dismissed):
LGTM. We probably want to denote somewhere for future reference, not sure where that this is the de-dedupe behavior. I can imagine a world where we ask why certain records arent showing up and we have to dig through the code

mrosata (dismissed):

PR #200, opened 2022-11-07
Ctw 512/hide delete in other patient records

luthfib:
CTW-512 
Hide delete in other patient records.


pmadria3 (approved):
Can you link the ticket in the description - ( CTW-XXX) makes it easier to PR review

LGTM

PR #201, opened 2022-11-07
Ctw 506/show notes previous text when editing

luthfib:
CTW-506
- If notes is populated it will now show when editing a condition (This should also happen when adding from other patient records - have to confirm if we want this). Notes can only be resized vertically now.

### DIFF
--- a/.changeset/slimy-moose-hide.md
+++ b/.changeset/slimy-moose-hide.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-If notes is populated it will now show when editing a condition. Notes can only be resized vertically now.
+If notes is pre-populated it will now show when editing a condition. Notes can only be resized vertically now.
--- a/.changeset/slimy-moose-hide.md
+++ b/.changeset/slimy-moose-hide.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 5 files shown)


.changeset/slimy-moose-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+If notes is populated it will now show when editing a condition. Notes can only be resized vertically now.
```
pmadria3, line 5: Maybe `Fix notes to show when editing a condition and can now be resized vertically.`

luthfib: Not sure if `Fix notes to show when editing a condition` is more clear. Also notes could be resized vertically before, but not can ONLY be resized vertically now.

luthfib: Maybe `If notes is pre-populated it will now show when editing a condition` ?

pmadria3: yeah that's better

src/components/content/forms/condition-schema.tsx
```diff
@@ -70,6 +70,7 @@ const sharedFields = (condition: ConditionModel) => [
   },
   {
     label: "Note",
+    value: condition.notes as unknown as string,
```
pmadria3, line 73: this should be `string[]` instead of `string` - probably can update FormEntry value field to be string or string[]?

pmadria3 (approved):

PR #205, opened 2022-11-08
CTW-522/css issues

WesleyKapow:
* Change how empty table message gets centered
* Put empty CSS vars in CTWProvider's div style
* Autocomplete changes
  * Make autocomplete use absolute positioned options.
  * Make autocomplete reopen when clicking in the input box.
 
Autocomplete with floating options
Screen Shot 2022-11-08 at 10 33 59 AM

### DIFF
--- a/src/styles/tailwind.theme.ts
+++ b/src/styles/tailwind.theme.ts
@@ -1,4 +1,10 @@
-import { Subset } from "../utils/typescript";
+// DO NOT IMPORT OUTSIDE THINGS WITHIN THIs FILE!
+// We compile this file and any imports will get compiled
+// to .js files which would then need to be cleaned up.
+
+type Subset<K> = {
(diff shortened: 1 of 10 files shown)


src/styles/tailwind.theme.ts
```diff
@@ -67,6 +67,40 @@ export const TailwindTheme = {
   },
 };
 
+// Some post processing workflows (older versions of PostCSS) will
```
luthfib, line 70: Kinda sucks postcss does this

luthfib (dismissed):

luthfib:
I think you have some prettier errors, fix those and add changeset, and LGTM

src/styles/tailwind.theme.ts
WesleyKapow: Old versions.. but figure best to get ahead of this and just add this work around instead of potentially hitting this again in the future.

pmadria3 (approved):
LGTM

PR #206, opened 2022-11-08
await queries to avoid drawer closing before time

luthfib:
CTW-507

Fix delayed UX when adding condition.

### DIFF
--- a/src/components/content/forms/medications.ts
+++ b/src/components/content/forms/medications.ts
@@ -99,9 +99,11 @@ export const createMedicationStatement = async (
     result.success = false;
   }
 
-  QUERY_KEYS.forEach(async (queryKey) => {
-    await queryClient.invalidateQueries([queryKey]);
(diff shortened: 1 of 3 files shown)


src/components/content/forms/medications.ts
```diff
@@ -99,9 +99,11 @@ export const createMedicationStatement = async (
     result.success = false;
   }
 
-  QUERY_KEYS.forEach(async (queryKey) => {
-    await queryClient.invalidateQueries([queryKey]);
-  });
+  await Promise.all(
+    QUERY_KEYS.map(async (queryKey) => {
+      await queryClient.invalidateQueries([queryKey]);
```
WesleyKapow, line 104: This `await` still makes it in series instead of parallel. And you'll want to do similar for the two awaits you do in conditions.

WesleyKapow (approved):

PR #208, opened 2022-11-08
Add Patient History hiding features

pmadria3:
[CTW-471](https://zeushealth.atlassian.net/jira/software/projects/CTW/boards/49?selectedIssue=CTW-471)

This PR sets up the Patient History in which patients that have had their history fetched, it is shown otherwise shows that they need to request records. 

Records has been fetched before: 

Screen Shot 2022-11-21 at 1 53 37 PM

Record has not been fetched before: 

Screen Shot 2022-11-21 at 1 59 33 PM


src/fhir/types.ts
```diff
@@ -12,3 +12,23 @@ export type ResourceType<T extends ResourceTypeString> = Extract<
...
+
+export type PatientRefreshHistoryMessage = {
+  status: "initialize" | "in_progress" | "done" | "error";
+  messageUuid: string;
+  initialData: {
+    patientId: string;
+  };
+  createdAt: string;
+  updatedAt: string;
+};
```
WesleyKapow, line 34: These aren't fhir types right? Should probably live with patient history stuff.

src/components/content/conditions.tsx
```diff
@@ -240,38 +264,55 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
...
+            <RetrievePatientHistory message="Retrieve patient clinical history.">
+              <button
+                type="button"
+                className="ctw-btn-clear ctw-link"
+                onClick={handleClinicalHistory}
+              >
+                Request Records
+              </button>
+            </RetrievePatientHistory>
+          )}
```
WesleyKapow, line 315: I think all of this can be pushed into the `RetrievePatientHistory` component. You also don't want the onClick to trigger `handleClinicalHistory`. Instead the onClick will open the future history request drawer.

src/components/core/patient-history/patient-history-message.tsx
```diff
@@ -0,0 +1,21 @@
+import { ReactNode } from "react";
+
+export type RetrievePatientHistoryProps = {
+  message?: string;
+  children?: ReactNode;
+};
+
+export const RetrievePatientHistory = ({
```
WesleyKapow, line 8: 1. Component name should match file name.
2. This is not a "core" component. I need to re-organize these but the thought is "core" components are for our building blocks and "content" components are for specific use cases, like conditions, medications, patient history. Another way of thinking about it is "core" components could be used in a non CTW application. They are generic and have nothing to do with healthcare or CTW/profile.

src/components/core/patient-history/patient-history-message.tsx
```diff
@@ -0,0 +1,21 @@
...
+  message,
+  children,
+}: RetrievePatientHistoryProps) => (
+  <div className="ctw-body-container">
```
WesleyKapow, line 12: Don't use this class. Inline whatever it is you need.

src/components/core/patient-history/patient-history-message.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <div className="ctw-flex ctw-justify-center ctw-space-y-3 ctw-space-x-2">
+      <div className="ctw-title-container" />
```
WesleyKapow, line 14: Don't use this class and this div is empty/useless.

src/components/content/patient-history/patient-history-types.ts
```diff
@@ -0,0 +1,19 @@
+export type RawPatientRefreshHistoryMessage = {
```
WesleyKapow, line 1: Move this and `patient-history.ts` into `src/services/patient-history/`. I'll later refactor `fhir` to be moved into this new `services` directory.

src/components/content/patient-history/patient-history-types.ts
```diff
@@ -0,0 +1,19 @@
...
+  status: "initialize" | "in_progress" | "done" | "error";
+  messageUuid: string;
+  initialData: {
+    patientId: string;
+  };
+  _createdAt: string;
+  _updatedAt: string;
+};
+
+export type PatientRefreshHistoryMessage = {
```
WesleyKapow, line 11: These two are so similar, why do we need both?

src/components/content/patient-history/patient-history.ts
```diff
@@ -0,0 +1,191 @@
...
+  const endpointUrl = `${getZusApiBaseUrl(
+    requestContext.env
+  )}/patient-history/messages?patient_id=${patientID}`;
+
+  try {
+    const response = await fetch(endpointUrl, {
+      headers: {
+        Authorization: `Bearer ${requestContext.authToken}`,
+        "practitioner-npi": "1245319599",
+        "practitioner-role": "Test Provider",
```
WesleyKapow, line 22: Oh dear, do we really need these? We won't have these!

src/components/content/patient-history/patient-history.ts
```diff
@@ -0,0 +1,191 @@
...
+      },
+    });
+    const result = await response.json();
+    return (
+      Object.values(result.data) as RawPatientRefreshHistoryMessage[]
+    ).map((entry: RawPatientRefreshHistoryMessage) => ({
```
WesleyKapow, line 28: What's the purpose of this mapping?

src/components/content/conditions.tsx
```diff
@@ -133,6 +139,24 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
...
+    );
+    if (message.data[0].status === "done") {
+      setClinicalHistoryExists(true);
+    } else {
+      setClinicalHistoryExists(false);
+      setRequestRecordsClinicalHistory(true);
+    }
+  };
+
+  handleClinicalHistory();
```
WesleyKapow, line 158: This should be done in a `useEffect`, otherwise this will get called many many times, like when the component re-renders.

WesleyKapow (changes requested):
Add screenshot of new UI!

src/components/content/conditions.tsx
```diff
@@ -303,15 +302,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
...
-            <RetrievePatientHistory message="Retrieve patient clinical history.">
-              <button
-                type="button"
-                className="ctw-btn-clear ctw-link"
-                onClick={handleClinicalHistory}
-              >
-                Request Records
-              </button>
-            </RetrievePatientHistory>
+            <PatientHistoryMessage message="Retrieve patient clinical history." />
```
WesleyKapow, line 305: Do we ever use a different `message` for this component? If not push it into `PatientHistoryMessage`.

src/services/patient-history/patient-history.ts
```diff
@@ -0,0 +1,54 @@
...
+  const endpointUrl = `${getZusApiBaseUrl(
+    requestContext.env
+  )}/patient-history/messages?patient_id=${patientID}`;
+
+  try {
+    const response = await fetch(endpointUrl, {
+      headers: {
+        Authorization: `Bearer ${requestContext.authToken}`,
+        "practitioner-npi": "1245319599",
+        "practitioner-role": "Test Provider",
```
WesleyKapow, line 20: TODO

src/services/patient-history/patient-history.ts
```diff
@@ -0,0 +1,54 @@
...
+      },
+    });
+    const result = await response.json();
+
+    /* eslint no-underscore-dangle: 0 */
```
WesleyKapow, line 25: Don't think this is needed.

src/services/patient-history/patient-history.ts
```diff
@@ -0,0 +1,54 @@
...
+    return Object.values(result.data) as PatientRefreshHistoryMessage[];
+  } catch (err) {
+    throw errorResponse(
+      "Failed fetching patient refresh history messages",
+      err
+    );
+  }
+}
+
+export async function getLatestPatientRefreshHistoryMessage(
```
WesleyKapow, line 35: How do you imagine this function will be used in other situations? Why this function and not something like `hasFetchedPatientHistory` that returns a `Promise<boolean>`?

WesleyKapow (dismissed):

src/components/content/conditions/story-helpers/mocks/patient-history-message.ts
```diff
@@ -0,0 +1,26 @@
+export const patientHistoryMessage = {
+  data: {
+    0: {
```
WesleyKapow, line 3: Do they really return a data object with `0` as a key instead of an array?

src/components/content/patient-history/patient-history-message.tsx
```diff
@@ -1,16 +1,14 @@
...
+export const PatientHistoryMessage = ({ children }: PatientHistoryProps) => (
   <div className="ctw-flex ctw-justify-center ctw-space-x-2 ctw-border ctw-border-solid ctw-border-divider-light ctw-p-5">
     <div className="ctw-flex ctw-justify-center ctw-text-icon-default">
-      {message}
+      Retrieve patient clinical history.
     </div>
-    <button type="button" className="ctw-btn-clear ctw-link">
-      Request Records
-    </button>
+    <div>{children}</div>
```
WesleyKapow, line 12: Do we vary what we pass in as children here or is it always the `Request Records` button? If it doesn't vary, let's not make it configurable -- YAGNI!

src/services/patient-history/patient-history.ts
```diff
@@ -32,23 +30,30 @@ export async function getPatientRefreshHistoryMessages(
   }
 }
 
-export async function getLatestPatientRefreshHistoryMessage(
+export async function hasFetchedPatientHistory(
   requestContext: CTWRequestContext,
   patientID: string,
   statuses: string[]
```
WesleyKapow, line 36: Why does this need to take `statuses`?

WesleyKapow (changes requested):

src/components/content/conditions/story-helpers/mocks/patient-history-message.ts
pmadria3: Yep that's what the response looks like 

Screen Shot 2022-11-29 at 1 11 18 PM

src/services/patient-history/patient-history.ts
pmadria3: Originally had it as an array since there can be multiple but I only need to worry about done so changed it - good catch!

src/components/content/conditions/story-helpers/mocks/patient-history-message.ts
WesleyKapow: yikes ok

src/services/patient-history/patient-history.ts
WesleyKapow: Why does this need to take a status at all?

WesleyKapow: What does it mean for a caller of this function to pass in something other than `done`? In english what is this function's contract? E.g. "Hey I'll let you know if the user has had any patient history data already fetched" OR "hey I'll let you know if the user has had any patient history fetched with the given status"? Which do we want?

pmadria3: ah good point there is no need to pass in anything

src/components/content/conditions.tsx
```diff
@@ -145,8 +145,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
 
     const patientHistoryFetched = await hasFetchedPatientHistory(
       requestContext,
-      patientResponse.data?.id as string,
-      ["done"]
+      patientResponse.data?.id as string
```
WesleyKapow, line 148: hmm also just noticed this `as string`.. we shouldn't lie! Looks like this could be `undefined`. What happens then?

src/components/content/conditions.tsx
```diff
@@ -319,15 +318,9 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
...
-              <button
-                type="button"
-                className="ctw-btn-clear ctw-link"
-                onClick={() => setRequestDrawerisOpen(true)}
-              >
-                Request Records
-              </button>
-            </PatientHistoryMessage>
+            <PatientHistoryMessage
+              onClick={() => setRequestDrawerisOpen(true)}
```
WesleyKapow, line 322: s/DrawerisOpen/DrawerIsOpen (capitalize the `i`)

WesleyKapow (dismissed):

WesleyKapow (approved):

PR #211, opened 2022-11-08
CTW-349 Capitalizing fields in details/history drawer

eric-gunther:
Capitalizing various fields in the details/history drawers. I figured there was 3 options here: 

1. Do it generically in the underlying core components with something like a `capitalize` boolean prop
2. Do it in the getters
3. Do it when setting up data to pass to the component for presentation

I chose option 3 because it felt like this kind of decision belongs close to the presentation layer. The underlying data may not be capitalized, and this is likely a case-by-case decision (eg. we're not going to capitalize the "h" at the beginning of a URL), so I thought this was the appropriate place to put this logic.

https://user-images.githubusercontent.com/97455910/200661327-a865a375-0b9a-498b-a996-4b35cd27ae10.mov

### DIFF
--- a/.changeset/pink-hairs-call.md
+++ b/.changeset/pink-hairs-call.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Capitalizing fields in details/history drawers
+Capitalize label and fields in detail and history drawers.
--- a/.changeset/pink-hairs-call.md
+++ b/.changeset/pink-hairs-call.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Capitalizing fields in details/history drawers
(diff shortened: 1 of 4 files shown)


.changeset/pink-hairs-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Capitalizing fields in details/history drawers
```
pmadria3, line 5: nit: maybe change to `Capitalize label and fields in detail and history drawers.`

pmadria3 (approved):
small nit otherwise LGTM

luthfib:
Do we also want to capitalize fields in the tables rows?

WesleyKapow (approved):
LGTM

Reminder -- We can use CSS to capitalize when we want the element to always be capitalized.

eric-gunther:
> Do we also want to capitalize fields in the tables rows? I noticed some of the medications names arent capitalized

I'm not sure about this. This capitalization inconsistency does appear to cause sorting problems for us, but these are the true display strings coming from our terminology resources. I'm not sure why some are capitalized and some are not, but that's possibly an intentional decision on the part of RxNorm, so I think it would be hasty to transform that data at this time.

PR #212, opened 2022-11-08
Ctw 543/fix edit add condition bug

luthfib:
CTW-543

Clicking add condition from other provider records after clicking edit condition does not result in an error anymore.

### DIFF
--- a/.changeset/rude-items-hear.md
+++ b/.changeset/rude-items-hear.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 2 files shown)


WesleyKapow (approved):

WesleyKapow:
This looks really clear but just FYI, good to include in the PR description what the cause of the bug was.

luthfib:
Sounds

> This looks really clear but just FYI, good to include in the PR description what the cause of the bug was.

Sounds good. I had included it in the ticket with a video demonstration, but I'll include it in the pr description as well going forward.

PR #214, opened 2022-11-09
Ctw 538/delete condition enters in error instead of delete

luthfib:
CTW-538

Delete functionality for validation status now edits to entered-in-error instead of deleting.

### DIFF
--- a/src/components/content/conditions-helper.ts
+++ b/src/components/content/conditions-helper.ts
@@ -23,6 +23,7 @@ export const onConditionDelete = async (
       },
     ],
   };
+  // We have to delete clinical status because it can't be present if verification is 'entered-in-error'
   delete clonedResource.clinicalStatus;
 
   const response = (await createOrEditFhirResource(
--- a/src/components/content/conditions-helper.ts
+++ b/src/components/content/conditions-helper.ts
@@ -1,8 +1,5 @@
 import { createOrEditFhirResource } from "@/fhir/action-helper";
-import {
-  QUERY_KEY_OTHER_PROVIDER_CONDITIONS,
-  QUERY_KEY_PATIENT_CONDITIONS,
-} from "@/utils/query-keys";
+import { QUERY_KEY_OTHER_PROVIDER_CONDITIONS } from "@/utils/query-keys";
 import { queryClient } from "@/utils/request";
 import Client, { FhirResource } from "fhir-kit-client";
 import { Resource } from "fhir/r4";
@@ -24,8 +21,5 @@ export const onConditionDelete = async (
     throw new Error(`Failed to edit resource with id of ${resource.id}`);
   }
 
-  await Promise.all([
-    queryClient.invalidateQueries([QUERY_KEY_PATIENT_CONDITIONS]),
-    queryClient.invalidateQueries([QUERY_KEY_OTHER_PROVIDER_CONDITIONS]),
-  ]);
+  await queryClient.invalidateQueries([QUERY_KEY_OTHER_PROVIDER_CONDITIONS]);
 };
--- a/src/components/content/conditions-helper.ts
+++ b/src/components/content/conditions-helper.ts
@@ -18,7 +18,7 @@ export const onConditionDelete = async (
     resourceToDelete.verificationStatus.coding
   ) {
     resourceToDelete.verificationStatus.coding[0].code = "entered-in-error";
-    // We have to delete clinical status because it may not be present if verification is 'entered-in-error'
+    // We have to delete clinical status because it can't be present if verification is 'entered-in-error'
     delete resourceToDelete.clinicalStatus;
   }
 
--- a/src/components/content/conditions-helper.ts
+++ b/src/components/content/conditions-helper.ts
@@ -1,15 +1,25 @@
 import { createOrEditFhirResource } from "@/fhir/action-helper";
-import { QUERY_KEY_OTHER_PROVIDER_CONDITIONS } from "@/utils/query-keys";
+import { QUERY_KEY_PATIENT_CONDITIONS } from "@/utils/query-keys";
 import { queryClient } from "@/utils/request";
 import Client, { FhirResource } from "fhir-kit-client";
-import { Resource } from "fhir/r4";
 
 export const onConditionDelete = async (
-  resource: Resource,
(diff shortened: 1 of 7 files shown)


WesleyKapow (changes requested):
The modal edit module still has `onDelete` and talks about removal.

I suggest leave it as it was w.r.t being a delete modal (and not an edit modal).

Have the modal NOT do the actual delete part and instead invoke `onDelete` and `await` that before closing.

The conditions `onDelete` should take the current condition resource, set the verification status and save that. That way you don't lose any parts of the resource.

src/components/content/conditions-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+  QUERY_KEY_PATIENT_CONDITIONS,
+} from "@/utils/query-keys";
+import { queryClient } from "@/utils/request";
+import Client, { FhirResource } from "fhir-kit-client";
+import { Resource } from "fhir/r4";
+
+export const onConditionDelete = async (
+  resource: Resource,
+  fhirClient: Client
+) => {
```
WesleyKapow, line 13: Have this function take a `fhir4.Condition` and set the clinicalStatus to `entered-in-error`.

src/components/content/conditions.tsx
```diff
@@ -317,17 +316,22 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
...
-            ]);
+          onDelete={async () => {
+            const requestContext = await getRequestContext();
+            await onConditionDelete(
+              await getDeleteConditionFhirResource(
+                selectedCondition,
+                patientResponse.data.id,
+                requestContext
+              ),
+              requestContext.fhirClient
```
WesleyKapow, line 333: Have `onConditionDelete` handle everything.

src/components/content/conditions-helper.ts
```diff
@@ -0,0 +1,31 @@
...
+    fhirClient
+  )) as FhirResource;
+
+  if (!response.id) {
+    throw new Error(`Failed to edit resource with id of ${resource.id}`);
+  }
+
+  await Promise.all([
+    queryClient.invalidateQueries([QUERY_KEY_PATIENT_CONDITIONS]),
+    queryClient.invalidateQueries([QUERY_KEY_OTHER_PROVIDER_CONDITIONS]),
```
WesleyKapow, line 29: Shouldn't need to re-query other provider conditions.

src/fhir/conditions.ts
```diff
@@ -188,3 +193,46 @@ export const filterConditionsWithConfirmedCodes = (
...
+  const display = practitioner.fullName;
+
+  return {
+    reference: `Practitioner/${practitionerId}`,
+    type: "Practitioner",
+    display,
+  };
+};
+
+export const getDeleteConditionFhirResource = async (
```
WesleyKapow, line 211: Should be able to remove this. The idea is we should take the raw `fhir4.Condition` resource, set the clinicalStatus and save that.

WesleyKapow (changes requested):

src/components/content/conditions-helper.ts
WesleyKapow: `conditionModel.resource`. All of our fhir models wrap an actual fhir resource.

src/components/content/conditions-helper.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const onConditionDelete = async (
+  resource: fhir4.Condition,
+  fhirClient: Client
+) => {
+  if (!resource.id) {
+    throw new Error("Tried to edit a resource that hasn't been created yet.");
+  }
+
+  const resourceToDelete = { ...resource };
```
WesleyKapow, line 14: ??? This so you don't modify the existing one? Use `cloneDeep` if you want to do that. Otherwise what will happen is you will create an object where all of the top level key/values are cloned but any nested are not. So in this case, you are in fact still modifying the `verificationStatus.coding` that exists on the original.

src/components/content/conditions-helper.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (
+    resourceToDelete.verificationStatus &&
+    resourceToDelete.verificationStatus.coding
+  ) {
```
WesleyKapow, line 19: This guard needed? What if there is no verificationStatus? We still want to set it to `entered-in-error` right? You can probably avoid the if and just ALWAYS setup `verificationStatus`. Something like `verificationStatus = {coding: [{code: ..., system}]}`

src/components/content/conditions-helper.ts
WesleyKapow: Also name implies you are going to delete the resource. While this is `onConditionDelete`, we aren't actually deleting it.

WesleyKapow (changes requested):

WesleyKapow (approved):
LGTM

I do wonder if there is a reason you wanted to clone the resource. Would it be a problem to set it directly? would that have impacted some display in the FE before it was really set/saved?

src/components/content/conditions-helper.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const onConditionDelete = async (
+  resource: fhir4.Condition,
+  fhirClient: Client
+) => {
+  if (!resource.id) {
+    throw new Error("Tried to edit a resource that hasn't been created yet.");
+  }
+
+  const clonedResource = cloneDeep(resource);
```
melocule, line 16: Hey Luthfi! Can I ask why the resource needs to be cloned here instead of modifying it directly?

melocule (changes requested):

src/components/content/conditions-helper.ts
luthfib: Yes, we can’t modify params. One of our eslint, usually because it can cause problems, so we clone it to avoid doing that.

luthfib: https://eslint.org/docs/latest/rules/no-param-reassign

WesleyKapow: Ah cool, didn't realize that guarded nested re-assignments too. YAY!

melocule (approved):

WesleyKapow (approved):

PR #216, opened 2022-11-09
CTW-526/Create ActionList (match canvas UI)

mrosata:
Creates an `<ActionList/>` component which will look something like this:
Screen Shot 2022-11-09 at 12 18 44 PM

And in the wild:
Screen Shot 2022-11-09 at 12 27 48 PM

Notable attributes:
List items marked "active" will show a (primary) colored border to the left by default and when hovered present a button to take action on that item. Optionally the opposite can be done for inactive items if an `onUndoAction` handler is passed in, but that is not a requirement.

It is the duty of the data owner to set the "active" property on their item through their custom "onAction" handler. The list is simply rendering state and wiring up the event listeners.

There is an optional "onRowClick" handler that will take action if the row is clicked or enter button is pushed while row is focused. It is the responsibility of the code-base utilizing this component to create a style for the focus state using CSS. We are not implementing focused css styles or class-names at the time. The keyboard handler here is less about accessibility than it is about ensuring we're not actively preventing consumers of the library from building accessible applications using Zus components.

### DIFF
--- a/.changeset/gentle-cameras-design.md
+++ b/.changeset/gentle-cameras-design.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Added new component <ActionList/>
+Add new component <ActionList/>
--- a/.changeset/gentle-cameras-design.md
+++ b/.changeset/gentle-cameras-design.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Added new component <ActionList/>
--- a/src/components/core/action-list/action-list.scss
+++ b/src/components/core/action-list/action-list.scss
@@ -17,18 +17,16 @@
     // because we will use "after" to create colors on the left border, we need a transparent border here so content
     // will respect layout, as content would shift under the border of an absolutely positioned "after" element.
     border-left: 0.5rem solid transparent;
-    &:after {
+    &:before {
       border-left: 0.5rem solid theme("colors.icon.active");
       content: " ";
       position: absolute;
-      height: 100%;
-      width: 100%;
-      top: 0;
-      bottom: -10px;
+      top: -1px;
+      bottom: -1px;
       // Move left to sit over the transparent border on the list item
       left: -0.5rem;
     }
-    &:not(.active):after {
+    &:not(.active):before {
       border-left: 0.5rem solid theme("colors.divider.light");
     }
 
--- a/src/components/core/action-list/action-list.scss
+++ b/src/components/core/action-list/action-list.scss
@@ -5,6 +5,7 @@
   overflow: hidden; // so the left padding colored on items doesn't break border-radius
 
   .ctw-action-list-item {
+    position: relative;
     /* Inside auto layout */
     margin: -1px 0;
     width: 100%;
@@ -12,8 +13,22 @@
     // Colors
     background: theme("colors.white");
     border: 1px solid theme("colors.divider.light");
-    border-left: 0.5rem solid theme("colors.icon.active");
-    &:not(.active) {
+
+    // because we will use "after" to create colors on the left border, we need a transparent border here so content
+    // will respect layout, as content would shift under the border of an absolutely positioned "after" element.
+    border-left: 0.5rem solid transparent;
+    &:after {
+      border-left: 0.5rem solid theme("colors.icon.active");
+      content: " ";
+      position: absolute;
+      height: 100%;
+      width: 100%;
+      top: 0;
+      bottom: -10px;
+      // Move left to sit over the transparent border on the list item
+      left: -0.5rem;
+    }
+    &:not(.active):after {
       border-left: 0.5rem solid theme("colors.divider.light");
     }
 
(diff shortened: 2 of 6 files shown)


src/components/core/action-list/action-list.tsx
```diff
@@ -0,0 +1,103 @@
...
+    role="row"
+    className={cx(
+      "ctw-action-list-item",
+      "ctw-border-lighter ctw-flex ctw-cursor-pointer ctw-p-4",
+      {
+        [activeClassName]: !item.complete,
+        undoable: isFunction(onUndoAction),
+      }
+    )}
+    onKeyDown={(evt) => {
```
luthfib, line 63: nit: best to just spell out the full name so evt -> event

luthfib (changes requested):
Looking good. 

I think for the button sizes we can probably just set a size to avoid cases like this: image

Also noticed this funky thing going on with the borders:

image

Where they look slanted. 


Also do we have a figma we can refer to as well?

mrosata:
@luthfib 
> I think for the button sizes we can probably just set a size to avoid cases like this:
I had a min-width on the element wrapping the button but removed it mainly because it only matters when the list item content is long and depending on where the component is used the list items will have different widths (and possibly different sized buttons).
The question is should we set a min-width in the component-library and only have consumers of the library override it if it doesn't fit their use case? (ie: in canvas I will need to set that min-width property, you can actually see it in the image with the brown colors there is a min-width set)... or should we not be prescriptive of the min-width on that item (like currently in this PR), in which case consumers of the library will still possibly need to set it as canvas will.

.changeset/gentle-cameras-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Added new component <ActionList/>
```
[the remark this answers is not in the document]
mrosata:
```suggestion
"@zus-health/ctw-component-library": minor
---

Add new component <ActionList/>
```

src/components/core/action-list/action-list.tsx
mrosata:
```suggestion
    onKeyDown={(event) => {
```

src/components/core/action-list/action-list.tsx
```diff
@@ -0,0 +1,103 @@
...
+      {item.subtitle && <div className="ctw-font-light">{item.subtitle}</div>}
+    </div>
+    <div className="ctw-action-list-item-action">
+      {!item.complete && (
+        <button
+          type="button"
+          className="ctw-btn-primary"
+          onClick={(evt) => {
+            evt.stopPropagation();
+            onAction(item);
```
luthfib, line 82: You'll want to update any references to `evt`

mrosata:
@luthfib
> Also do we have a figma we can refer to as well?
There is a figma for how this should look in canvas https://www.figma.com/file/BX8WkUMyecKimerE4VjdkB/Medication-Overview?node-id=984%3A6753
I brought the weirdness with the lines up to @eric-gunther yesterday and we landed on not spending time working out that spacing for a first pass. If you know off the top of your head the way to eliminate that it would be good to not have it (maybe using ::after content instead of border colors, idk. It has to do with the way css does borders, there will always be an angle where two non-zero width borders meet

src/components/core/action-list/action-list.scss
```diff
@@ -0,0 +1,36 @@
...
+    margin: -1px 0;
+    width: 100%;
+
+    // Colors
+    background: theme("colors.white");
+    border: 1px solid theme("colors.divider.light");
+    border-left: 0.5rem solid theme("colors.icon.active");
+    &:not(.active) {
+      border-left: 0.5rem solid theme("colors.divider.light");
+    }
```
luthfib, line 18: I think you can do:

```
 // Colors
    background: theme("colors.white");
    border-bottom: 1px solid theme("colors.divider.light");
    border-top: 1px solid theme("colors.divider.light");
    background: linear-gradient(theme("colors.icon.active"), theme("colors.icon.active")) left/.5rem 100% no-repeat;
    margin-bottom: 0px;
    
    &:not(.active) {
      background: none;
    }

```

luthfib: You may have to add more left padding since this wont push content like before

luthfib:
> @luthfib
> 
> > I think for the button sizes we can probably just set a size to avoid cases like this:
> > I had a min-width on the element wrapping the button but removed it mainly because it only matters when the list item content is long and depending on where the component is used the list items will have different widths (and possibly different sized buttons).
> > The question is should we set a min-width in the component-library and only have consumers of the library override it if it doesn't fit their use case? (ie: in canvas I will need to set that min-width property, you can actually see it in the image with the brown colors there is a min-width set)... or should we not be prescriptive of the min-width on that item (like currently in this PR), in which case consumers of the library will still possibly need to set it as canvas will.

Sounds good. I think the main weirdness is that the button should have enough space to not wrap like it is doing, but maybe we need to play around with it. For first pass Im fine with leaving as is.

mrosata:
@luthfib This is as good as I can get the side colors without going too crazy. It doesn't sit over the outermost border, but made the current border transparent and then using `:before` to add an absolute positioned empty element and then putting the left border on that and shifting it over to sit on top of the transparent border.

Screen Shot 2022-11-09 at 3 07 51 PM

luthfib (approved):

PR #218, opened 2022-11-09
CTW-515/improve-onset-and-abatement

pmadria3:
[CTW-515](https://zeushealth.atlassian.net/jira/software/projects/CTW/boards/49?selectedIssue=CTW-515)

This PR reduces verification status options for adding conditions and does some form validation on hitting submit for the date fields. 

Screen Shot 2022-11-14 at 1 54 32 PM

Fixed a bug where the value in the date field was getting reset on error. 

Screen Shot 2022-11-14 at 2 29 24 PM


pmadria3:
Right now - I'm throwing an error but am open for suggestions on what we want to do @WesleyKapow

src/components/content/forms/condition-schema.tsx
```diff
@@ -140,6 +140,7 @@ export const conditionAddSchema = conditionSchema
         required_error: "Please choose a condition.",
       }),
     }),
+    verificationStatus: z.enum(["confirmed", "unconfirmed"]),
```
WesleyKapow, line 143: Interesting, `conditionSchema` already has `verificationStatus`, I guess this just overwrites it? Maybe more clarity, move it out of the shared `conditionSchema` and put it in the conditionEditSchema.

src/components/content/forms/form-field.tsx
```diff
@@ -5,7 +5,7 @@ import { startCase } from "lodash";
 import type { InputHTMLAttributes } from "react";
 
 export type FormFieldProps = {
-  errors?: string[];
+  errors?: string | string[];
```
WesleyKapow, line 8: Undo these changes, this should continue taking an array of errors.

Being able to pass a string for `errors` has a few issues:
1. It's called `errors` (plural) implying it is an array.
2. The bigger issue is we're treating it as an array. You're even doing `as string[]` below. And as a reminder, you RARELY should ever use `as` in typescript. It's only when we know something has to be a certain type but the TS compiler doesn't.

src/components/content/forms/drawer-form.tsx
```diff
@@ -50,6 +51,29 @@ export const DrawerForm = <T,>({
...
+
+    Object.entries(errorMessagesArray).forEach(([key, value]) => {
+      if (!value) {
+        setIsSubmitting(false);
+        setErrors({
+          isInvalid: true,
+        });
+        throw Error(`There is a problem with your field ${key}`);
+      }
+    });
```
WesleyKapow, line 75: We don't need the `isInvalid` state. There's already state for handling and displaying errors, `setErrors`. This is why I had you console.log other form errors to see what `setErrors` should look like. You should loop over the input elements and for each one that is invalid, add an error for that key to `inputErrors`. Then, if `inputErrors` is not empty, `setErrors(inputErrors)` and `return`.

src/components/content/forms/drawer-form.tsx
```diff
@@ -50,6 +51,29 @@ export const DrawerForm = <T,>({
     event.preventDefault();
     setIsSubmitting(true);
     const form = event.target;
+
+    const formDataValidation = (event.target as HTMLElement).querySelectorAll(
+      "input"
+    );
+    const formDataValidationToArray = Array.from(formDataValidation);
```
WesleyKapow, line 58:
```suggestion
    const inputs = Array.from((event.target as HTMLElement).querySelectorAll(
      "input"
    ));
```

`formDataValidation` and `formDataValidationToArray` are really awkward names. `formDataValidation` does not point to any type of validation, it points to inputs. `formDataValidationToArray` doesn't point to anything that will transform it, e.g. `to array` it. It instead points to an array. Generally it is good practice to not have the type included in the name variables as we can easily see the type. You also do all of this in one pass to avoid the extra variable, and just call it `inputs` or `inputElements` to make it clear these are DOM elements.

WesleyKapow (changes requested):

src/components/content/forms/drawer-form.tsx
```diff
@@ -52,28 +52,31 @@ export const DrawerForm = <T,>({
...
     Object.entries(errorMessagesArray).forEach(([key, value]) => {
       if (!value) {
-        setIsSubmitting(false);
-        setErrors({
-          isInvalid: true,
-        });
-        throw Error(`There is a problem with your field ${key}`);
+        inputErrors[key] = [`The value is not valid for ${key}`];
       }
     });
```
WesleyKapow: You don't need `errorMessagesArray`. Populate `inputErrors` in a single pass.

src/components/content/forms/drawer-form.tsx
```diff
@@ -52,28 +52,31 @@ export const DrawerForm = <T,>({
...
 
+    if (Object.keys(inputErrors).length) {
```
WesleyKapow, line 74: `!isEmpty` from lodash maybe? Would make this line a bit more explicit of the intent.

package-lock.json
```diff
@@ -3772,6 +3772,7 @@
       "integrity": "sha512-5dgndWOriYSm5cnYaJNhalLNDKOqFwyDB/rr1E9ZsGciGvKPs8R2xYGCacuf3z6K1YKDz182fd+fY3cn3pMqXQ==",
```
WesleyKapow, line 3772: You shouldn't have any package-lock.json changes when you didn't make any package.json changes (unless you are explicitly wanting to upgrade a lib version). This likely happened because you deleted the lock file and ran `npm install`. See if you can get things working right with reverting this change.

WesleyKapow (changes requested):

src/components/content/forms/drawer-form.tsx
```diff
@@ -56,22 +57,15 @@ export const DrawerForm = <T,>({
...
-      errorMessagesArray[c.name] = c.checkValidity();
-      return errorMessagesArray;
-    });
-
     const inputErrors: InputError = {};
 
-    Object.entries(errorMessagesArray).forEach(([key, value]) => {
-      if (!value) {
-        inputErrors[key] = [`The value is not valid for ${key}`];
+    Object.entries(inputs).forEach(([_, value]) => {
```
WesleyKapow, line 62: What type is `inputs`? Why `Object.entries` here?

WesleyKapow (dismissed):

PR #219, opened 2022-11-10
CTW-544 Changing condition history title to category

eric-gunther:
Previously we were displaying the code display value which was ending up pretty redundant. Per CTW-544 changing this to display the condition category which is either "Problem List Item" or "Encounter Diagnosis". Essentially distinguishing whether "this visit was to address this Dx" vs "we verified the patient has this condition". 

**Before**
image

**After**
image

### DIFF
--- a/.changeset/angry-dryers-boil.md
+++ b/.changeset/angry-dryers-boil.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Changing title in conditions history to category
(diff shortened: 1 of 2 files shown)


.changeset/angry-dryers-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Changing title in conditions history to category
```
pmadria3, line 5: nit: Changing -> Change

pmadria3 (approved):
small nit but otherwise LGTM

pmadria3 (approved):

PR #235, opened 2022-11-10
CTW-518/filter condition summary resources

WesleyKapow:
CTW-518

We now filter out other provider conditions if there are ANY (active or inactive) patient conditions that match and are newer. In order to do this we fetch all conditions.

NOTE: There is currently a bug here if there are more than 250 conditions as that is the max we get from ODS without paging.

### DIFF
--- a/src/components/content/conditions/helpers.test.ts
+++ b/src/components/content/conditions/helpers.test.ts
@@ -1,0 +1,71 @@
+import { ConditionModel } from "@/models";
+import { filterOtherConditions } from "./helpers";
+
+describe("Condition Helpers", () => {
+  describe("filterOtherConditions", () => {
+    it("should filter out when there's a match and other record has no date", () => {
+      const { others, patients } = setupConditions();
+      const filtered = filterOtherConditions(others, patients);
+      expect(filtered).toHaveLength(2);
+      expect(filtered[0].id).toEqual("other2");
+      expect(filtered[1].id).toEqual("other3");
+    });
+
+    it("should filter out when there's a match and patient record is newer", () => {
+      const { others, patients } = setupConditions("2022-11-09", "2022-11-10");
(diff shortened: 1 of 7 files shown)


src/components/content/conditions/helpers.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      ];
+      const patients = [
+        getCondition("patient2"),
+        patientCondition,
+        getCondition("patient3"),
+      ];
+      return { others, patients };
+    }
+  });
+});
```
luthfib, line 71: Tests!

luthfib (approved):

src/components/content/conditions/helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+  otherConditions.filter(
+    (otherCondition) =>
+      !patientConditions.some((patientCondition) => {
+        const otherRecordedDate = otherCondition.resource.recordedDate;
+        const patientRecordedDate = patientCondition.resource.recordedDate;
+
+        return (
+          otherCondition.knownCodingsMatch(patientCondition) &&
+          (!otherRecordedDate ||
+            (patientRecordedDate && otherRecordedDate <= patientRecordedDate))
```
luthfib, line 19: Why do we need to check that `patientRecordedDate is <= patientRecordedDate`

pmadria3 (approved):
LGTM

src/components/content/conditions/helpers.ts
WesleyKapow: Because we want to rule out other provider records that are older. If the other provider record is newer than the patient one, then we want to show it as it could mean new information from a third party. Fair question though if we should also hide it in that case if it has the same status o_O. I'll follow up with Michelle on that.

PR #237, opened 2022-11-14
CTW-391 Provenance

melocule:
CTW-391
Adds provenance records when data is added or edited to track changes and add accountability.

**Examples**
- [Updated Condition](https://fhirplace.dev.zusapi.com/#/Condition/8fc91a05-3db6-49d4-98ae-d3f7f6c8b26d/show/provenance)
- [Created Condition](https://fhirplace.dev.zusapi.com/#/Condition/7e4dae0f-6240-4a8b-8efa-38f5e2f5c3ba/show/provenance)


src/utils/auth.ts
```diff
@@ -12,7 +12,7 @@ export const AUTH_USER_TYPE = "https://zusapi.com/user_type";
 export const AUTH_PATIENT_ID = "https://zusapi.com/patient_id";
 export const AUTH_AUTHENTICATED_BY = "https://zusapi.com/authenticated_by";
 
-type ZusJWT = {
+export type ZusJWT = {
```
WesleyKapow, line 15: Exported but not used, leave it as not exported?

src/fhir/action-helper.ts
```diff
@@ -1,23 +1,115 @@
...
   resource: Resource,
-  fhirClient: Client
+  requestContext: CTWRequestContext
 ) {
+  const practitionerId = claimsPractitionerId(requestContext.authToken);
+  const builderName = claimsBuilderName(requestContext.authToken);
+  const authEmail = claimsAuthEmail(requestContext.authToken);
+  const { fhirClient } = requestContext;
+  const resourceModified = resource;
+  const createProvenance = async (type: "CREATE" | "UPDATE") => {
```
WesleyKapow, line 24: Move to a helper in `fhir/provenance.ts` that way we have a single place for provenance stuff and we can use it wherever.

src/fhir/action-helper.ts
```diff
@@ -1,23 +1,115 @@
...
+    let provenance: Provenance = {
+      resourceType: "Provenance",
+      agent: [
+        {
+          who: {
+            reference:
+              practitionerId === ""
+                ? undefined
+                : `Practitioner/${practitionerId}`,
+            display: authEmail,
```
WesleyKapow, line 34: Let's use name if we have it.

WesleyKapow, line 34: Let's use practitioner name if we have it.

The problem with how you have it now is you would end up pointing to a practitioner resource but the display is an email.. which may not even exist on the practitioner resource.

src/fhir/action-helper.ts
```diff
@@ -1,23 +1,115 @@
...
+        activity: {
+          coding: [
+            {
+              system: SYSTEM_PROVENANCE_ACTIVITY_TYPE,
+              code: "CREATE",
+              display: "create",
+            },
+          ],
+        },
+      };
```
WesleyKapow, line 71: re `...provenance`:

`foo = {...foo, bar: 3}`.. why not simply `foo.bar = 3`?

WesleyKapow (dismissed):

src/fhir/provenance.ts
```diff
@@ -0,0 +1,90 @@
...
+  };
+  if (type === "CREATE") {
+    provenance.activity = {
+      coding: [CREATE_CODING],
+    };
+  } else {
+    provenance.activity = {
+      coding: [UPDATE_CODING],
+    };
+  }
```
WesleyKapow, line 85: You can even inline this in the above object and use a ternary:

`activity: {coding: [type === 'CREATE' ? CREATE_CODING : UPDATE_CODING]}`

src/fhir/provenance.ts
```diff
@@ -0,0 +1,90 @@
...
+  const practitionerId = claimsPractitionerId(requestContext.authToken);
+  const builderName = claimsBuilderName(requestContext.authToken);
+
+  let practitionerDisplay: string;
+  try {
+    const practitioner = await getPractitioner(practitionerId, requestContext);
+    practitionerDisplay = practitioner.fullName;
+  } catch {
+    practitionerDisplay = claimsAuthEmail(requestContext.authToken);
+  }
```
WesleyKapow, line 47: This is good but now you've got 1 condition here and 1 condition in the `who`. You can simplify further by having this codeblock setup the `who`.

WesleyKapow (changes requested):

src/fhir/provenance.ts
melocule: Thanks for the reviews, code looks a lot cleaner!

src/fhir/provenance.ts
WesleyKapow: Bump. Use a single `who` variable instead of the two (display & reference). That way you avoid having to set `reference` to an `undefined`.

WesleyKapow: I'm literally meaning setup a `who` variable instead of having to manage both. That should be a bit clearer. This is also a nitpick for sure!

WesleyKapow: so like, you can create an `agentWho` variable and have that hold both the `reference` and `display` or just the `display`.

WesleyKapow (approved):

PR #243, opened 2022-11-14
CTW-534/medication side panel

mrosata:
- Added mocks with msw
- Created med history story
- Updated UX of med details and history for CTW-534

---
Fixes to the sidepanel
Screen Shot 2022-11-14 at 4 38 51 PM

Screen Shot 2022-11-14 at 4 39 26 PM


src/components/core/data-list.tsx
```diff
@@ -18,9 +18,9 @@ export function entryFromArray(label: string, array: string[]) {
 export function DataList({ title, data }: DataListProps) {
   return (
     <div className="ctw-space-y-4 ctw-rounded-lg ctw-bg-bg-lighter ctw-p-4">
-      <span className="ctw-text-sm ctw-uppercase ctw-text-content-light">
+      <h4 className="ctw-m-0 ctw-text-sm ctw-uppercase ctw-text-content-light">
```
WesleyKapow, line 21: Hmmm why `h4` here? Are there `h3`, `h2` and `h1` above it? I think it is risky to depend on `h4` styling as that is often changed globally in CSS. Instead switch back to `span` and explicitly set the text styles.

src/components/content/medication-history/mocks/patient.ts
```diff
@@ -0,0 +1,137 @@
...
+            {
+              system: "http://terminology.hl7.org/CodeSystem/v3-ActReason",
+              code: "HTEST",
+            },
+          ],
+          tag: [
+            {
+              system: "https://zusapi.com/accesscontrol/owner",
+              code: "builder/009204f2-5e4c-48da-99cc-6d5f6a99282d",
+              display: "Canvas Medical - Test Customer",
```
WesleyKapow, line 44: Remove customer names.

src/components/content/medication-history/mocks/patient.ts
```diff
@@ -0,0 +1,137 @@
+export default {
+  resourceType: "Bundle",
+  id: "c1db2d18-e0f2-420e-88d8-c0e32676e894",
+  meta: {
+    lastUpdated: "2022-11-14T18:57:45.140+00:00",
+  },
+  type: "searchset",
+  link: [
```
WesleyKapow, line 8: Probably don't need this section and those URLs wouldn't work.

src/components/content/medication-history/mocks/medication-statement.ts
```diff
@@ -0,0 +1,239 @@
...
+        managingOrganization: {
+          reference: "Organization/93bb6f30-ca25-4ca3-bb22-acae1356d8b5",
+        },
+      },
+      search: {
+        mode: "include",
+      },
+    },
+    {
+      fullUrl:
```
WesleyKapow, line 209: Remove fullUrls as they are invalid.

src/components/content/medication-history/mocks/medication-statement.ts
```diff
@@ -0,0 +1,239 @@
+export default {
```
WesleyKapow, line 1: Change file name to be plural to indicate it is more than one.

src/components/content/medication-history/index.tsx
```diff
@@ -12,9 +12,13 @@ const MEDICATION_HISTORY_LIMIT = 10;
 
 export type MedicationHistoryProps = {
   medication: MedicationStatementModel;
+  title?: string;
 };
 
-export function MedicationHistory({ medication }: MedicationHistoryProps) {
+export function MedicationHistory({
+  medication,
+  title = "Medication History",
```
WesleyKapow, line 20: Does this title ever change? If not, hardcode the string instead of setting up a prop that is never used.

src/components/content/medication-drawer.tsx
```diff
@@ -30,25 +31,28 @@ function getDataEntriesFromMedicationStatement(
 
 export const MedicationDrawer = ({
   medication,
+  title = "Medication Details",
```
WesleyKapow, line 34: Does this title ever change?

WesleyKapow (changes requested):
I still need to review the stories but thought I'd submit this now to unblock you in making changes.

src/components/content/medication-drawer.tsx
mrosata: Yeah I think this is the second time I've had to change it. Figured that lots of these title changes might be to suit the designs of a single consumer of the library so it can't hurt to just make it variable so future changes can be done outside the scope of this library.

I'm not attached to the change, but I think it's sane to have. I'll defer to you.

src/components/content/medication-history/medication-history.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+export default {
+  component: MedicationHistory,
+  decorators: [
+    (Story, { args }) => (
+      <CTWProvider env="dev" authToken="12345" builderId="12345">
+        <PatientProvider
+          patientID="007"
+          systemURL="https://zusapi.com/fhir/identifier/universal-id"
+        >
+          <MedicationHistory medication={medicationStatementModel} />
```
WesleyKapow, line 37: This should be something like:

```tsx
<Story args={args} />
```

The idea is that the decorator can wrap the story, so you still want to use the story in the JSX here.

src/components/content/medication-history/medication-history.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  MedicationHistoryProps,
+} from "@/components/content/medication-history";
+import { MedicationStatementModel } from "@/models";
+import { CTWProvider } from "@/components/core/ctw-provider";
+import { PatientProvider } from "@/components/core/patient-provider";
+import aggregatedFromMedicationStatement from "@/components/content/medication-history/mocks/aggregated-from-med-statement";
+import medicationStatement from "@/components/content/medication-history/mocks/medication-statement";
+import medicationRequest from "@/components/content/medication-history/mocks/medication-request";
+import medicationDispense from "@/components/content/medication-history/mocks/medication-dispense";
+import mockedUser from "@/components/content/medication-history/mocks/patient";
```
WesleyKapow, line 14: s/mockedUser/patient

Or `mockedPatient` but you don't have `mocked` prefix for the other ones.

WesleyKapow (changes requested):

src/components/content/medication-drawer.tsx
WesleyKapow: YAGNI!

We should not have props that are not used. This is a general coding best practice of not having code that is unused. It becomes confusing to the reader of the code -- "why is this here, nothing uses it?". And of course YAGNI! Our requirements down the road may change and so the un-used flexibility you add now is likely the wrong flexibility that we'll want in the future. For example, we may want customers to be able to customize ALL strings we use. In that case, we probably wouldn't use a prop here but instead use patterns used for internationalization where you lookup the string from a global map.

WesleyKapow: Also.. outsiders would still not be able to change this title a.t.m if they are using the top level `Medication` component as the prop is not plumbed all the way down.

src/components/core/data-list.tsx
mrosata: These changes are for semantics rather than style. I did have to add styles _(like `ctw-m-0`)_ in order to make the heading tags look as the span did.

The other option would be to use aria attributes I suppose
```suggestion
      <span
        className="ctw-text-sm ctw-uppercase ctw-text-content-light"
        role="heading"
        aria-level="4"
      >
```

mrosata: @WesleyKapow for reference this figma screenshot is where the levels came from

Screen Shot 2022-11-15 at 10 12 25 AM

https://www.figma.com/file/BX8WkUMyecKimerE4VjdkB/Medication-Overview?node-id=682%3A6088

src/components/content/medication-drawer.tsx
mrosata: Canvas uses the component directly instead of the top-level one. I see your point though.

src/components/core/data-list.tsx
WesleyKapow: Ah ok awesome! We are using them correctly then! Carry on!

WesleyKapow (approved):
LGTM

PR #250, opened 2022-11-16
CTW-523/storybook conditions

WesleyKapow:
CTW-523

Add condition stories and fix a few bugs:
1. Don't show multiple `required` condition errors when adding a condition.
2. Hide empty table's Add Condition button when readOnly
3. Don't modify the actual other provider condition when adding and setting default

Screen Shot 2022-11-16 at 10 49 45 AM

### DIFF
--- a/src/components/content/forms/condition-schema.tsx
+++ b/src/components/content/forms/condition-schema.tsx
@@ -133,11 +133,15 @@ export const conditionEditSchema = conditionSchema
 export const conditionAddSchema = conditionSchema
   .extend({
     condition: z.object({
-      display: z.string(),
       code: z.string({
         required_error: "Please choose a condition.",
       }),
-      system: z.string(),
+      // These are technically required but we mark them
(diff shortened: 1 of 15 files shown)


src/components/content/forms/condition-schema.tsx
```diff
@@ -133,11 +133,15 @@ export const conditionEditSchema = conditionSchema
...
       code: z.string({
         required_error: "Please choose a condition.",
       }),
-      system: z.string(),
+      // These are technically required but we mark them
+      // as optional to avoid duplicative error messages.
+      // The condition autocomplete will set us up so that
+      // all three of these values are set.
+      display: z.string().optional(),
+      system: z.string().optional(),
```
luthfib, line 144: This is good, but ideally we would actually check that these values exist in the object and then the error message would throw if any of them fail. But for now I think this is adequate

luthfib (approved):
LGTM

PR #251, opened 2022-11-16
CTW-352 Dedupe Conditions History

melocule:
CTW-352
Prevents duplicate history records from appearing, based on uniqueness of all displayed fields.

Before, entries 3 and 4 are duplicates. Entry 1 differs slightly, entry 5 differs significantly.
image

After, only one of 3 and 4 is kept and other records remain separate.
image


.changeset/sour-socks-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Only show unique condition history entries.
```
pmadria3, line 5:
```suggestion
Show unique condition history entries.
```

src/components/content/conditions-history.tsx
```diff
@@ -2,7 +2,7 @@ import { Loading } from "@/components/core/loading";
 import { getIncludedResources } from "@/fhir/bundle";
 import { useConditionHistory } from "@/fhir/conditions";
 import { ConditionModel } from "@/fhir/models/condition";
-import { capitalize, orderBy, startCase } from "lodash";
+import _, { capitalize, orderBy, startCase } from "lodash";
```
pmadria3, line 5: why are you doing underscore?

src/components/content/conditions-history.tsx
```diff
@@ -107,13 +107,18 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
         const conditionsFilteredWithoutDate =
           filterEnteredinErrorConditions.filter((c) => !c.recordedDate);
 
-        setConditionsWithDate(
-          conditionsFilteredWithDate.map((model) => setupData(model))
+        const conditionsDataWithDateDeduped = _.uniqBy(
+          conditionsFilteredWithDate.map((model) => setupData(model)),
+          (record) => record.data.map((data) => data.value?.toString()).join()
```
luthfib, line 112: I dont think this approach work if data.value returns an object, in this case 

image

Is an object

luthfib (changes requested):
Looking good, but I think I found a blocker

src/components/content/conditions-history.tsx
luthfib: You should just import uniqBy instead of doing the underscore

src/components/content/conditions-history.tsx
```diff
@@ -102,18 +102,21 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
+          (record) => JSON.stringify(record.data)
         );
 
-        setConditionsWithoutDate(
-          conditionsFilteredWithoutDate.map((model) => setupData(model))
+        const conditionsDataFilteredWithDate = conditionsDataDeduped.filter(
+          (d) => d.date
+        );
+        const conditionsDataFilteredWithoutDate = conditionsDataDeduped.filter(
+          (d) => !d.date
```
WesleyKapow, line 114: Inline these as there's no real benefit to having them split out and using a temporary variable.

WesleyKapow: Ah I see this is what existed before.. still, I don't think it is needed.

melocule: Just fixed it!

src/components/content/conditions-history.tsx
```diff
@@ -102,18 +102,21 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
-        const conditionsFilteredWithDate =
-          filterEnteredinErrorConditions.filter((c) => c.recordedDate);
-        const conditionsFilteredWithoutDate =
-          filterEnteredinErrorConditions.filter((c) => !c.recordedDate);
-
-        setConditionsWithDate(
-          conditionsFilteredWithDate.map((model) => setupData(model))
+        const conditionsDataDeduped = uniqBy(
+          filterEnteredinErrorConditions.map((model) => setupData(model)),
+          (record) => JSON.stringify(record.data)
```
WesleyKapow, line 107: Could simplify and be a bit more explicit using `uniqWith` and `isEqual` from lodash: `uniqWith(array, isEqual)`

src/components/content/conditions-history.tsx
```diff
@@ -102,18 +102,14 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
+        const conditionsDataDeduped = uniqWith(
+          filterEnteredinErrorConditions.map((model) => setupData(model)),
+          (a, b) => isEqual(a.data, b.data)
         );
 
-        setConditionsWithoutDate(
-          conditionsFilteredWithoutDate.map((model) => setupData(model))
-        );
+        setConditionsWithDate(conditionsDataDeduped.filter((d) => d.date));
+        setConditionsWithoutDate(conditionsDataDeduped.filter((d) => !d.date));
```
luthfib, line 111: Im confused, why did we change it to be `date` from `recordedDate`?

src/components/content/conditions-history.tsx
melocule: Just changed it! Didn't realize isEqual does a nested check in Typescript instead of just comparing the references, that's nice. 
Note: Had to use the comparator `(a, b) => isEqual(a.data, b.data)` to avoid comparing id's

src/components/content/conditions-history.tsx
melocule: I moved the part that calls setupData earlier in the code, so `d` is actually the return from that helper and not a Condition. `d.date = condition.date` though, so it's the same value

src/components/content/conditions-history.tsx
```diff
@@ -102,18 +102,14 @@ export function ConditionHistory({ condition }: { condition: ConditionModel }) {
...
-        const conditionsFilteredWithDate =
-          filterEnteredinErrorConditions.filter((c) => c.recordedDate);
-        const conditionsFilteredWithoutDate =
-          filterEnteredinErrorConditions.filter((c) => !c.recordedDate);
-
-        setConditionsWithDate(
-          conditionsFilteredWithDate.map((model) => setupData(model))
+        const conditionsDataDeduped = uniqWith(
+          filterEnteredinErrorConditions.map((model) => setupData(model)),
+          (a, b) => isEqual(a.data, b.data)
```
WesleyKapow, line 107: Ah I was hoping it'd be even cleaner of just passing in `isEqual` but yes you need to narrow it down to `data`. This looks good though and I think is still cleaner and more explicit.

WesleyKapow (approved):
LGTM

PR #253, opened 2022-11-16
More condition stories

WesleyKapow:
Story changes:
* Add empty conditions story.
* Fix bug where react query was caching mock responses across stories.
* Add a fixed-width prop option for className that will trigger stacked view.
* Setup default values for className and readOnly, this makes the toggles in the storybook actions UI immediately available.

Style fixes -- opted to copy over some styles into conditions.scsss that were previously refactored into main. The goal here is to have isolation within our components by using properly namespaced classes.
* Fixed centering bug for empty stacked tables.
* Fixed pagination whitespace when stacked.
* Fixed bug where bottom border was missing when stacked and showing pagination.

# Screenshots
**New empty story and property selections**
Screen Shot 2022-11-16 at 2 27 34 PM

**BEFORE** (centering bug)
Screen Shot 2022-11-16 at 2 25 25 PM

**AFTER**
Screen Shot 2022-11-16 at 2 36 12 PM

**BEFORE** (missing bottom border & bad pagination whitespace)
Screen Shot 2022-11-16 at 2 24 39 PM

**AFTER**
Screen Shot 2022-11-16 at 2 24 08 PM

### DIFF
--- a/src/components/content/conditions.scss
+++ b/src/components/content/conditions.scss
@@ -1,8 +1,28 @@
 .ctw-conditions {
   @apply ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white;
+
+  .ctw-conditions-body-container {
+    @apply ctw-m-3 ctw-space-y-5 ctw-bg-bg-white;
+  }
+
+  .ctw-conditions-heading-container {
+    @apply ctw-flex ctw-h-11 ctw-items-center ctw-justify-between ctw-bg-bg-light ctw-p-3;
+  }
+
+  .ctw-conditions-title-container {
+    @apply ctw-flex ctw-items-center ctw-justify-between;
+  }
 }
 
-.ctw-conditions.ctw-stacked {
+.ctw-conditions.ctw-conditions-stacked {
+  .ctw-conditions-body-container {
+    @apply ctw-m-0;
+  }
+
+  .ctw-conditions-title-container {
+    @apply ctw-m-3 ctw-mb-0;
+  }
+
   // When there's zero records, we won't
   // show pagination and so we'll get a double border
   // unless we bump down by 1px.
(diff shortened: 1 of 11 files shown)


luthfib (approved):

luthfib:
You have no changeset, although if you don't want necessarily release yet and wait till something else goes out im fine with leaving this without a changeset.

pmadria3:
LGTM - except missing changeset potentially

src/components/content/conditions.scss
```diff
@@ -1,8 +1,28 @@
...
-.ctw-conditions.ctw-stacked {
+.ctw-conditions.ctw-conditions-stacked {
+  .ctw-conditions-body-container {
+    @apply ctw-m-0;
+  }
+
+  .ctw-conditions-title-container {
+    @apply ctw-m-3 ctw-mb-0;
+  }
+
```
mrosata, line 25: What is the driving principle behind creating completely new classnames (ie: `ctw-conditions-stacked`, `ctw-conditions-body-container`) instead of something like: 
```scss
.ctw-conditions.ctw-stacked {
  .ctw-body-container {
    // ..overrides & additions
  }

  .ctw-title-container {
    // ..overrides & additions
  }
}
```
I'm not saying it's wrong, just curious of the choice

WesleyKapow: Name-spacing / isolation. I don't want to reason about the other uses of `.ctw-body-container` and the existing styles. I want to ONLY focus on conditions.

src/components/content/conditions/conditions.stories.tsx
```diff
@@ -19,6 +20,20 @@ type Props = ConditionsProps;
...
+      options: ["Blank", "Fixed Width"],
+      control: "select",
+      mapping: {
+        Blank: "",
+        "Fixed Width": "ctw-m-auto ctw-max-w-[600px]",
+      },
+    },
+  },
+  args: {
+    className: "Empty",
```
mrosata, line 34: guessing at this, should "Empty" map to a key of `argTypes.className.mapping`
```suggestion
    className: "Blank",
```
if not just ignore

WesleyKapow: No that's correct, fixed!

src/components/content/conditions.scss
WesleyKapow: In general, I think we want very few global classes that are very basic, like `ctw-btn` and `ctw-link`. Anything more complex that deals with multiple elements should be abstracted into a component. We can also respect the wisdom of "a little duplication is better than the wrong abstraction" and duplicate code when the abstraction/common component isn't clear.

mrosata (approved):

PR #256, opened 2022-11-16
CTW-546/Prescriber sometimes blank

mrosata:
https://zeushealth.atlassian.net/browse/CTW-546

When MedicationDrawer shows details of a medication it usually has a blank prescriber. This is because the details come from a MedicationStatement which only knows about a prescriber if the `"https://zusapi.com/lens/extension/medicationLastPrescriber"` lens is set.

It is usually (if not always) the case that this lens is not set and so the field is blank in the details ui. Now the component will check if that field is blank again after fetching related medication* resources from ODS. It will then find the latest MedicationRequest (and included requester resource) and use those to fill in the prescriber field.

### DIFF
--- a/src/components/content/medication-drawer.tsx
+++ b/src/components/content/medication-drawer.tsx
@@ -79,7 +79,7 @@ export const MedicationDrawer = ({
             </h3>
           </div>
           <DataList title="Summary" data={data} />
-          {medication.rxNorm && <MedicationHistory medication={medication} />}
+          <MedicationHistory medication={medication} />
         </div>
       </Drawer.Body>
       <Drawer.Footer>
--- a/src/components/content/medication-drawer.tsx
+++ b/src/components/content/medication-drawer.tsx
@@ -37,10 +37,10 @@ export const MedicationDrawer = ({
   medication,
   ...drawerProps
 }: MedicationDrawerProps) => {
-  const [lastPrescriber, setLastPrescriber] = useState<string | null>();
+  const [lastPrescriber, setLastPrescriber] = useState<string | undefined>();
   const data = getDataEntriesFromMedicationStatement(
     medication,
-    lastPrescriber || ""
+    lastPrescriber
   );
   const historyQuery = useMedicationHistory(medication?.resource);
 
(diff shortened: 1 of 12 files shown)


src/components/content/medication-drawer.tsx
```diff
@@ -1,49 +1,85 @@
-import type { MedicationStatementModel } from "@/fhir/models/medication-statement";
+import { MedicationStatementModel } from "@/fhir/models/medication-statement";
```
WesleyKapow, line 1: Hmmm eslint or TS use to complain about importing types without `type`... what happened?

luthfib (dismissed):

src/components/content/medication-drawer.tsx
```diff
@@ -1,49 +1,85 @@
...
+    ...entryFromArray("Note", medication.notesDisplay),
+  ];
 }
 
 export const MedicationDrawer = ({
   medication,
   ...drawerProps
 }: MedicationDrawerProps) => {
-  const data = getDataEntriesFromMedicationStatement(medication);
+  const [data, setData] = useState(
```
WesleyKapow, line 37: Can all of the changes here be refactored and delegated to some combo of:
1. `getDataEntriesFromMedicationStatement`
2. The models.

src/fhir/models/medication-statement.ts
```diff
@@ -165,13 +166,12 @@ export class MedicationStatementModel extends FHIRModel<fhir4.MedicationStatemen
...
-    return formatDateISOToLocal(
-      this.resource.extension?.find(
-        (x) => x.url === LENS_EXTENSION_MEDICATION_LAST_PRESCRIBED_DATE
-      )?.valueReference?.display
-    );
+    const value = this.resource.extension?.find(
+      (x) => x.url === LENS_EXTENSION_MEDICATION_LAST_PRESCRIBER
+    )?.valueString;
+
+    return value && value !== "NO-VALUE" ? value : undefined;
```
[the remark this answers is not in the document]
mrosata: yeah they do, I guess it's because they can't put an empty string. We can push back if you'd like.

src/components/content/medication-drawer.tsx
mrosata: yeah this was a ts complaint that I'd started using the concrete value, this was one of the earlier commits though and  refactors have made it where `import type` would again be acceptable.

src/components/content/medication-drawer.tsx
mrosata: it uses `setLastPrescriber` now and passes that prescriber into the `getDataEntriesFromMedicationStatement` so the useEffect doesn't need to `setData`. Code looks cleaner now, it feels weird passing the lastPrescriber into `getDataEntriesFromMedicationStatement`, but it does make everything a bit clearer.

src/fhir/models/medication.ts
```diff
@@ -63,4 +67,52 @@ export class MedicationModel extends FHIRModel<Medication> {
...
   }
+
+  /**
+   * This accessor will try to get the prescriber for the underlying medication
+   * models resource. Depending on the type of fhir resource, it will delegate
+   * the work to a more specific fhir/model/*.ts class before simply grabbing
+   * the `display` property from an actor/performer/requester. If all else
+   * should fail, the accessor returns an empty string.
+   */
+  get prescriber(): string {
```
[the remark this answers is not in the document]
mrosata: meaning have
```
prescriber = new MedicationDispenseModel(
          this.resource,
          this.includedResources
        ).includedPerformer;
```
fallback to `this.resource.performer?.[0]?.actor.display` from inside the accessor `MedicationDispenseModel.includedPerformer`?

src/fhir/medications.ts
```diff
@@ -180,9 +181,15 @@ export function useMedicationHistory(medication: fhir4.MedicationStatement) {
 
   return useQueryWithPatient(
     QUERY_KEY_MEDICATION_HISTORY,
-    [medication.id],
+    [medication?.id || "empty"],
     async (requestContext, patient) => {
       try {
+        if (!medication) {
```
WesleyKapow, line 187: Can pull this out of the `try`.

mrosata: Could you leave it as a suggestion please.

src/fhir/models/medication.ts
```diff
@@ -75,44 +74,23 @@ export class MedicationModel extends FHIRModel<Medication> {
    * the `display` property from an actor/performer/requester. If all else
    * should fail, the accessor returns an empty string.
    */
-  get prescriber(): string {
-    let prescriber: fhir4.Resource | undefined;
-    let fallback: string | undefined;
-
+  get prescriber(): string | undefined {
```
WesleyKapow, line 77: Can probably clean up the comment?

WesleyKapow (approved):
One minor comment otherwise LGTM.

Really like the `useLastPrescriber` hook. Helps keep the drawer from getting too wild and nicely encapsulates everything needed for fetching the last prescriber.

PR #257, opened 2022-11-17
Ctw 472/patient history request drawer

luthfib:
Adds the following: 
- [x] Patient Request Drawer
- [x] Allows user to pre-populate their patient data from the zus patient api
- [x] Modify patient data on save
- [x] Hit the patient history api once the data save is confirmed 
- [x] Handle patient history api errors

image


src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -51,7 +62,27 @@ export const DrawerFormWithFields = <T,>({
...
+                  return (
+                    <FormField
+                      key={label}
+                      readonly={readonly}
+                      render={render}
+                    />
+                  );
+                }
+
+                // TODO: figure out why typescript thinks this can be undefined
```
WesleyKapow, line 85: If `errors` can be `undefined` then `errors?.[field]` can be `undefined`.

src/components/content/condition-header.tsx
```diff
@@ -12,3 +13,9 @@ export const ConditionHeader = ({ condition }: ConditionHeaderProp) => (
     <div className="ctw-text-sm">{condition.ccsGrouping}</div>
   </div>
 );
+
+export const ConditionHeaderSummary = ({
```
WesleyKapow, line 17: This is pretty trivial and used in one place. I vote inline it and remove this extra component.

src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -1,21 +1,32 @@
...
+  | {
+      presentational: boolean;
+      label: string;
+      field?: never;
+      value?: never;
+      lines?: never;
+      readonly?: never;
+      hidden?: never;
+      render: () => JSX.Element;
+    };
```
WesleyKapow, line 29: See if you can get by without using `presentational`. You can still use this union type and `nevers` to specify the two types of uses. You can also pull these subtypes out to give you something like `export type FormType = FormFieldType | FormContentType` (where those subtypes are defined just above and are also exported).

src/components/content/patient-history-request-drawer.tsx
```diff
@@ -0,0 +1,61 @@
...
+  getRequestData,
+  requestHistorySchema,
+} from "./forms/request-history-schema";
+
+type PatientHistoryRequestDrawer<T> = Pick<
+  DrawerFormWithFieldsProps<T>,
+  "patientID" | "isOpen" | "action" | "onClose" | "header"
+>;
+
+export class RequestData {
```
WesleyKapow, line 15: What's this? How about a simple object `{name: "", ...}`?

WesleyKapow (changes requested):

src/components/content/patient-history-request-drawer.tsx
luthfib: This will get removed, it was simply to mock an object

src/components/core/form/drawer-form-with-fields.tsx
luthfib: Yup, but theoretically if check for presentational being false then errors should be defined, which indicates im typing incorrectly somewhere

src/components/core/form/drawer-form-with-fields.tsx
luthfib: You mean without adding a new prop of presentational?

WesleyKapow: Yes. I don't think it is needed.

luthfib: specifically if we do..

```
 if (presentational) {
                  return (
                    <FormField
                      key={label}
                      readonly={readonly}
                      render={render}
                    />
                  );
                }
```

WesleyKapow: `if (field)`

WesleyKapow: `presentational` is redundant.. you know if it is presentational based on the other props.

src/fhir/models/patient.ts
```diff
@@ -1,6 +1,7 @@
 import { findReference } from "@/fhir/resource-helper";
```
WesleyKapow, line 1: Models shouldn't have any setters. This was a failed pattern from CTW. Instead, any mutations should happen directly on a `fhir4.resource`, and it is that resource that should be saved. No model should be involved in mutations!

src/components/content/forms/patients.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  if (!result.success) {
+    return { formResult: result, requestErrors: undefined };
+  }
+
+  const requestContext = await getRequestContext();
+  const managingOrganization = patient.organization?.id;
+  if (!managingOrganization) {
+    throw Error(
+      `Managing organization not found with id of ${patient.organization?.id} and is needed in order to perform patient scheduling.`
```
WesleyKapow, line 33: Why do I care about patient scheduling here? This is `editPatient` not some setup scheduling helper.

WesleyKapow: Ah you do it lower! Refactor this! editing a patient and fetching history are two separate actions. Our drawer may do both, but they are separate actions.

src/components/content/forms/patients.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  if (isFhirError(response) && isOperationOutcome(response.response.data)) {
+    requestErrors = new OperationOutcomeModel(response.response.data).issues
+      .filter((issue) => issue.severity !== "warning")
+      .map((issue) => issue.display);
+    result.success = false;
+  } else if (response instanceof Error) {
+    requestErrors = [response.message];
+    result.success = false;
+  }
```
WesleyKapow, line 78: This is gross. Can you refactor this out? Maybe into `createOrEditFhirResource`?

WesleyKapow (changes requested):

src/components/content/patient-history-request-drawer.tsx
```diff
@@ -9,53 +10,24 @@ import {
...
-      data={getRequestData(new RequestData())}
-      schema={requestHistorySchema}
-      isOpen={isOpen}
-      onClose={() => onClose(false)}
-    />
-  );
-};
+}: PatientHistoryRequestDrawer<T>) => (
+  <DrawerFormWithFields
+    patientID={patient.id}
```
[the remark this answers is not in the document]
luthfib: This has been in here for a while. I'll see about refactoring this, however I am pretty sure at this point medications, and conditions all use this prop.

src/components/content/forms/patients.ts
luthfib: I don't think it makes sense to put it into `createOrEditFhirResource` because it edits the result object, not sure thats something we want to introduce into `createOrEditFhirResource `

src/components/content/forms/patients.ts
```diff
@@ -61,20 +54,43 @@ export const editPatient = async (
...
+
+  const result = editPatientResponse.formResult;
+  let { requestErrors } = editPatientResponse;
+
+  const requestContext = await getRequestContext();
+  if (!patient.organization?.id) {
+    throw Error(
+      `Managing organization not found with id of ${patient.organization?.id} and is needed in order to perform patient scheduling.`
+    );
   }
```
WesleyKapow, line 94: Why is this needed? I don't see us passing any of this to the `schedulePatientHistory` function. Can we instead rely on the request returning an error?

src/components/content/forms/patients.ts
```diff
@@ -61,20 +54,43 @@ export const editPatient = async (
...
-      .filter((issue) => issue.severity !== "warning")
-      .map((issue) => issue.display);
-    result.success = false;
-  } else if (response instanceof Error) {
-    requestErrors = [response.message];
-    result.success = false;
+  return { formResult: result, requestErrors };
+};
+
+export const editPatientAndScheduleHistory = async (
```
WesleyKapow, line 71: Ideally this would just be `scheduleHistory`. Having smaller helpers is more flexible. You can then define `editPatientAndScheduleHistory` where you use it and have it call the two helpers.

WesleyKapow (changes requested):

src/components/content/forms/patients.ts
luthfib: It already combined two helpers, `schedulePatientHistory` and `editPatient`. Not sure why a even more generic version would look like?

WesleyKapow (approved):

PR #260, opened 2022-11-18
CTW-336 User sort conditions

melocule:
CTW-336
Conditions are sortable by name and group in ascending and descending order by the user.
- By default, conditions are sorted by group in ascending order
- Within the same group, conditions are always sorted alphabetically in ascending order
- Records with a blank in the field being sorted with always show last

https://user-images.githubusercontent.com/33408946/203149787-b96e78d2-af7f-4a48-bc79-856bc490e4ef.mov


luthfib (changes requested):
Seems you have some changes you havent pulled into your branch. Somethings funky going, fix that and request review because it is a bit hard to tell which files you changed. (Let me know if you need help fixing the conflicts)

src/utils/sort.ts
```diff
@@ -22,9 +23,11 @@ export function getSortInfo<T>(model: Newable<T>, sortValue?: string | null) {
     return {};
   }
 
+  type NewType = SortDir;
+
   return {
     sortColumn: sortColumn as keyof T,
-    sortOrder: sortOrder as "asc" | "desc",
+    sortOrder: sortOrder as NewType,
```
luthfib, line 30: Do we need `newType` here? I think it is you can just do `sortOrder as SortDir`

src/components/content/conditions-table-base.tsx
```diff
@@ -24,12 +26,16 @@ export function ConditionsTableBase({
...
+      sortFn: (a, b, dir) => alphaSortBlankLast(a.display, b.display, dir),
     },
     {
       title: "Category",
       dataIndex: "ccsGrouping",
       widthPercent: 25,
       minWidth: 192,
+      sortFn: (a, b, dir) =>
+        alphaSortBlankLast(a.ccsGrouping, b.ccsGrouping, dir) +
+        0.1 * alphaSortBlankLast(a.display, b.display, "asc"),
```
luthfib, line 38: Can this be abstracted to where we dont have to pass in the sort function everytime but instead we just specify that we want the column to be sortable?

luthfib (changes requested):
The Chevron should only show up when the column is not actively being sorted or being hover. Having the Chevron show up, even dimmed looks a bit odd.

melocule:
Todo: Sync the sort for the two tables

src/components/content/conditions-table-base.tsx
melocule: Hey! I think if I replace `sortFn` with just the `sortProperty` we'd lose some control over how each column is sorted-- e.g. right now, the lambda for sorting by category secondarily sorts by condition name in ascending order. Would there be a workaround for that?

melocule: Just pushed up the new version where, for a column to be sortable, either a comparator function or the index to sort by should be provided

luthfib: There seems to be a couple ts issues

luthfib (approved):

PR #263, opened 2022-11-22
CTW-481 Updating meds component to use active meds lens

eric-gunther:
We are currently displaying "summary medications" which include meds from long ago that the users do not care about and pollutes the UI. We are now going to display "active medications".

Also did some light renaming to help clarify a distinction between "summary" resources and "lens" resources.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@
 VITE_ENV=dev
 VITE_BUILDER_ID=d037b2ef-66d6-4fcc-b228-2bc748ce4d0e
 VITE_SYSTEM_URL=https://zusapi.com/fhir/identifier/universal-id
-VITE_PATIENT_ID=2f0d1b24-b012-403f-b24c-8bf741484783
+VITE_UPID=2f0d1b24-b012-403f-b24c-8bf741484783
 
 # Copy your auth token from postman.
 VITE_AUTH_TOKEN=ACCESS_TOKEN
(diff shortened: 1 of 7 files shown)


.env.example
```diff
@@ -2,7 +2,7 @@
 VITE_ENV=dev
 VITE_BUILDER_ID=d037b2ef-66d6-4fcc-b228-2bc748ce4d0e
 VITE_SYSTEM_URL=https://zusapi.com/fhir/identifier/universal-id
-VITE_PATIENT_ID=2f0d1b24-b012-403f-b24c-8bf741484783
+VITE_UPID=2f0d1b24-b012-403f-b24c-8bf741484783
```
eric-gunther, line 5: I was constantly confusing this

src/fhir/search-helpers.ts
```diff
@@ -86,8 +86,27 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
...
+    `${SYSTEM_ZUS_OWNER}|builder/${requestContext.builderId}`,
+  ];
+  const params = mergeParams(searchParams, {
+    _tag: tagFilter,
+  });
+  return searchAllRecords(resourceType, requestContext, params);
+}
+
+// Like searchAllRecords, but filters down to only summary records.
+export async function searchSummaryRecords<T extends ResourceTypeString>(
```
eric-gunther, line 109: This diff is a little confusing. I renamed `searchLensRecords` to `searchSummaryRecords` and then updated conditions to call that function. Then I added a function `searchLensRecords` which specifically searches using the lens tag that is provided to the function.

mrosata (approved):

PR #265, opened 2022-11-22
Add version to Provenance target reference

melocule:
CTW-558

Provenance now saves which historical version of a resource resulted from an operation, for more thorough tracking of changes. It saves this information as part of the target resource reference.

[Example of a Provenance record this produces](https://fhirplace.dev.zusapi.com/#/Condition/e823c9e5-806d-4fb6-864e-c3ab5e70aa86/show/provenance)

### DIFF
--- a/src/fhir/action-helper.ts
+++ b/src/fhir/action-helper.ts
@@ -29,7 +29,6 @@ export async function createOrEditFhirResource(
     });
     if (!isFhirError(response)) {
       resourceModified.id = response.id;
-      console.log(createProvenance("CREATE", resourceModified, requestContext));
     }
     return response;
   } catch (err) {
--- a/src/fhir/action-helper.ts
+++ b/src/fhir/action-helper.ts
@@ -19,7 +19,7 @@ export async function createOrEditFhirResource(
         body: omitEmptyArrays(resource),
       });
       if (!isFhirError(response)) {
-        createProvenance("UPDATE", resource, requestContext);
+        createProvenance("UPDATE", response, requestContext);
       }
       return response;
     }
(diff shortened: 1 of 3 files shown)


src/fhir/action-helper.ts
```diff
@@ -29,7 +29,7 @@ export async function createOrEditFhirResource(
     });
     if (!isFhirError(response)) {
       resourceModified.id = response.id;
-      createProvenance("CREATE", resourceModified, requestContext);
+      console.log(createProvenance("CREATE", resourceModified, requestContext));
```
pmadria3, line 32: remove

pmadria3 (approved):
LGTM

src/fhir/provenance.ts
```diff
@@ -43,27 +43,28 @@ export const createProvenance = async (
...
-  const practitionerId = claimsPractitionerId(requestContext.authToken);
   const builderName = claimsBuilderName(requestContext.authToken);
+  const versionId = parseInt(resource.meta?.versionId || "0", 10) + 1;
 
-  let practitionerDisplay: string;
-  let practitionerReference: string | undefined;
-  try {
+  const practitionerId = claimsPractitionerId(requestContext.authToken);
+  let practitionerReference: Reference = {};
+  if (practitionerId !== "") {
```
luthfib, line 51: Couldn't you do 

```suggestion
  if (!practitionerId) {
```

luthfib (approved):
Looks LGTM. I left one nitpick change

src/fhir/provenance.ts
melocule: Hey! I tried this with Priyank yesterday and it didn't work 😭 I think it's because an empty string is different from an undefined

PR #268, opened 2022-11-22
Ctw 547 subtitles for medication filled

mrosata:
sometimes it will only have "x days supplied"
Screen Shot 2022-11-22 at 2 59 11 PM

but wherever possible, it will have units and days supplied
Screen Shot 2022-11-22 at 3 00 15 PM

Where the quantity is
```
{ value: 10, unit: undefined }
```
we would show "10 units" because we can't say "tablets", "pills", ect.

### DIFF
--- a/src/fhir/models/medication-dispense.ts
+++ b/src/fhir/models/medication-dispense.ts
@@ -40,6 +40,9 @@ export class MedicationDispenseModel extends FHIRModel<fhir4.MedicationDispense>
 
   get quantityDisplay() {
     const { value, unit } = this.resource.quantity || {};
+    if (value && !unit) {
+      return `${value} units`;
+    }
     return compact([value, unit]).join(" ");
   }
 
--- a/src/components/content/medication-history/index.tsx
+++ b/src/components/content/medication-history/index.tsx
@@ -149,9 +149,9 @@ function createMedicationDispenseCard(medication: MedicationModel) {
     hideEmpty: false,
     id: medication.id,
     title: "Medication Filled",
-    subTitle: compact([
+    subtitle: compact([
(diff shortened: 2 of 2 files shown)


eric-gunther (dismissed):

src/fhir/models/medication-dispense.ts
```diff
@@ -38,15 +37,18 @@ export class MedicationDispenseModel extends FHIRModel<fhir4.MedicationDispense>
...
+    return `${value} ${unit}`;
   }
 
-  get supplied(): string {
-    const { value, unit = "days" } = this.resource.daysSupply || {};
-    if (!value) {
-      return "";
+  get supplied(): string | undefined {
+    const { value = "", unit = "days" } = this.resource.daysSupply || {};
+    if (value === "") {
```
WesleyKapow, line 50: Probably don't need to the default value for `value` and can go back to `if (!value)`

src/fhir/models/medication-dispense.ts
```diff
@@ -38,15 +37,18 @@ export class MedicationDispenseModel extends FHIRModel<fhir4.MedicationDispense>
     return this.resource.status;
   }
 
-  get quantityDisplay() {
-    const { value, unit } = this.resource.quantity || {};
-    return compact([value, unit]).join(" ");
+  get quantityDisplay(): string | undefined {
+    const { value = "", unit = "units" } = this.resource.quantity || {};
+    if (value === "") {
```
WesleyKapow, line 42: Same thing here, don't need default `value = ""` as you can instead use `undefined`.

WesleyKapow (approved):
Minor nitpicks otherwise LGTM

PR #269, opened 2022-11-22
CTW-532 Merge builder-owned and lens medications

eric-gunther:
This PR modifies our logic that splits out medications into "confirmed medications" (ie. those you as a builder know about) and "other provider records" (ie. those you don't know about). 

If there is any lens latency then a builder may have records that are not yet captured in our lens resources. This ensures we still display those.

The display value for a builder's medication may be different than the display value for a lens medication. When that happens we'll prefer the builder's display value - for Canvas this should ensure that the Zus component display values match what's in the Canvas sidebar.

https://www.loom.com/share/d28140dde76c4e468b69a51ebfef7b24


src/App.tsx
```diff
@@ -46,18 +46,6 @@ const DemoApp = ({ accessToken = "" }) => (
...
-          <h3>
-            Patient Meds <small>(no add button or confirmed meds)</small>
-          </h3>
-          <ErrorBoundary>
-            <PatientMedications
-              showConfirmedMedsTable={false}
-              readOnly={false}
-            />
-          </ErrorBoundary>
-        </div>
```
eric-gunther, line 60: The network traffic was a little confusing to debug with 2 copies of the meds component on the sample app page.

src/fhir/medications.ts
```diff
@@ -146,31 +146,56 @@ export function filterMedicationsWithNoRxNorms(
   );
 }
 
-// Splits summarized medications into those that the builder already knows about ("Provider Medications")
+// Splits medications into those that the builder already knows about ("Provider Medications")
 // and those that they do not know about ("Other Provider Medications").
-export function splitSummarizedMedications(
-  summarizedMedications: MedicationStatement[],
+export function splitMedications(
```
eric-gunther, line 151: This function's logic could probably be made more concise, but this is somewhat complicated logic so I was trying to make it more explicit.

.changeset/fair-adults-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Updated logic for displaying builder owned medications.
```
WesleyKapow, line 5: nit: What was changed from a user perspective?

src/App.tsx
WesleyKapow: Storybook FTW!

src/fhir/medications.test.ts
```diff
@@ -65,6 +79,14 @@ describe("splitSummarizedMedications", () => {
       "0.medicationCodeableConcept.coding.0.code",
       "known"
     );
+    expect(builderMedications).toHaveProperty(
```
WesleyKapow, line 82: TIL `toHaveProperty`, that's convenient!

src/fhir/medications.ts
```diff
@@ -146,31 +146,56 @@ export function filterMedicationsWithNoRxNorms(
...
+  const builderMedications = builderOwnedMedications.filter(
+    (builderMed) =>
+      !activeMedications.some(
+        (activeMed) =>
+          getIdentifyingRxNormCode(activeMed, includedResources) ===
+          getIdentifyingRxNormCode(builderMed, includedResources)
+      )
+  );
+
+  // Get builder owned medications that line up with an active medication.
```
WesleyKapow, line 176: So the lens record has a bunch of data you want to display in the builder table? What does it have? Why can't the builder record have it? Is it things like request/dispenses? Should we instead fetch those and link to the actual resources?

src/fhir/medications.ts
```diff
@@ -146,31 +146,56 @@ export function filterMedicationsWithNoRxNorms(
...
+  // Take the active medication but modify it to use the builder's display properties.
+  builderOwnedMedications.forEach((builderMed) => {
+    const mergedMed = clone(
```
WesleyKapow, line 179: Probably better to use `cloneDeep`, otherwise you get a shallow clone where modifying any of the copied objects would modify the original.

WesleyKapow (changes requested):

src/fhir/medications.ts
eric-gunther: Figured that cloneDeep was overkill since I'm not modifying any of the nested objects, but that probably makes this unsafe/vulnerable to future changes.

src/fhir/medications.ts
eric-gunther: The lens record has last prescribed date, last prescriber, last dispensed date, (all of which we display on the table) and a few other properties that the lens process calculates and then writes to the lens med statement as an extension. There's been a bunch of conversation as to whether that's an abuse of extensions and perhaps this data is better expressed as just a reference to med requests/dispenses. Evan has some more opinions/background on this, but from my perspective it's about providing a convenient way of accessing a bunch of data about a "medication summary". The builder record doesn't have any of this stuff, it's just some medication statement belonging to the builder, without any additional context.

One reason for this change is to display the builder's display text for the med (in case it is different than the lens display text).

Perhaps more importantly, we don't want this part of the component to be too sensitive to lens latency, especially if users expect data to show up immediately after writing it. So if we're stuck waiting a few seconds for lens to process then there may be a lag in the last prescribed/last dispensed information showing up, but there won't be a lag in the row showing up.

WesleyKapow: Right I think I get what you're doing but it seems brittle. You are taking a lens resource and treating it as a builder resource by replacing a few things. What if we want to replace more? What if the lens has stuff that's "wrong" w.r.t a builder resource.

I think ideally we'd either:
1. Fetch the associated resources we need to display the data we want to display.
2. Create a new FHIR resource type to contain all of the "medication" bits we want in a single resource.

I'm fine with this change as is but definitely encourage you and lens folk think through this and decide on what solution you want going forward.

WesleyKapow (dismissed):

eric-gunther:
Latest testing round:

https://user-images.githubusercontent.com/97455910/205123485-7130a50b-c9d7-450a-8ed8-4d3c5077ca56.mov

mrosata (approved):

PR #272, opened 2022-11-23
Ctw 538/fix entered error now show up in lens

luthfib:
Conditions with entered-in-error in patient record will no longer filter out other provider conditions. 

_Leaving in draft until I get a repro of the bug. I believe the current fix work and have tests to prove it but still wanted to verify against the repro_


WesleyKapow (approved):
Beautiful!

PR #273, opened 2022-11-23
Ctw 524/storybook tests

WesleyKapow:
Added a bunch of tests to conditions. One tricky aspect here is properly mocking mutations. To this end, I've setup a `requests.ts` file that both handles setting up and resetting a cache AND providing the mocks that will mutate this cache.

Additionally:
* Customized story docs page to not include stories.
* Add `for` and `id` to labels and drawer form inputs to aid testing (can now target inputs by label).
* CollapsibleDataList now labels the expand/collapse button with `details`, making it easier to target in tests.
* DropdownMenu labels the trigger with `dropdown` making it easier to target in tests.
* DropdownMenu no longer closes menu when focusing on an outside element. This fixes an issue with `userEvent.click` in tests. See https://github.com/testing-library/user-event/issues/1080
* Auth returns empty object when jwt decode fails. This fixes issue with tests.
* Fixed bug in medications where we weren't setting up the `Promise.all` correctly.
* Added ESLint rule to guard against unhandled promises (must `await` or `void` any returned promises).

Following tests are hidden when deployed on github (TODO: verify when published):
Screen Shot 2022-11-23 at 1 46 49 PM


luthfib (approved):
HUGE effort. Going to make all of our lives easier. I like specially how have an object conditionTable to handle the different tests, and make the tests easier to follow

PR #282, opened 2022-11-28
Provenance reference correct version

luthfib:
Do not increment the version number by 1 in provenance and instead use the version id of the response.


melocule (approved):
Tested this together

PR #284, opened 2022-11-29
eslint imports with default sorting of groups

mrosata:
Explicitly setting the `import/order` configuration to the [default sort order](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md).

> The only allowed strings are: "builtin", "external", "internal", "unknown", "parent", "sibling", "index", "object", "type". The enforced order is the same as the order of each element in a group. Omitted types are implicitly grouped together as the last element.

This means since we are setting `"builtin", "external", "parent", "sibling", "index"`, groups like "object"/"type" don't have a specific ordering besides needing to come after the builtin, external, parent, sibling, index groups. We might think about making this more explicit if we run into issues where `eslint --fix` prints import-order errors but doesn't actually have enough context to fix the ordering for us.

It's also possible to enforce empty lines between groups and sort order within groups. Personally I'm not a fan of the empty lines between groups. I wouldn't mind setting sort order within groups though if we run into sort issues down the road.

This should be good to merge though. Maybe someone wants to test it out in VSCode first though @WesleyKapow

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -34,7 +34,25 @@ module.exports = {
     "import/order": [
       "error",
       {
-        groups: ["builtin", "external", "parent", "sibling", "index"],
+        alphabetize: {
+          caseInsensitive: true,
+          order: "asc",
+        },
+        groups: [
+          ["builtin", "external", "object", "type"],
+          ["internal", "parent", "sibling", "index"],
+        ],
+        "newlines-between": "never",
+      },
+    ],
+    "sort-imports": [
+      "error",
+      {
+        allowSeparatedGroups: true,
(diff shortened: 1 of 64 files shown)


melocule (dismissed):
Fix conflict with Prettier import rules-- rn the automatic formatting by Prettier undoes the import ordering of ESLint

mrosata:
> Fix conflict with Prettier import rules-- rn the automatic formatting by Prettier undoes the import ordering of ESLint

For documentation purposes. We tracked this down to the VSCode setting `"organizeImports"` being on.

WesleyKapow (approved):

PR #285, opened 2022-11-29
Fix managing organization patient bug

luthfib:
Fix bug where managing org was added when it was not present.


WesleyKapow (approved):

WesleyKapow:
We may also want to consider switching to using patch to only update the fields we are modifying.

pmadria3 (approved):
LGTM

luthfib:
> We may also want to consider switching to using patch to only update the fields we are modifying.

I agree. I'll look at doing adding this for patient.

PR #288, opened 2022-11-29
CTW-336 user sort conditions redo

melocule:
CTW-336
Redo condition sorting to allow all columns to be sortable, and sort primarily by last recorded date and condition name.

Based on some updated AC:
- Conditions Tables should remain sorted by descending recorded date, followed by condition name
- Instead of only two sortable columns, make all sortable
- Flip the sort arrow (up chevron means ascending order, down chevron means descending order)

### DIFF
--- a/.changeset/small-gifts-refuse.md
+++ b/.changeset/small-gifts-refuse.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Redo condition sorting to allow all columns to be sortable, and sort primarily by last recorded date and condition name.
+Condition sorting now allows all columns to be sortable. Sort primarily defaults to be by last recorded date and condition name.
--- a/.changeset/small-gifts-refuse.md
+++ b/.changeset/small-gifts-refuse.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Redo condition sorting to allow all columns to be sortable, and sort primarily by last recorded date and condition name.
--- a/src/components/core/table/table-helpers.tsx
+++ b/src/components/core/table/table-helpers.tsx
@@ -1,5 +1,5 @@
-import { localeCompareBlankLast, SortDir } from "@/utils/sort";
 import { ReactNode } from "react";
+import { localeCompareBlankLast, SortDir } from "@/utils/sort";
 
 export interface MinRecordItem {
   id: string | number;
--- a/src/components/core/table/table-helpers.tsx
+++ b/src/components/core/table/table-helpers.tsx
@@ -1,4 +1,4 @@
-import { alphaSortBlankLast, SortDir } from "@/utils/sort";
+import { localeCompareBlankLast, SortDir } from "@/utils/sort";
 import { ReactNode } from "react";
 
 export interface MinRecordItem {
@@ -17,8 +17,7 @@ export type TableColumn<T extends MinRecordItem> = {
   className?: string;
   widthPercent?: number;
   minWidth?: number;
-  sortIndex?: keyof T;
-  sortFnOverride?: (a: T, b: T, dir: SortDir) => number;
+  sortIndices?: (keyof T)[];
 } & (DataIndexSpecified<T> | RenderSpecified<T>);
 
 export function sortRecords<T extends MinRecordItem>(
@@ -30,21 +29,32 @@ export function sortRecords<T extends MinRecordItem>(
     (column) => column.title === sort?.columnTitle
   );
   // If there is a sort applied to a column, then sort the records.
-  if (sort && sortColumn) {
-    // May be given a function, or just a property to be sorted by alphanumerically.
-    const sortFn =
-      sortColumn.sortFnOverride ||
-      (sortColumn.sortIndex &&
-        ((a, b, dir) =>
-          // Cast because sortIndex is a property of a and b.
-          alphaSortBlankLast(
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            (a as any)[sortColumn.sortIndex],
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            (b as any)[sortColumn.sortIndex],
-            dir
-          )));
-    if (sortFn) return records.sort((a, b) => sortFn(a, b, sort.dir));
+  if (sort && sortColumn?.sortIndices) {
+    return sortByCols(records, sortColumn.sortIndices, sort.dir);
   }
   return records;
+}
+
+function sortByCols<T>(records: T[], indices: (keyof T)[], dir: SortDir) {
+  return records.sort((a, b) => {
+    let firstCompare = a;
+    let secondCompare = b;
+    if (dir === "desc") {
+      firstCompare = b;
+      secondCompare = a;
+    }
+    let compareTotal = 0;
+    for (let i = 0; i < indices.length; i += 1) {
+      compareTotal +=
+        localeCompareBlankLast(
+          // Cast as any because we know we can index using indices, which are all keys.
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (firstCompare[indices[i]] as any).toString(),
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (secondCompare[indices[i]] as any).toString()
+        ) *
+        0.1 ** i;
+    }
+    return compareTotal;
+  });
 }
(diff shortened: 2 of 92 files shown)


src/components/core/table/table-rows.tsx
```diff
@@ -20,6 +20,7 @@ export const TableRows = <T extends MinRecordItem>({
   emptyMessage,
   handleRowClick,
 }: TableRowsProps<T>) => {
+  console.log("TableRows", records);
```
luthfib, line 23: Remove console.log
```suggestion
```

src/components/core/table/table.tsx
```diff
@@ -50,6 +50,7 @@ export const Table = <T extends MinRecordItem>({
   const [count, setCount] = useState(DEFAULT_PAGE_SIZE);
 
   const sortedRecords = sortRecords(records, columns, sort);
+  console.log("Result of sortRecords(...)", sortedRecords.slice(0, count));
```
luthfib, line 53: Remove 

```suggestion
```

.changeset/small-gifts-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Redo condition sorting to allow all columns to be sortable, and sort primarily by last recorded date and condition name.
```
luthfib, line 5:
```suggestion
Condition sorting now allows all columns to be sortable. Sort primarily defaults to be by last recorded date and condition name.
```

src/components/core/table/table-helpers.tsx
```diff
@@ -30,21 +29,35 @@ export function sortRecords<T extends MinRecordItem>(
...
+    for (let i = 0; i < indices.length; i += 1) {
+      // Sort the primary index reversely if descending.
+      if (i === 0 && dir === "desc") {
+        firstCompare = b;
+        secondCompare = a;
+      }
+      compareTotal +=
+        localeCompareBlankLast(
+          // Cast as any because we know we can index using indices, which are all keys.
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
pmadria3, line 53: do we need this eslint disable?

pmadria3: should we be setting this as any?

melocule: I also thought it shouldn't be needed, but I get this issue when I remove the cast:
image

WesleyKapow: TS is right, you are saying this function takes anything (`T`) and then you try and call `toString()` on whatever `T` is. 

What are the possible types of T? Anything a model getter could return ya?

I think this won't work for numbers, `undefined`, & `null`.
```js
true.toString()
// 'true'
"foo".toString()
// 'foo'
5.toString()
// VM192:1 Uncaught SyntaxError: Invalid or unexpected token
```

Maybe use `${firstCompare[indices[i]]}` instead? Also verify this properly works for undefined and nulls.

WesleyKapow: May also wan't to consider if using lodash would help or not (I'm not sure).

https://stackoverflow.com/questions/41138820/lodash-orderby-with-null-and-real-values-not-ordering-correctly

melocule: Hey! I used your solution Wesley and it worked, thanks for sharing it.

src/components/core/table/table-helpers.tsx
```diff
@@ -30,21 +29,32 @@ export function sortRecords<T extends MinRecordItem>(
...
+      // Sort the primary index reversely if descending.
+      if (i === 0 && dir === "desc") {
+        firstCompare = b;
+        secondCompare = a;
+      }
+      compareTotal +=
+        localeCompareBlankLast(
+          `${firstCompare[indices[i]]}`,
+          `${secondCompare[indices[i]]}`
+        ) *
```
WesleyKapow, line 54: Have you verified behavior when these are `null` or `undefined`? This string conversion will give you the literal string `undefined` which will not be treated as a blank in your comparison function.

WesleyKapow (changes requested):

src/components/core/table/table-helpers.tsx
melocule: Thanks for this callout, just submitted a patch that I tested for this case

WesleyKapow (commented):
hmmm this is fairly complicated code, I can't easily determine "correctness" here and as we noted there are numerous edge cases and tricky aspects to this.

Which means... dun dun dunnn... UNIT TESTS! Please write unit tests to prove correctness in a handful of different situations (null, numbers, undefined, strings, blanks, etc).

src/components/core/table/table-helpers.tsx
```diff
@@ -30,29 +31,39 @@ export function sortRecords<T extends MinRecordItem>(
...
+  records: T[],
+  indices: (keyof T)[],
+  dir: SortDir
+) {
+  // Generates a value that only looks at whether the value at a given index is blank or not.
+  // Keeps blanks last by making them true (higher value) when ascending, false (lower value) when descending.
+  const getIterateeBlanksLast = (o: T, index: keyof T, indexDir: SortDir) =>
+    // ESLint disabled as the index may not exist on this object.
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    (dir === "asc" && !o[index]) || (dir === "desc" && o[index]);
```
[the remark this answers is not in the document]
WesleyKapow: Also `!o[index]` will rule out columns with a zero `0`. This should explicitly check if the value is empty or not.

Check out this helper I made in CTW -- https://github.com/zeus-health/ctw/blob/main/app/utils.ts#L18

I'd probably change it to `isEmptyValue` (not plural).

src/components/core/table/table-helpers.tsx
```diff
@@ -30,29 +31,39 @@ export function sortRecords<T extends MinRecordItem>(
...
+
+  // Makes a list of iteratees, where each index iteratee is preceded by an iteratee that ensures blanks go last.
```
WesleyKapow, line 45: This is a great approach!

src/components/core/table/table-helpers.tsx
```diff
@@ -17,8 +18,7 @@ export type TableColumn<T extends MinRecordItem> = {
   className?: string;
   widthPercent?: number;
   minWidth?: number;
-  sortIndex?: keyof T;
-  sortFnOverride?: (a: T, b: T, dir: SortDir) => number;
+  sortIndices?: (keyof T)[];
 } & (DataIndexSpecified<T> | RenderSpecified<T>);
 
 export function sortRecords<T extends MinRecordItem>(
```
WesleyKapow, line 24: Change this to `T extends Record<string, unknown>` and 🤞 you can then remove your eslint stuff.

src/components/core/table/table-helpers.tsx
```diff
@@ -30,29 +31,39 @@ export function sortRecords<T extends MinRecordItem>(
...
+// Sorts records based on 1 or more indices.
+export function sortByIndex<T>(
+  records: T[],
+  indices: (keyof T)[],
+  dir: SortDir
+) {
+  // Generates a value that only looks at whether the value at a given index is blank or not.
+  // Keeps blanks last by making them true (higher value) when ascending, false (lower value) when descending.
+  const getIterateeBlanksLast = (o: T, index: keyof T, indexDir: SortDir) =>
+    // ESLint disabled as the index may not exist on this object.
```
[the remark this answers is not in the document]
melocule: Oh actually using isEmptyValue(...) removes the need to disable this rule

src/components/core/table/table-helpers.tsx
```diff
@@ -31,33 +33,23 @@ export function sortRecords<T extends MinRecordItem>(
...
-    } else {
-      // Sort other indices ascending. Disable ESlint, index may not exist.
-      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-      iteratees = [...iteratees, (o) => o[sortIndex] === undefined, sortIndex];
-      orders = orders.concat(["asc", "asc"]);
-    }
+  indexSorts.forEach((indexSort) => {
+    const { index, dir } = indexSort;
+    iteratees = iteratees.concat([(o) => isEmptyValue(o[index]), index]);
+    orders = orders.concat(["asc", dir || "asc"]);
```
WesleyKapow, line 52: So we always want blanks last regardless of asc/desc?

WesleyKapow (dismissed):
LGTM, note I left one open question.

src/components/core/table/table-helpers.tsx
melocule: Yupyup, asked the team about this when I was initially working on this ticket and apparently that's the behavior elsewhere

PR #289, opened 2022-11-30
CTW-471 check message length rather than message status

pmadria3:
This PR checks message length as the current API has some issues and makes a decision: 

 If any status messages appear in the below API (regardless of content), the component will render as normal (if there are no conditions returned, shows the no condition available). If empty array is returned, we will show the Request Records view.

Data there (Sarah dev):
Screen Shot 2022-11-30 at 12 15 39 PM

Data not there (Allen dev): 
Screen Shot 2022-11-30 at 12 16 38 PM

### DIFF
--- a/.changeset/silent-cats-run.md
+++ b/.changeset/silent-cats-run.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Show other provider records if the data exists otherwise prompt to request records.
+Show other provider records if patient history messages exists, otherwise show show empty state with request records button.
--- a/.changeset/silent-cats-run.md
+++ b/.changeset/silent-cats-run.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/silent-cats-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Show other provider records if the data exists otherwise prompt to request records.
```
luthfib, line 5:
```suggestion
Show other provider records if patient history messages exists, otherwise show show empty state with request records button.
```

.changeset/silent-cats-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Show other provider records if patient history messages exists, otherwise show show empty state with request records button.
```
luthfib, line 5:
```suggestion
Show other provider records if patient history messages exists, otherwise show empty state with request records button.

```

luthfib (approved):

luthfib (approved):

PR #293, opened 2022-11-30
Ctw 520/on patient save prop

luthfib:
- Refactors so that form validation does not happen inside of the action but instead happens in the drawer form. This eliminates boiler plate as keeping the drawer consistent. Also ensures that we have custom function such as onPropSave they can be decoupled from what goes on in the form. 
- Add onPatientSave prop and adds it to the patient schedule history

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,29 @@ The `PatientProvider` component provides patient details needed by other compone
 
 1. Add `<PatientProvider>` wrapper at a high level within your patient portal (or wherever you have selected down to a single patient).
 2. Pass in identifier information for this patient via `patientID` and `systemURL`.
+3. `<PatientProvider>` also has an onPatientSave prop which can be used to override the behavior when saving patient information. onPatientSave should be an async function.
+
+An example onPatientSave function could be as follows:
+
+```
+const onPatientSave = async (data) => {
+  /* Logic to send data to data
+   data here is of type formResult
+
+    formResult: {
+    success: boolean;
+    data: OmitMatch<ActionReturn<unknown>>;
+    errors: undefined;
+}
+*/
+
+// Do whatever you want with data
+
+  return response
+}
+
+
+```
 
 Example:
 
--- a/README.md
+++ b/README.md
@@ -58,13 +58,20 @@ An example onPatientSave function could be as follows:
 ```
 const onPatientSave = async (data) => {
   /* Logic to send data to data
-   data here is of type formResult
+   data here is of type PatientFormData
 
-    formResult: {
-    success: boolean;
-    data: OmitMatch<ActionReturn<unknown>>;
-    errors: undefined;
-}
+  type PatientFormData = {
+    lastName: string;
+    firstName: string;
+    gender: fhir4.Patient["gender"];
+    dateOfBirth: Date;
+    email: string;
+    phone: string;
+    address: string;
+    city: string;
+    state: string;
+    zipCode: string;
+};
 */
 
 // Do whatever you want with data
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ The `PatientProvider` component provides patient details needed by other compone
 
 1. Add `<PatientProvider>` wrapper at a high level within your patient portal (or wherever you have selected down to a single patient).
 2. Pass in identifier information for this patient via `patientID` and `systemURL`.
-3. `<PatientProvider>` also has an onPatientSave prop which can be used to override the behavior when saving patient information. onPatientSave should be an async function.
+3. `<PatientProvider>` also has an `onPatientSave` prop which can be used to override the behavior when saving patient information. `onPatientSave` should be an async function.
 
 An example onPatientSave function could be as follows:
 
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -361,7 +361,6 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
           patient={patientResponse.data}
           isOpen={requestRecordsDrawerIsOpen}
           onClose={() => {
-            console.warn("This was called. close");
             setRequestDrawerIsOpen(false);
           }}
         />
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -75,7 +75,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
   const { getRequestContext } = useCTW();
   const [sort, setSort] = useState<TableSort>();
 
-  const [clinicalHistoryExists, setClinicalHistoryExists] = useState(false);
+  const [clinicalHistoryExists, setClinicalHistoryExists] = useState<boolean>();
 
   const patientRecordsMessage = patientRecordsResponse.isError
     ? ERROR_MSG
@@ -158,12 +158,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       patientID
     );
 
-    if (patientHistoryFetched) {
-      setClinicalHistoryExists(true);
-      setRequestDrawerIsOpen(false);
-    } else {
-      setClinicalHistoryExists(false);
-    }
+    setClinicalHistoryExists(patientHistoryFetched);
   };
 
   useEffect(() => {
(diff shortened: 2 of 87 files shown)


src/components/content/forms/conditions.ts
```diff
@@ -54,19 +48,10 @@ export function getAddConditionWithDefaults(condition: Condition): Condition {
...
-  formResult: ActionReturn<FormErrors>;
-  requestErrors: string[] | undefined;
-}> => {
-  const result = await getFormData(data, schema);
-  let requestErrors: string[] = [];
-
-  if (!result.success) {
-    return { formResult: result, requestErrors: undefined };
-  }
+  formValidation: { success: boolean; data: any; errors: undefined },
```
pmadria3, line 51: need a type that's not any

pmadria3: I'll wait to review the rest - didn't realize it was in draft

src/components/content/patient-history-request-drawer.tsx
```diff
@@ -1,32 +1,83 @@
...
+      onPatientSaveResponse = await onPatientSave();
+    } else {
+      onPatientSaveResponse = await editPatient(
+        patient,
+        formValidation,
+        getRequestContext
+      );
+    }
+
+    // TODO: How do we want to handle errors from responses that arent defined by us?
```
pmadria3, line 48: just flagging

README.md
```diff
@@ -51,6 +51,29 @@ The `PatientProvider` component provides patient details needed by other compone
 
 1. Add `<PatientProvider>` wrapper at a high level within your patient portal (or wherever you have selected down to a single patient).
 2. Pass in identifier information for this patient via `patientID` and `systemURL`.
+3. `<PatientProvider>` also has an onPatientSave prop which can be used to override the behavior when saving patient information. onPatientSave should be an async function.
```
melocule, line 54: Nit
```suggestion
3. `<PatientProvider>` also has an `onPatientSave` prop which can be used to override the behavior when saving patient information. `onPatientSave` should be an async function.
```

melocule (dismissed):

README.md
```diff
@@ -51,6 +51,29 @@ The `PatientProvider` component provides patient details needed by other compone
 
 1. Add `<PatientProvider>` wrapper at a high level within your patient portal (or wherever you have selected down to a single patient).
 2. Pass in identifier information for this patient via `patientID` and `systemURL`.
+3. `<PatientProvider>` also has an `onPatientSave` prop which can be used to override the behavior when saving patient information. `onPatientSave` should be an async function.
+
+An example onPatientSave function could be as follows:
+
+```
+const onPatientSave = async (data) => {
+  /* Logic to send data to data
```
WesleyKapow, line 60: fix comment

README.md
```diff
@@ -51,6 +51,29 @@ The `PatientProvider` component provides patient details needed by other compone
...
+   data here is of type formResult
+
+    formResult: {
+    success: boolean;
+    data: OmitMatch<ActionReturn<unknown>>;
+    errors: undefined;
+}
```
WesleyKapow, line 67: Why do they care about success or errors? We should do our own validation and only call this callback when the data is valid. They should return success or failure (via async promise resolve/reject).

src/components/core/patient-provider.tsx
```diff
@@ -27,11 +34,19 @@ type ProviderState = {
...
+      data: { npi: string; role: string; name: string };
+    };
+    requestErrors: unknown;
+  }>;
 };
 
 type PatientProviderProps = {
   children: ReactNode;
   tags?: Tag[];
+  onPatientSave?: (data: unknown) => Promise<unknown>;
```
WesleyKapow, line 49: The types for these two `onPatientSave` types don't match. The function should take a well defined data type and should return a promise void, e.g. `(data: PatientData) => Promise<void>`

src/components/core/patient-provider.tsx
```diff
@@ -78,6 +95,24 @@ export function usePatient(): UseQueryResult<PatientModel, unknown> {
   );
 }
 
+export function useHandlePatientSave(patient: PatientModel) {
+  const { getRequestContext } = useCTW();
+  const { onPatientSave } = useContext(CTWPatientContext);
+
+  const handleSave = useCallback(
```
WesleyKapow, line 102: nit: can just `return useCallback`... no need for `handleSave` variable.

src/components/core/patient-provider.tsx
```diff
@@ -78,6 +95,24 @@ export function usePatient(): UseQueryResult<PatientModel, unknown> {
...
+    async (data) => {
+      if (typeof onPatientSave === "function") {
```
WesleyKapow, line 104: It can only be undefined or a function right? Could just do `if (onPatientSave)`

src/components/core/form/drawer-form.tsx
```diff
@@ -2,24 +2,20 @@ import { isEmpty } from "lodash";
...
 export type DrawerFormProps<T> = {
   action: (
-    data: FormData,
-    getRequestContext: () => Promise<CTWRequestContext>,
-    schema: AnyZodSchema
-  ) => Promise<{
-    formResult: ActionReturn<T>;
-    requestErrors: string[] | undefined;
-  }>;
+    data: { success: boolean; data: T; errors: undefined },
```
WesleyKapow, line 16: Do we need to call action on errors? Why not simplify this and only call action when it passes schema validation.

src/components/content/patient-history-request-drawer.tsx
WesleyKapow: We need to wrap the `await onPatientSave` line in a try/catch and have the catch render a generic error

src/components/content/forms/patients.ts
```diff
@@ -1,28 +1,38 @@
...
+  dateOfBirth: Date;
+  email: string;
+  phone: string;
+  address: string;
+  city: string;
+  state: string;
+  zipCode: string;
+};
+
+type OmitMatch<T extends { data: unknown }> = Omit<T, "data"> &
```
WesleyKapow, line 23: ??? what's this doing? Why not use `PatientFormData` directly (and also remove the `Edit` from the type)?

WesleyKapow (changes requested):
If the drawer is handling validation, then let it handle it. No need for the action/onPatientSave/etc to be called with errors. What would a consumer do with that?

I think you should be able to cleanup and hopefully remove all of these `OmitMatch` types.

src/components/core/form/drawer-form.tsx
luthfib: You mean only pass data down to the action? Instead of passing errors and success as well?

WesleyKapow: Yes. Why do actions need that info?

src/components/content/conditions.tsx
```diff
@@ -196,7 +190,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       }
     }
     void load();
-    if (patientResponse.data?.id) {
+    if (patientResponse.data?.id && clinicalHistoryExists === undefined) {
       void handleClinicalHistory(patientResponse.data.id);
     }
```
mrosata, line 195: Is the idea here that `clinicalHistoryExists` is a tribool (true, false, or indeterminate) where an undefined clinicalHistoryExists means "we don't yet know the answer"?

If so we probably could just put that `handleClinicicalHistory` bit in a separate useEffect that only triggers when the `patientResponse.data.id` changes.

```tsx
useEffect(() => {
  if (patientResponse.data?.id) {
      handleClinicalHistory(patientResponse.data.id);
  }
}, [patientResponse.data?.id])
```

src/components/content/conditions.tsx
```diff
@@ -76,7 +75,8 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
   const { getRequestContext } = useCTW();
   const [sort, setSort] = useState<TableSort>();
 
-  const [clinicalHistoryExists, setClinicalHistoryExists] = useState(false);
+  const [clinicalHistoryExists, setClinicalHistoryExists] =
+    useState<boolean>(false);
```
mrosata, line 79:
```suggestion
  const [clinicalHistoryExists, setClinicalHistoryExists] = useState(false);
```

mrosata (approved):

PR #295, opened 2022-12-01
Fix bug with detecting active conditions

WesleyKapow:
CTW-577

Previously we used `this.clinicalStatus` which uses `codeableConceptLabel` which grabs the `text` field if it exists. The production bug was that this value was `Active` (capitalized). We now go off the actual `code` for the correct `system`. NOTE: Code's here HAVE TO BE lowercase (ODS rejects them otherwise).


luthfib (approved):

pmadria3 (approved):
LGTM

PR #296, opened 2022-12-01
Ctw 536/multiple elem in single row

luthfib:
Add ability to include multiple elements in one row in forms.

image

Still need to: 
- [x] Add optional word to field that are optional

### DIFF
--- a/src/components/core/form/drawer-form-with-fields.tsx
+++ b/src/components/core/form/drawer-form-with-fields.tsx
@@ -1,6 +1,8 @@
+import { isArray } from "lodash";
 import { InputHTMLAttributes, ReactNode } from "react";
-import { DrawerForm, DrawerFormProps } from "./drawer-form";
+import { DrawerForm, DrawerFormProps, FormErrors } from "./drawer-form";
 import { FormField } from "@/components/content/forms/form-field";
+import { FormFieldLabel } from "@/components/content/forms/form-field-label";
 import { AnyZodSchema, useFormInputProps } from "@/utils/form-helper";
 
 export type FormFieldType = {
@@ -26,7 +28,7 @@ export type FormContentType = {
   label: string;
 };
 
-export type FormEntry = FormFieldType | FormContentType;
+export type FormEntry = FormFieldType | FormContentType | FormFieldType[];
 
 export type DrawerFormWithFieldsProps<T> = {
   title: string;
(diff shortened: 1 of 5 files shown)


WesleyKapow:
screens or it didn't happen

luthfib:
> screens or it didn't happen

Good point.

src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -58,77 +60,127 @@ export const DrawerFormWithFields = <T,>({
...
-                      errors={fieldErrors}
-                      hidden={hidden}
-                      render={render}
-                    />
-                  );
-                }
-
-                const props = inputProps(field, schema);
-                const required = props["aria-required"];
+            {data.map((record) => {
```
WesleyKapow, line 63: nit: s/record/entry

`record` is a little confusing

src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -58,77 +60,127 @@ export const DrawerFormWithFields = <T,>({
...
+                    disabled={submitting}
+                    readonly={readonly}
+                    defaultValue={value}
+                    errors={fieldErrors}
+                    hidden={hidden}
+                    render={render}
+                  />
+                </div>
+              );
+            })}
```
WesleyKapow, line 131: This all looks redundant to what is in `FormFieldEntry`, how about a `FormFieldEntry` and a `FormFieldEntries` which uses it?

src/components/core/main.scss
```diff
@@ -34,6 +34,7 @@
     border-style: solid;
     background-color: initial;
     resize: none;
+    height: 2.25rem;
```
WesleyKapow, line 37: hmmm specific height's are a little scary. Height should usually be determined by font-size/padding/etc.

Why was this needed?

WesleyKapow (changes requested):
Jumped on this PR because I was curious how you solved it, love the solution, very elegant!

src/components/core/main.scss
luthfib: Specifically targeting this use case:

image

I can play around with it and see I find a solution that doesn't require specific height setting.

melocule: Not so related to this PR but I just noticed an issue with the border color for when the listbox is selected, would you mind fixing it here? I noticed that it's purple for the other fields, but blue for the listboxes which looks a bit awkward
image
image

luthfib: I'll address it in a followup pr

WesleyKapow (dismissed):
Approving with a few nits.

src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -136,51 +124,74 @@ export const DrawerFormWithFields = <T,>({
...
-              defaultValue={value}
-              errors={fieldErrors}
-              hidden={hidden}
-              render={render}
-            />
-          </div>
-        );
-      })}
+      {recordList.map((record) => (
+        <Fragment key={record.label}>
```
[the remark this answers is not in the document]
luthfib: I only did `Fragment` because I need to add the React key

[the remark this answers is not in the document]
WesleyKapow: Ah of course, for some reason I was thinking you could do that without but of course not.

src/components/core/form/drawer-form-with-fields.tsx
```diff
@@ -64,14 +64,15 @@ export const DrawerFormWithFields = <T,>({
         <div className="ctw-space-y-4">
           {header}
           <div className="ctw-space-y-6">
-            {data.map((entry) => {
+            {data.map((entry, idx) => {
```
WesleyKapow, line 67: nit: idx not used?

WesleyKapow (approved):

PR #301, opened 2022-12-02
Minor cleanup

luthfib:
- Revert useEffect based on feedback that it is more correct to have `handleClinicalHistory ` called once based on state being undefined. 
- Update readme
- Change logic around showing empty "request records" box in other providers to only show if `clinicalHistoryExists` OR patient has no other provider records.

### DIFF
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -75,7 +75,7 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
   const { getRequestContext } = useCTW();
   const [sort, setSort] = useState<TableSort>();
 
-  const [clinicalHistoryExists, setClinicalHistoryExists] = useState(false);
+  const [clinicalHistoryExists, setClinicalHistoryExists] = useState<boolean>();
 
   const patientRecordsMessage = patientRecordsResponse.isError
     ? ERROR_MSG
@@ -190,6 +190,11 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       }
     }
     void load();
+    if (patientResponse.data?.id && clinicalHistoryExists === undefined) {
+      void handleClinicalHistory(patientResponse.data.id);
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     includeInactive,
     patientResponse.data,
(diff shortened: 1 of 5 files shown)


src/components/content/conditions.tsx
```diff
@@ -190,21 +190,20 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       }
     }
     void load();
+    if (patientResponse.data?.id && clinicalHistoryExists === undefined) {
```
WesleyKapow, line 193: This is fine, you also could have kept it pulled out into its own useEffect but added the guard.

And actually now that I think about it.. I think the guard should live in `handleClinicalHistory` itself. That way, anywhere we call it, the function itself takes care to never refetch.

While you're at it, `handleClinicalHistory` is also not a great name. "handle" is used for `on*` props like `onClick` goes with `handleClick`. How about `checkClinicalHistory`.

WesleyKapow (dismissed):
Minor nits but lgtm

src/components/content/conditions.tsx
luthfib: Im actually debating whether we should have the undefined check and whether we should add the useEffect back. Reason being is... what if someone makes a patient history request when they havent made one in the past. Would would want the `checkClinicalHistory` to fire to hide that request screen.

WesleyKapow: Yes this is a great callout. How would avoiding the `undefined` check help that case? What should we do here?

WesleyKapow: Yes this is a great callout. How would avoiding the `undefined` check help that case? What should we do here?

luthfib: Actually you are right, avoiding undefined wouldnt change that here

luthfib: Easiest thing might honestly be to put this into react query hook, and just invalidate it when calling patient history in the drawer

WesleyKapow: Could we call `setClinicalHistoryExists(true)` when patient history request form is saved?

luthfib: We could, the only issue I see with that I'd prefer to do a refetch as it is more correct because... the history api returns a 500 currently by default. So checking if it was successful might be trick. I think it'd be better to refetch to see if there are any current messages

WesleyKapow (approved):
LGTM

PR #311, opened 2022-12-06
Do patch for patient edit

luthfib:
use PATCH instead of PUT for patient updates.


WesleyKapow:
NOTE: This was closed as we will likely not use PATCH and instead fetch the existing resource, update the fields we care about, and save it using PUT. This is because PATCH requires us to know if a field exists or not so that we can correctly use `add` or `replace`. This requirement removes much of the benefit of using PATCH.

PR #316, opened 2022-12-08
CTW-572/patient conditions component

WesleyKapow:
First pass at setting up new `PatientConditions` component (conditions 2.0). Special care is taken to try and reduce how much code is actually in the main component and instead rely on more helpers. Also worth noting that filtering does not happen within a `useEffect` as it should be fast and it is ok if it happens every time the component re-renders.

TODO:
* Pagination
* "..." menu

Screen Shot 2022-12-08 at 1 01 37 PM

Screen Shot 2022-12-08 at 1 01 22 PM

Screen Shot 2022-12-09 at 10 13 53 AM

### DIFF
--- a/src/components/content/conditions/helpers.ts
+++ b/src/components/content/conditions/helpers.ts
@@ -1,10 +1,11 @@
 import { ConditionModel } from "@/fhir/models";
 
 // Filter out other conditions where:
-//  1. There is an existing patient condition with a matching known code.
-//  2. The other condition is older than the patient condition OR they
+//  1. Condition is archived and includeArchived is false.
+//  2. CCS Category code starts with FAC or XXX.
+//  3. There is an existing patient condition with a matching known code.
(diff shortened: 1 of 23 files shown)


src/components/content/conditions/helpers.ts
```diff
@@ -13,6 +14,10 @@ export const filterOtherConditions = (
   otherConditions.filter((otherCondition) => {
     if (otherCondition.isArchived && !includeArchived) return false;
 
+    if (["FAC", "XXX"].includes(otherCondition.ccsChapterCode ?? "")) {
+      return false;
+    }
```
luthfib, line 19: Curious what `FAC` and `XXX` are? Is this something that could potentially be set as variable with a descriptive name?

luthfib (approved):

luthfib (approved):

PR #317, opened 2022-12-08
Ctw 476/ccd condition history viewer

pmadria3:
This PR helps to incorporate the CCD into the condition history page to help provide source of truth and additional information for the conditions if there is a CCD document for the condition. If there is no CCD document associated with the file then no document will show up.

https://user-images.githubusercontent.com/98342697/209174468-585c39a1-374e-43d9-8283-bfa4dcc2a86f.mov


src/components/content/CCDA/ccda_viewer/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { DocumentDetails } from "./components/DocumentDetails";
+import { Header } from "./components/Header/Header";
+import { Section } from "./components/Section/Section";
+import { DocumentOnlyProps, SectionType } from "./types";
+
+const useStyles = makeStyles(() => ({
+  container: {
+    width: "calc(100vw - 200px - 24px - 24px - 16px - 16px - 2px)",
+  },
+}));
```
mrosata, line 12: We probably shouldn't add `@material-ui/core` _(ie: `makeStyles()`)_ to the project.

These can just be refactored out into SCSS and/or with the `classnames` library

src/components/content/CCDA/ccda_viewer/helpers/parseToISOString.ts
```diff
@@ -0,0 +1,153 @@
...
+            result += `${timezoneSign}${timezoneHour}`;
+            if (isCorrectValue(timezoneMinute)) result += `:${timezoneMinute}`;
+            else result += ":00";
+          } else result += "-00:00";
+        } else result += "T00:00:00-00:00";
+      }
+    }
+  }
+  return result;
+};
```
mrosata, line 100: Even though this was probably copied from fhirplace and we may know that it works, it would be better to use a date library to parse and build date stings. The one that is already installed is [`date-fns`](https://date-fns.org/).

> `date-fns` depends on `moment.js` which is a deprecated library as of a few years ago, eventually we may want to migrate off `date-fns`, but that's for another day.

src/components/content/CCDA/ccda_viewer/helpers/parseMany.ts
```diff
@@ -0,0 +1,10 @@
+import { isEmpty } from "lodash";
+
+export function parseMany<T>(
+  parser: (el: Document) => T,
+  data: Document[]
+): T[] {
+  if (isEmpty(data)) return [];
+
+  return data.map((val) => parser(val));
+}
```
mrosata, line 10: This file isn't needed. Assuming this was added because:
```ts
const data = null;

data.map(parser);  // Throws
```
would throw an error because `.map` only works on Arrays.. but lodash `_.map` already accounts for this. So all the instances of parseMany can be replaced with lodash map.

An example:
```ts
      given: parseMany<string>(
        (n) => String(xpath.select1("string(text())", n)),
        xpath.select("(*[name()='given'])", nameXml) as Document[]
      ),
```
Would refactor into:
```ts
import { map } from "lodash/fp";

      given: map(
        (n) => `${xpath.select1("string(text())", n)}`,
        xpath.select("(*[name()='given'])", nameXml)
      ),
```
or if using "lodash" instead of "lodash/fp"
```ts
import { map } from "lodash;

      given: map(
        xpath.select("(*[name()='given'])", nameXml),
        (n) => `${xpath.select1("string(text())", n)}`
      ),
```

package.json
```diff
@@ -103,14 +104,22 @@
   "dependencies": {
     "@headlessui/react": "^1.7.3",
     "@heroicons/react": "^1.0.6",
+    "@material-ui/core": "^4.12.4",
```
luthfib, line 107: I think we if we can avoid installing material-ui/core, which seeing the surface area it is not that big, we should. This is quite a big library, and we dont use it much. Either in this pr or a additional ticket.

src/fhir/conditions.ts
```diff
@@ -158,3 +162,84 @@ function filterAndSort(conditions: ConditionModel[]): ConditionModel[] {
...
+) => {
+  const practitioner = await getPractitioner(practitionerId, requestContext);
+  const display = practitioner.fullName;
+
+  return {
+    reference: `Practitioner/${practitionerId}`,
+    type: "Practitioner",
+    display,
+  };
+};
```
luthfib, line 178: Why did you add this? I don't see it being used anywhere

src/components/core/modal-ccda.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  useEffect(() => {
+    if (rawBinary.isBinary) {
+      setIsXMLData(true);
+    }
+  }, [isXMLData, rawBinary.isBinary]);
+
+  return (
+    <Modal {...modalProps}>
+      {isXMLData && rawBinary.xmlBinary?.data && rawBinary.xmlBinary.data && (
```
luthfib, line 27: why do we have `rawBinary.xmlBinary?.data && rawBinary.xmlBinary.data`. Seems you are checking the same thing twice

src/components/content/conditions-history.tsx
```diff
@@ -180,17 +238,49 @@ export function ConditionHistory({
...
+                    <DocumentButton
+                      onClick={() => {
+                        setIsModalOpen(true);
+                        setRawBinary(idMap.get(entry.id));
+                      }}
+                    />
+                  )}
+                </>
+              ),
+            }))}
```
luthfib, line 262: This is really confusing use of children. It just display next to the button. Please change this to be something a bit more indicative of what it is actually doing. Might need to change it in Details as well. This prop makes me think I would be changing the actual content in the component.

src/components/content/conditions-history.tsx
```diff
@@ -141,6 +162,29 @@ export function ConditionHistory({
         setConditionsWithDate(conditionsDataDeduped.filter((d) => d.date));
         setConditionsWithoutDate(conditionsDataDeduped.filter((d) => !d.date));
         setLoading(false);
+
+        // Binary Doc stuff
+        const requestContext = await getRequestContext();
+
+        const turnConditiontoDataList = setupData(condition);
```
luthfib, line 169: `turnConditiontoDataList` is a confusing name. Still not quite sure what it is

src/components/content/conditions-history.tsx
```diff
@@ -141,6 +162,29 @@ export function ConditionHistory({
...
+        );
+
+        setIDMap(binaryDocs);
+
+        if (binaryDocs.get(condition.id)?.isBinary) {
+          setIsBinaryDocumentForOriginalEntry(true);
+          setRawBinary(binaryDocs.get(condition.id));
+        }
+
+        setLoadingSourceDocument(false);
```
luthfib, line 187: This whole block could probably be a function in itself.

src/components/content/conditions-history.tsx
```diff
@@ -107,10 +115,23 @@ export function ConditionHistory({
...
     useState<CollapsibleDataListStackEntries>([]);
   const [loading, setLoading] = useState(true);
+  const [loadingSourceDocument, setLoadingSourceDocument] = useState(true);
   const [conditionForSearch, setConditionForSearch] =
     useState<ConditionModel>();
   const historyResponse = useConditionHistory(conditionForSearch);
+  const { getRequestContext } = useCTW();
+
+  const [
+    isBinaryDocumentForOriginalEntry,
```
luthfib, line 125: What is `isBinaryDocumentForOriginalEntry`. What is original entry?

src/components/content/CCDA/ccda_viewer/components/Custodian/data.ts
```diff
@@ -0,0 +1,32 @@
...
+export const getCustodianData = (
+  document: Document
+): GeneralInfo | undefined => {
+  const representedCustodianOrganization = xpath.select1(
+    "*[name()='ClinicalDocument']/*[name()='custodian']/*[name()='assignedCustodian']/*[name()='representedCustodianOrganization']",
+    document
+  ) as Document;
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!representedCustodianOrganization) return undefined;
```
luthfib, line 14: I think adding   `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` is actually wrong. The reason why it thinks it is not necessary is because you cast `representedCustodianOrganization` as Document. If xpath can return undefined then you could also just cast `as Document | undefined`

luthfib (commented):
Seems we have a lot of `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` in the code that was ported over. In almost all instances it is incorrect use of it. Not sure how worth it is to fix some of those since we will be maintaining this code.

pmadria3:
Moving storybook to follow up 2.0 ticket

src/components/content/CCDA/ccda_viewer/components/Authorization/Authorization.tsx
```diff
@@ -0,0 +1,23 @@
+import { isEmpty } from "lodash";
+import { DocumentOnlyProps } from "../../types";
+import { Table } from "../Table/Table";
+import { getAuthorizationData } from "./data";
+
+export const Authorization = ({ document }: DocumentOnlyProps) => {
+  const authorizations = getAuthorizationData(document);
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!authorizations || isEmpty(authorizations)) return null;
```
luthfib, line 10: Why are we checking the same thing twice? We are checking if it is not empty and then checking isEmpty right after it. Seems redundant?

src/components/content/CCDA/ccda_viewer/components/InFulfillmentOf/InFulfillmentOf.tsx
```diff
@@ -0,0 +1,24 @@
+import { isEmpty } from "lodash";
+import { DocumentOnlyProps } from "../../types";
+import { Table } from "../Table/Table";
+import { getInFulfillmentOfData } from "./data";
+
+export const InFulfillmentOf = ({ document }: DocumentOnlyProps) => {
+  const inFulfillmentOfs = getInFulfillmentOfData(document);
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!inFulfillmentOfs || isEmpty(inFulfillmentOfs)) return null;
```
luthfib: Same here as the comment above

src/components/content/CCDA/ccda_viewer/helpers/getHumanName.ts
```diff
@@ -0,0 +1,51 @@
...
+  R: "Religious",
+  SNDX: "Soundex",
+  SRCH: "Search",
+  SYL: "Syllabic",
+};
+
+export const getHumanName = (name: Document | Document[]): string => {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!name) return "";
+
```
luthfib, line 25: Remove   `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` and make name a optional param

src/components/content/CCDA/ccda_viewer/helpers/getPeriod.ts
```diff
@@ -0,0 +1,46 @@
+import { isBefore, isEqual } from "date-fns";
+import { Period } from "fhir/r4";
+import xpath from "xpath";
+import { ccdaDatetimeToISO, displayDateTimeasString } from "@/fhir/formatters";
+
+export const getPeriod = (xmlData: Document): Period => {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!xmlData) return {};
```
luthfib, line 8: Make `xmlData` optional if we are going to have that check

src/App.tsx
```diff
@@ -16,7 +16,7 @@ const {
   VITE_AUTH0_CLIENT_ID,
   VITE_AUTH0_AUDIENCE,
   VITE_AUTH0_CALLBACK_PATH,
-  VITE_ENV = "dev",
+  VITE_ENV = "sandbox",
```
luthfib, line 19: Don't change this

pmadria3: good catch

src/fhir/conditions.ts
```diff
@@ -174,3 +177,70 @@ function filterAndSort(conditions: ConditionModel[]): ConditionModel[] {
...
+              headers: {
+                Authorization: `Bearer ${requestContext.authToken}`,
+                accept: "application/json",
+                "Content-Type": "application/json+fhir",
+              },
+            });
+            documentBinary = await response.json();
+          }
+        }
+      }
```
luthfib, line 235: I am pretty sure we can change this to....

```suggestion
 if (conditionsJSON.entry[0].resource.entity[0].role === "source") {
        const binaryID =
          conditionsJSON.entry[0].resource.entity[0].what.reference;

        const endpointBinaryUrl = `${getZusApiBaseUrl(
          requestContext.env
        )}/fhir/${binaryID}`;

        const response = await fetch(endpointBinaryUrl, {
          headers: {
            Authorization: `Bearer ${requestContext.authToken}`,
            accept: "application/json",
            "Content-Type": "application/json+fhir",
          },
        });
        documentBinary = await response.json();
      }
```

src/components/content/conditions-history.tsx
```diff
@@ -183,14 +237,42 @@ export function ConditionHistory({
...
+                        <DocumentButton
+                          onClick={() => {
+                            setIsModalOpen(true);
+                            setRawBinary(idMap.get(entry.id));
+                          }}
+                        />
+                      )}
+                    </>
+                  ),
+                }))}
```
luthfib, line 275: Can we make a component for this? They seem almost identical

luthfib (approved):

mrosata (dismissed):

PR #321, opened 2022-12-09
Revert "Revert "CTW-555 request header""

melocule:
Reverts zeus-health/ctw-component-library#319

[Original PR](https://github.com/zeus-health/ctw-component-library/pull/315)

### DIFF
--- a/.changeset/friendly-berries-tease.md
+++ b/.changeset/friendly-berries-tease.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Re-add header that indicates the source of outgoing requests.
--- a/.changeset/quiet-cats-repeat.md
+++ b/.changeset/quiet-cats-repeat.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Re-add header changes.
--- a/.changeset/quiet-cats-repeat.md
+++ b/.changeset/quiet-cats-repeat.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Re-add header changes.
+Remove header indicating source of requests.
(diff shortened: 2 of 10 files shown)


.changeset/friendly-berries-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Re-add header that indicates the source of outgoing requests.
```
luthfib, line 5:
```suggestion
Re-add header that indicates the source of outgoing requests..
```

luthfib:
I think @mrosata suggestion makes sense where he said we could do:

```
const { fetch: originalFetch } = window;

const {
  SERVICE_NAME,
  ZUS_HEADER_NAME,
} = process.env;

window.fetch = async (...args) => {
  let [resource, config ] = args;

  if (isZusOrigin(resource)) {
    config.headers[ZUS_HEADER_NAME] = SERVICE_NAME;
  }

  return originalFetch(resource, config);
};
```

@eric-gunther @mrosata would this get flagged at all by security teams since we are modifying the window fetch?

eric-gunther:
> I think @mrosata suggestion makes sense where he said we could do:
> 
> ```ts
> const { fetch: originalFetch } = window;
> 
> const {
>   SERVICE_NAME,
>   ZUS_HEADER_NAME,
> } = process.env;
> 
> window.fetch = async (...args) => {
>   let [resource, config ] = args;
> 
>   if (isZusOrigin(resource)) {
>     config.headers[ZUS_HEADER_NAME] = SERVICE_NAME;
>   }
> 
>   return originalFetch(resource, config);
> };
> ```
> 
> @eric-gunther @mrosata would this get flagged at all by security teams since we are modifying the window fetch?

Couldn't this potentially break functionality in whatever app is embedding our component? I suppose it's limited to just requests that have a `resource` argument but it still seems too broad to me.

pmadria3 (commented):
This looks good to me if we are keeping the original PR @luthfib @eric-gunther thoughts?

luthfib (approved):

PR #325, opened 2022-12-10
Ctw 600/add error msg to bottom

luthfib:
Add error message at bottom of drawer in addition to top of drawer.

### DIFF
--- a/.changeset/eleven-bulldogs-kneel.md
+++ b/.changeset/eleven-bulldogs-kneel.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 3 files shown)


eric-gunther (commented):
Testing locally, error content doesn't exactly match that at the top of the screen (missing the details line) - unsure if that's intentional.

image

luthfib:
> Testing locally, error content doesn't exactly match that at the top of the screen (missing the details line) - unsure if that's intentional.
> 
> image

My understanding is that the bottom is supposed to be a generic message and the top is supposed to be more descriptive. I can double check with UX, but the top can also be lists, so wouldnt make much sense for it to match the bottom.

eric-gunther (approved):

PR #329, opened 2022-12-13
CTW-597/PatientTable

mrosata:
- Created `<PatientTable />` component based on the one in ctw.
- Created a new `<Pagination />` component that lazy fetches and caches pages.
- The pagination component previously used is now named `<PaginationList />` because the newer pagination component more aptly fits the description of "paging". The older pagination was only used internally by components, should be no issue with name change.
- Added storybook story for patients table. The story uses the faker js library to generate fhir resources for the mocked API.

### DIFF
--- a/src/components/content/patients/patients-table.scss
+++ b/src/components/content/patients/patients-table.scss
@@ -1,0 +1,25 @@
+.ctw-patients-table {
+  @apply ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white;
+  .ctw-patients-table-email {
+    @apply ctw-text-content-black;
+  }
+  .ctw-patients-table-phone {
+    @apply ctw-text-content-light;
+  }
+  .ctw-patients-table-search {
+    @apply ctw-listbox-button ctw-bg-bg-white ctw-py-2 ctw-pl-10 ctw-pr-3;
+  }
+}
+
+.ctw-patients-table.ctw-patients-table-stacked {
+  .ctw-body-container {
+    @apply ctw-m-0;
+  }
+}
+
+.ctw-search-icon-wrapper {
+  @apply ctw-pointer-events-none ctw-absolute ctw-inset-y-0 ctw-left-0 ctw-flex ctw-items-center ctw-pl-3;
+  .ctw-search-icon {
+    @apply ctw-h-5 ctw-w-5 ctw-text-content-lighter;
+  }
+}
--- a/src/components/content/patients/patients-table.scss
+++ b/src/components/content/patients/patients-table.scss
@@ -1,13 +1,13 @@
 .ctw-patients-table {
   @apply ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white;
 
-  .ctw-patients-table-email {
+  &-email {
     @apply ctw-text-content-black;
   }
-  .ctw-patients-table-phone {
+  &-phone {
     @apply ctw-text-content-light;
   }
-  .ctw-patients-table-search {
+  &-search {
     @apply ctw-listbox-button ctw-bg-bg-white ctw-py-2 ctw-pl-10 ctw-pr-3;
   }
 }
--- a/src/components/content/patients/patients-table.scss
+++ b/src/components/content/patients/patients-table.scss
@@ -8,7 +8,7 @@
     @apply ctw-text-content-light;
   }
   &-search {
-    @apply ctw-listbox-button ctw-bg-bg-white ctw-py-2 ctw-pl-10 ctw-pr-3;
+    @apply ctw-rounded-md ctw-border ctw-border-solid ctw-border-icon-light ctw-bg-bg-white ctw-px-3 ctw-py-2 ctw-pl-10 ctw-pr-3 ctw-text-sm ctw-shadow-sm;
   }
 }
 
--- a/src/components/content/patients/patients-table.scss
+++ b/src/components/content/patients/patients-table.scss
@@ -1,5 +1,6 @@
 .ctw-patients-table {
   @apply ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white;
+
   .ctw-patients-table-email {
     @apply ctw-text-content-black;
   }
@@ -19,6 +20,7 @@
 
 .ctw-search-icon-wrapper {
   @apply ctw-pointer-events-none ctw-absolute ctw-inset-y-0 ctw-left-0 ctw-flex ctw-items-center ctw-pl-3;
+
   .ctw-search-icon {
     @apply ctw-h-5 ctw-w-5 ctw-text-content-lighter;
   }
--- a/src/components/content/patients/patients-table.tsx
+++ b/src/components/content/patients/patients-table.tsx
@@ -1,0 +1,180 @@
+import type { TableColumn } from "@/components/core/table/table-helpers";
+import type { PatientModel } from "@/fhir/models/patient";
+import { SearchIcon } from "@heroicons/react/solid";
+import type { Argument } from "classnames";
+import cx from "classnames";
+import { debounce } from "lodash";
+import { useCallback, useEffect, useState } from "react";
+import * as CTWBox from "@/components/core/ctw-box";
+import { useQueryWithCTW } from "@/components/core/ctw-provider";
+import { Pagination } from "@/components/core/pagination/pagination";
+import { Table } from "@/components/core/table/table";
+import { MinRecordItem } from "@/components/core/table/table-helpers";
+import { getBuilderPatientsList } from "@/fhir/patient-helper";
+import { QUERY_KEY_PATIENTS_LIST } from "@/utils/query-keys";
+import "./patients-table.scss";
+
+export type PatientsTableProps = {
+  className?: cx.Argument;
+  handleRowClick: (row: PatientModel) => void;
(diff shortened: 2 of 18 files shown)


src/components/content/patients/patients-table.scss
```diff
@@ -0,0 +1,27 @@
+.ctw-patients-table {
+  @apply ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white;
+
+  .ctw-patients-table-email {
```
melocule, line 4: Can you replace these nested classes with the format `$-email` etc? It should compile to `.ctw-patients-table-email`
https://css-tricks.com/the-sass-ampersand/#aa-modifying-the-ampersand

src/components/content/patients/patients-table.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { useQueryWithCTW } from "@/components/core/ctw-provider";
+import { Pagination } from "@/components/core/pagination/pagination";
+import { Table } from "@/components/core/table/table";
+import { MinRecordItem } from "@/components/core/table/table-helpers";
+import { getBuilderPatientsList } from "@/fhir/patient-helper";
+import { QUERY_KEY_PATIENTS_LIST } from "@/utils/query-keys";
+import "./patients-table.scss";
+
+export type PatientsTableProps = {
+  className?: cx.Argument;
```
melocule, line 18: This is so cool! didn't know you could type className like that

src/components/content/patients/patients-table.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * PatientsTable displays a paginated list of all patients for a builder. In
+ * addition to having configurable page size, lazy loading and name search, the
+ * component accepts an `onRowClick` prop so developers can add their own
+ * logic when a row is clicked. The `onRowClick` receives the targeted
+ * `PatientModel` as its sole argument, which contains the underlying FHIR
+ * object as `.resource`.
+ *
+ * This component uses the `usePatientsList` hook to query Zus ODS. If devs
```
melocule, line 51: Could this part of the documentation for `PatientsTable` down be moved to where `usePatientsList(...)` is defined since it describes that helper? The docs for this component are v long

src/components/content/patients/story-helpers/mocks/requests.ts
```diff
@@ -0,0 +1,156 @@
...
+      msw: [
+        rest.get("https://api.dev.zusapi.com/fhir/Patient", (req, res, ctx) => {
+          const params = req.url.searchParams;
+          if (params.has("_count")) {
+            const count = parseInt(params.get("_count") || "0", 10);
+            const offset = parseInt(params.get("_offset") || "0", 10);
+            const filteredPatients = patients.filter((patient) => {
+              const { given, family } = patient.resource?.name?.[0] || {};
+              return (
+                `${given?.[0] || ""} ${family || ""}`
```
melocule, line 30: If we have the family name but not the given name, this would return " Lastname" (with a space before it) instead of "Lastname"-- is this what we want?

src/components/content/patients/story-helpers/mocks/requests.ts
```diff
@@ -0,0 +1,156 @@
...
+          family: faker.name.lastName(),
+          given: [faker.name.firstName()],
+          prefix: [faker.name.prefix()],
+        },
+      ],
+      telecom: [
+        {
+          system: "phone",
+          value: faker.phone.number(
+            faker.helpers.arrayElement(PHONE_NUMBER_FORMATS)
```
melocule, line 138: I think this is a nit: 
If PHONE_NUMBER_FORMATS is a constant, can the other arrays like ["home", "work", "mobile"] be defined as constants too? Or making none of them constants and putting them directly in the code

src/components/core/ctw-provider.tsx
```diff
@@ -154,6 +154,24 @@ function useCTW() {
   };
 }
 
+export function useQueryWithCTW<T, T2>(
```
melocule, line 157: Instead of `T` and `T2`, could we give these generic types meaningful name like `Value` and `Key`? Not sure of what conventions we want to use here ( @luthfib ?), I've also seen people use `T` and then subsequent letters like `U`

melocule (changes requested):

eric-gunther:
I did some testing locally. Issues found (see video below):

- I had 6 records total, 5 on first page. When I clicked 2nd page I saw 5 records still, just shifted so I records 2 through 6. Pretty sure we want that page to just show 1 record.
- After removing my search the page number in the bottom left got set to 0.
- I think a loading state is a requirement and not a nice-to-have here. When you page or search and it takes more than a second to load data (which I think will be common with large patient lists) we should really show a loading spinner.

https://user-images.githubusercontent.com/97455910/207636139-1ac737d1-e6c4-4cd6-af17-902b23731f4c.mov

src/components/content/patients/patients-table.scss
mrosata: good point, forgot about using ampersand for creating new class names.

src/components/content/patients/patients-table.tsx
mrosata: I do agree that it's long for a code doc, but storybook only parses the docstrings for the component being storied. The other way to do this would be in a markdown file.

@WesleyKapow you had advocated for only using doc-strings and not markdown files for stories, maybe you want to weigh in quickly. The obvious paths available:
 - keep the comment as-is (maybe try to trim some)
 - don't document the usePatientsList hook on storybook, would probably move some of the docstring down to the other usePatientsList function, (but if it's not displayed in storybook it probably doesn't need the code example)
 - use a markdown file for long comments

src/components/content/patients/story-helpers/mocks/requests.ts
mrosata: For this case it's fine, we're just tying to mock a patient search filter so the value won't actually be displayed anywhere. I could trim the string to avoid someone typing a `" "` into the storybook search _(that just return all results as-is)_.

Probably the search input should be doing a trim also before making the web request so `"   mich"` and `"mich"` would both match the same set of patients.

src/components/content/patients/story-helpers/mocks/requests.ts
mrosata: I can have this bit be a const, it wasn't used more than once atm anyways

src/components/core/ctw-provider.tsx
mrosata: I would prefer more descriptive names if anything.

I chose `T2` simply b/c `T` is what is used around the codebase. I imagine `T` standing for "type", kind of how react uses `P` in types for "prop", or `Record<K, V>` might be "key" and "value".

I've noticed lodash uses both forms, sometimes it writes type defs like `<TObject, TSource1, TSource2, TSource3, TSource4>` and other times like `<T1, T2, T3, T4, T5, R>`.

If we don't shift to more descriptive naming, `T2` could be `K` if people really don't like T2.

src/components/content/patients/patients-table.tsx
mrosata: > * I think a loading state is a requirement and not a nice-to-have here. When you page or search and it takes more than a second to load data (which I think will be common with large patient lists) we should really show a loading spinner.

For this bit @eric-gunther, probably need to talk to @tai-terry, as I thought on this the other day it wasn't obvious where the loading state should go.
 - Does the whole table become a loading state? _(if so might want make sure sure this is good UX and do the work in another ticket as there are still considerations such as whether the spinner is just shown over the results there already or do we clear the rows to show a spinner? Clearing the rows would collapse the whole table so then maybe we do it as an overlay instead so the height of the table only changes once the new data loads in)_.
 - Do we just put a loading spinner below/above the table? _(If so this is easy enough to put in this PR)_

mrosata:
> * I think a loading state is a requirement and not a nice-to-have here. When you page or search and it takes more than a second to load data (which I think will be common with large patient lists) we should really show a loading spinner.

For this bit @eric-gunther, probably need to talk to @tai-terry, as I thought on this the other day it wasn't obvious where the loading state should go.
 - Does the whole table become a loading state? _(if so might want make sure sure this is good UX and do the work in another ticket as there are still considerations such as whether the spinner is just shown over the results there already or do we clear the rows to show a spinner? Clearing the rows would collapse the whole table so then maybe we do it as an overlay instead so the height of the table only changes once the new data loads in)_.
 - Do we just put a loading spinner below/above the table? _(If so this is easy enough to put in this PR)_

mrosata:
@eric-gunther Added a loading spinner to the bottom of table that replaces (Showing n of n results) while records are being fetched. As you pointed out I believe, that count changes to (Showing 0 of 0 results) after a search which isn't correct, but showing the count unchanged while search results are being fetched feels frozen. So this isn't perfect but slightly better than what ctw has currently. I synced w/Tai quickly, she agrees this is good for now (better than what ctw shows now and we can iterate).

Here's the updated look. I think this addresses your concerns, let me know.

https://www.loom.com/share/911ee783fb3246bdb14ce00ea09c7ca9

I also made it so the prev/next page buttons and current page button in pagination don't show a hover state or mouse pointer when they are "disabled" or "active".

eric-gunther (approved):
Everything works great for me locally.

PR #330, opened 2022-12-13
CTW-615 Builder id everywhere

melocule:
CTW-615
Adds the builder ID to all requests to Zus services in the `Zus-Account` header when available. Excludes auth requests.

image


luthfib:
The headers should also be... `Zus-Account`

luthfib (approved):

PR #333, opened 2022-12-14
Ctw 151/auto complete medications

luthfib:
This pr:
- Adds new medications dropdown
- Renames dosage to instructions 
- Remove note field as it is not used

image

image


pmadria3 (approved):
LGTM

PR #335, opened 2022-12-14
CTW-621/Add to Record button on other provider meds

mrosata:
Created an "Add to Record" button in the hamburger menu of `<OtherProviderMedsTable/>`.

Added to storybook (based on conditions mocked cache that @WesleyKapow did) so when someone adds an "other provider" med to their record in storybook, that change is updated in their provider meds table. Note: you will only see this happen when looking at the  `<PatientMedications />` story as the `<OtherProviderMedsTable />` story doesn't show the `<ProviderMedsTable />` and the cache doesn't persist story to story.

### DIFF
--- a/.changeset/fuzzy-seahorses-shop.md
+++ b/.changeset/fuzzy-seahorses-shop.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Created an "Add to Record" button in the hamburger menu of <OtherProviderMedsTable/>
--- a/src/components/content/forms/medications.tsx
+++ b/src/components/content/forms/medications.tsx
@@ -124,7 +124,7 @@ export const getMedicationFormData = (
       <MedicationsAutoComplete
         readonly={readonly}
         {...inputProps}
-        defaultCoding={{}}
+        defaultCoding={medication.rxNormCoding ?? {}}
       />
     ),
   },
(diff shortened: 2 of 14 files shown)


luthfib (changes requested):
Currently this does not work with the autocomplete medications

mrosata:
> Currently this does not work with the autocomplete medications, lmk if you need any help setting that up.

Yeah if you have a quick solution. I just merged main into this branch and it doesn't work with the new autocomplete medication endpoint. It also seems that errors on that autocomplete field never clear. If I do "Add to Record" and submit the medication field is an error (because it's just the text and not selected med), then after selecting a med though, the error state remains (even though the form validates and can submit successfully).
Screen Shot 2022-12-16 at 1 02 27 PM

luthfib:
Seeing this when trying to save. Not sure what we want to do about date asserted. Its locked

image

luthfib:
Maybe don't lock date asserted in add from other medications?

mrosata:
> Maybe don't lock date asserted in add from other medications?

yeah that's a tough one.. if there's no date asserted I suppose just use the current date or fallback to prescribed date? @luthfib @eric-gunther

luthfib:
What if we made the field not locked, that way they practitioner can see the date and change it they field it is necessary. We still fill it in with a default but they have an option of filling it in?

luthfib:
Actually ignore what I said

luthfib (approved):

PR #337, opened 2022-12-15
Ctw 623/fix builder impoersonation issue

luthfib:
Fix issue where zus-account header was still being passed even when logged in normal user.

### DIFF
--- a/src/components/content/forms/medications-autocomplete.tsx
+++ b/src/components/content/forms/medications-autocomplete.tsx
@@ -23,11 +23,7 @@ export const MedicationsAutoComplete = ({
   const handleSearchChange = async (query: string) => {
     setIsLoading(true);
     const requestContext = await getRequestContext();
-    const medications = await getAutoCompleteMedications(
-      requestContext,
-      requestContext.env,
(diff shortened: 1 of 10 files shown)


src/components/content/forms/medications-autocomplete.tsx
```diff
@@ -22,8 +22,12 @@ export const MedicationsAutoComplete = ({
 
   const handleSearchChange = async (query: string) => {
     setIsLoading(true);
-    const { authToken, env } = await getRequestContext();
-    const medications = await getAutoCompleteMedications(authToken, env, query);
+    const requestContext = await getRequestContext();
+    const medications = await getAutoCompleteMedications(
+      requestContext,
+      requestContext.env,
```
pmadria3, line 28: do we need to pass both? Maybe we can just use the .env inside the function instead of passing as param. Seems redundant to pass requestContext and requestContext.env.

luthfib: Agreed

pmadria3 (approved):

PR #339, opened 2022-12-15
CTW-499/create medication transaction with provenance

mrosata:
When creating a med now a Bundle transaction will be used so that a Provenance may also be used.

### DIFF
--- a/src/fhir/action-helper.ts
+++ b/src/fhir/action-helper.ts
@@ -1,8 +1,16 @@
-import { Resource } from "fhir/r4";
+import { FhirResource, Resource } from "fhir/r4";
+import { v4 as uuidv4 } from "uuid";
 import { omitEmptyArrays } from "./client";
 import { isFhirError } from "./errors";
-import { createProvenance } from "./provenance";
+import {
+  ASSEMBLER_CODING,
+  CREATE_CODING,
+  createProvenance,
+  UPDATE_CODING,
+} from "./provenance";
 import { CTWRequestContext } from "@/components/core/ctw-context";
+import { getUsersPractitionerReference } from "@/fhir/practitioner";
+import { claimsBuilderName } from "@/utils/auth";
 
 export async function createOrEditFhirResource(
   resource: Resource,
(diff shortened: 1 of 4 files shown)


mrosata:
Suggest that we update createFhirResource function to work the same way, currently that creates a provenance as a separate request https://github.com/zeus-health/ctw-component-library/blob/f0cf9adc3aefc7c774f1e214fc09df8ab4bd6548/src/fhir/action-helper.ts#L7-L37

luthfib:
> Suggest that we update createFhirResource function to work the same way, currently that creates a provenance as a separate request
> 
> https://github.com/zeus-health/ctw-component-library/blob/f0cf9adc3aefc7c774f1e214fc09df8ab4bd6548/src/fhir/action-helper.ts#L7-L37

That sounds good.  I think currently this would cause the whole transaction to fail for a care team user based on permissions, but they are going to to change that soon.

src/fhir/action-helper.ts
```diff
@@ -36,6 +44,69 @@ export async function createOrEditFhirResource(
...
 }
 
+export async function createOrEditFhirResourceWithProvenance(
+  resource: FhirResource,
+  requestContext: CTWRequestContext
+) {
+  const { fhirClient } = requestContext;
+  const builderName = claimsBuilderName(requestContext.authToken);
+  const type = resource.id ? "UPDATE" : "CREATE";
+  const resourceId = resource.id || uuidv4();
```
luthfib, line 54: Why do we do the `uuidv4`?

src/fhir/action-helper.ts
```diff
@@ -36,6 +44,69 @@ export async function createOrEditFhirResource(
   }
 }
 
+export async function createOrEditFhirResourceWithProvenance(
+  resource: FhirResource,
```
luthfib, line 48: Should we wrap this in a try/catch?

src/fhir/action-helper.ts
```diff
@@ -36,6 +44,69 @@ export async function createOrEditFhirResource(
...
+  const resourceId = resource.id || uuidv4();
+  const provenanceFullUrl = uuidv4();
+
+  const bundle: fhir4.Bundle = {
+    resourceType: "Bundle",
+    type: "transaction",
+    entry: [
+      {
+        request: {
+          method: type === "CREATE" ? "POST" : "PATCH",
```
luthfib, line 63: Not sure if this is correct.... pretty sure for FHIR, or at least the way we do `update` this would still need to be a POST. For a patch you would need to do fhir json patch, which we dont do.

src/fhir/action-helper.ts
mrosata: The medication will need some identifier so the Provenance in the Bundle can reference it, this is just setting a random UUID to act as that temporary relative reference. If there is a nuance to bundle transactions (or a better way) to do this reference we can update it.

src/fhir/action-helper.ts
mrosata: yeah let me dig into where the best place to put that would be. Feels too granular to do the try/catch in that function. I'm not 100% sure how the DrawerForm handles errors, will have to peek at it

src/fhir/action-helper.ts
mrosata: Patch probably is wrong, it might need to be a PUT. What makes you think it would be a POST out of curiousity?

A solution for now could be to rename the function `createFhirResourceWithProvenance` and punt on the update part since atm it only gets called on create. I'm not sure that we actually "edit/update" medications anyways.

luthfib: Actually it might be a PUT, I forget. I just remember it not being a PATCH

luthfib: PUT makes more sense

src/fhir/action-helper.ts
luthfib: I guess technically we have a try/catch around actions now, so it doesnt necessarily need a try catch unless we want to throw specific errors here

mrosata: Looks like the form handles the try/catch already
https://github.com/zeus-health/ctw-component-library/blob/c0b9e6e898d7a48144104cb552352369d10ec97d/src/components/core/form/drawer-form.tsx#L79-L88

Form (initiator) submits -> calls action -> action calls `createOrEditFhirResource` -> throw ERROR to action -> throw ERROR to form -> Form (initiator) catches.

This feels like the correct pattern to have the initiator/caller of the request function catch any errors than always having the function making the API call never throw.

luthfib: Sounds good, I agree

src/fhir/action-helper.ts
luthfib: Actually I remember why we did provenance after the request. It was so that we could reference the condition that was just updated. This might actually do the same, just test and make sure that the resources are still linking correctly.

src/fhir/action-helper.ts
mrosata: I updated the function (and function name) to only handle creation and if we need updates down the road we can modify the function

src/fhir/action-helper.ts
mrosata: yeah the created provenance will reference the id of the newly created fhir resource (medication in this case).

luthfib (approved):

luthfib:
Actually can you add a changeset?

PR #347, opened 2022-12-19
Adding UPID to CPR query

eric-gunther:
When retrieving the history for a given medication we look up a list of medicationX resources in ODS by ID.

In order to trigger "CPR mode" in ODS, where we surface data across builders, we must search by UPID. This was thought to be a redundant/unnecessary filter, but it is actually required in order to access the data. It's also worth noting that the CPR data is also only available via a search (ie. using the `_id` query parameter) and is not accessible directly by ID.

This does not impact conditions as the condition history fetch is already passing a UPID (see line 122 of `conditions.ts`).

### DIFF
--- a/.changeset/tall-islands-tickle.md
+++ b/.changeset/tall-islands-tickle.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Adding patient UPID to queries for medication data that may span builders, this triggers "CPR mode" in ODS.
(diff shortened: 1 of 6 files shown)


luthfib (dismissed):
LGTM. Im thinking we are going to need to propagate this to most of our other searches throughout the app as well at some point, no?

luthfib:
Looks like there are some validation error, seem to be type related

luthfib (dismissed):

luthfib (approved):

PR #348, opened 2022-12-20
Ctw 589/conditions 2.0 hover

luthfib:
Add hover actions in conditons 2.0.

image


src/components/core/table/table.scss
```diff
@@ -14,6 +14,10 @@
   );
 }
 
+.ctw-action-hover {
+  background: linear-gradient(90deg, rgba(249, 250, 251, 0) 0%, #f9fafb 4.17%);
```
mrosata, line 18: this should be equivalent if we don't want to mix rgba and hex values
```suggestion
  background: linear-gradient(90deg, #f9fafb00 0%, #f9fafb 4.17%);
```

src/components/content/conditions/patient-conditions-history.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const shouldHistoryDrawerBeReadOnly = ({
+  data,
+  patientRecords,
+  handleEditCondition,
+}: HistoryDrawerReadOnlyCheck) => {
+  if (data && patientRecords.some((record) => record.id === data.id)) {
+    return () => handleEditCondition(data);
+  }
+  return undefined;
+};
```
mrosata, line 19: I can't think of a name off the top of my head, but if I were reading the code that used `shouldHistoryDrawerBeReadOnly` the expectation would be that it returns a boolean. Maybe something like `createHandleEditConditionFn` _(probably something more descriptive if you can come up with one, my name idea isn't the greatest either)_

src/components/core/table/table-rows.tsx
```diff
@@ -43,11 +45,23 @@ export const TableRows = <T extends MinRecordItem>({
     <>
       {records.map((record) => (
         <tr
-          className={cx({
-            "ctw-cursor-pointer": typeof handleRowClick === "function",
-          })}
+          className={cx(
+            "ctw-group ctw-relative ctw-z-10 hover:ctw-bg-bg-lighter",
+            {
+              "ctw-cursor-pointer": typeof handleRowClick === "function",
```
mrosata, line 51: This is a non-blocking suggestion (because it would also require an import statement);
In the file above we're using `isFunction`, then here `typeof handleRowClick === "function"`. If there ends up being something else in this PR you need to manually change it would be good to also update here
```suggestion
              "ctw-cursor-pointer": isFunction(handleRowClick),
```

src/components/core/table/table-rows.tsx
```diff
@@ -43,11 +45,23 @@ export const TableRows = <T extends MinRecordItem>({
...
+            }
+          )}
           key={record.id}
-          onClick={() => {
+          onClick={(event) => {
+            const { target } = event;
+            // This is for the use case where we do not want have the onRowClick called in areas near the button as that will cause confusion to the user
```
mrosata, line 58:
```suggestion
          onClick={({ target }) => {
            // This is for the case where a user clicks area near the button (but not on), we do not want have the onRowClick handler trigger as that will cause confusion to the user
```

mrosata (approved):

PR #353, opened 2022-12-22
Ctw 590/update edit add conditions panel

luthfib:
Update edit/add conditions panel to automatically fill in verification and clinical status based on the status value in the dropdown.

image

### DIFF
--- a/src/components/content/forms/schemas/condition-schema.tsx
+++ b/src/components/content/forms/schemas/condition-schema.tsx
@@ -50,7 +50,7 @@ const sharedFields = (condition: ConditionModel) => [
   },
   {
     label: "Status",
-    value: condition.status,
+    value: condition.status === "Unknown" ? "Select One" : condition.status,
     field: "status",
   },
 
--- a/src/components/content/forms/schemas/condition-schema.tsx
+++ b/src/components/content/forms/schemas/condition-schema.tsx
@@ -49,15 +49,11 @@ const sharedFields = (condition: ConditionModel) => [
     hidden: true,
   },
   {
-    label: "Clinical Status",
-    value: levelTwoToOneMapping(condition.clinicalStatus),
-    field: "clinicalStatus",
+    label: "Status",
+    value: condition.status,
+    field: "status",
   },
-  {
-    label: "Verification Status",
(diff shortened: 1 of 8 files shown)


src/components/content/forms/schemas/condition-schema.tsx
```diff
@@ -81,7 +77,13 @@ const conditionSchema = z.object({
   subjectID: z.string({
     required_error: "Condition subjectID must be specified.",
   }),
-  clinicalStatus: z.enum(["active", "inactive"]),
```
pmadria3, line 84: are we adding back the other options? In an earlier ticket we removed them so just want to confirm.

luthfib: Which other options? We removed verification and clinical status and status now has `Active, Pending, Inactive, Refuted, Entered in Error}`

src/components/content/forms/schemas/condition-schema.tsx
```diff
@@ -49,15 +49,11 @@ const sharedFields = (condition: ConditionModel) => [
...
-    label: "Clinical Status",
-    value: levelTwoToOneMapping(condition.clinicalStatus),
-    field: "clinicalStatus",
-  },
-  {
-    label: "Verification Status",
-    value: levelTwoToOneMapping(condition.verificationStatus),
-    field: "verificationStatus",
+    label: "Status",
+    value: condition.status === "Unknown" ? "Select One" : condition.status,
```
pmadria3, line 53: could status be Unknown?

pmadria3 (commented):
Looks good - left a couple comments

src/components/content/forms/schemas/condition-schema.tsx
luthfib: Yes it could. If it is not defined.

src/components/content/forms/schemas/condition-schema.tsx
pmadria3: Oh right we don't have verification anymore - took a peek at your ticket

pmadria3 (approved):
LGTM

PR #359, opened 2022-12-23
Fix pdf issue by removing it - part 2

pmadria3:
Removing the PDF viewer as there is no need for it and wasn't being used anyway.


.changeset/tricky-eyes-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Remove pdf libraries to fix up an import issue.
```
luthfib, line 5:
```suggestion
Remove `react-pdf-viewer` and `react-json-view` to fix server rendering issues. 
```

luthfib (dismissed):

src/components/content/CCDA/base64toBinaryField.tsx
```diff
@@ -1,111 +1,28 @@
...
-            useSelfClosingElement: true,
-          }
-        )}
-      />
-    );
-  }
-
-  if (contentType.includes("/pdf")) {
-    const blob = base64toBlob(data);
-    const url = URL.createObjectURL(blob);
```
mrosata, line 55: so we need to remove the parts that don't use ReactJsonViewer? Or do these also use view Viewer?

pmadria3: yeah so none of this was being used at all so removed it all

luthfib: We werent actually using any of that code

luthfib (approved):

mrosata:
@pmadria3, @luthfib We could just write our own JsonViewer. It's just rendering a table of nested data correct?

```tsx
import { isArray, isObject, keys } from "lodash";

type ZusJsonViewerDataCellProps =  {
  value: unknown|Record<string,unknown>;
}

type ZusJsonViewerProps = {
  data: Record<string, ZusJsonViewerDataCellProps>;
}

const ZusJsonViewerDataCell = ({ value }: ZusJsonViewerDataCellProps) => {
  if (isObject(value)) {
    return <ZusJsonViewer data={value as Record<string, unknown>} />
  }
  return <>{value}</>;
}

const ZusJsonViewer = ({ data }: ZusJsonViewerProps) => {
  return (
    <div className="ctw-table-container">
      <table className="ctw-json-viewer-table">
        <tbody>
        {keys(data).map((key) => (
          <tr key={key}>
            {!Array.isArray(data) &&
              <td className="ctw-font-bold ctw-text-capitalize">
                {key}
              </td>}
            <td>
              <ZusJsonViewerDataCell value={data[key]} />
            </td>
          </tr>
        ))}
        </tbody>
      </table>
    </div>
  )
}
```

luthfib:
Yeah, we probably could, but at the moment I dont think we have a need for it.

PR #360, opened 2022-12-23
CTW-638/Add To Record Date and MedicationCodeableConcept text

mrosata:
When adding medication to record from the other provider table:
 - Asserted Date will now be set as the current date rather than the original medication statement date asserted.
 - The display text is now also properly set on the medication resources `medicationCodeableConcept` value.
 - When using "Add to Record" button from Storybook, the mocked response is now intercepting `POST /fhir` rather than `/fhir/MedicationStatement` to reflect that medication statements from this form are now created in a Bundle-transaction.

### DIFF
--- a/src/components/content/medications/story-helpers/mocks/requests.ts
+++ b/src/components/content/medications/story-helpers/mocks/requests.ts
@@ -1,4 +1,5 @@
-import { cloneDeep } from "lodash";
+import { MedicationStatement } from "fhir/r4";
+import { cloneDeep, find } from "lodash/fp";
 import { rest } from "msw";
 import { ComponentType, createElement } from "react";
 import { v4 as uuidv4 } from "uuid";
@@ -53,18 +54,30 @@ const mockMedicationStatementGet = rest.get(
   }
 );
 
-// Mocked post will add the new medication
+// Mocked post will add the new medication to patientProviderMedsCache
 const mockMedicationStatementPost = rest.post(
-  "https://api.dev.zusapi.com/fhir/MedicationStatement",
+  "https://api.dev.zusapi.com/fhir",
   async (req, res, ctx) => {
-    const newMedication = await req.json();
+    const delayMs = 500;
+    const findMedStatementInBundleFn = find({
+      resource: { resourceType: "MedicationStatement" },
+    });
+    const bundle = await req.json();
+    const newMedication = findMedStatementInBundleFn(bundle.entry)?.resource as
+      | MedicationStatement
+      | undefined;
+    if (!newMedication) {
+      return res(ctx.status(400));
+    }
+
     newMedication.id = uuidv4();
     patientProviderMedsCache.entry?.push({
       resource: newMedication,
       search: { mode: "match" },
     });
     patientProviderMedsCache.total = patientProviderMedsCache.entry?.length;
-    return res(ctx.status(200), ctx.json(newMedication));
+
+    return res(ctx.delay(delayMs), ctx.status(200), ctx.json(newMedication));
   }
 );
 
--- a/src/components/content/medications/story-helpers/mocks/requests.ts
+++ b/src/components/content/medications/story-helpers/mocks/requests.ts
@@ -58,7 +58,6 @@ const mockMedicationStatementGet = rest.get(
 const mockMedicationStatementPost = rest.post(
   "https://api.dev.zusapi.com/fhir",
   async (req, res, ctx) => {
-    const delayMs = 500;
     const findMedStatementInBundleFn = find({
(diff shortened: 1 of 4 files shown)


src/components/content/medications/story-helpers/mocks/requests.ts
```diff
@@ -53,18 +54,30 @@ const mockMedicationStatementGet = rest.get(
...
+      resource: { resourceType: "MedicationStatement" },
+    });
+    const bundle = await req.json();
+    const newMedication = findMedStatementInBundleFn(bundle.entry)?.resource as
+      | MedicationStatement
+      | undefined;
+    if (!newMedication) {
+      return res(ctx.status(400));
+    }
+
```
mrosata, line 72: Here we're just finding the medication statement in a Bundle and returning 400 if it's not there (because the request is bad).

src/components/content/medications/story-helpers/mocks/requests.ts
```diff
@@ -53,18 +54,29 @@ const mockMedicationStatementGet = rest.get(
...
     newMedication.id = uuidv4();
     patientProviderMedsCache.entry?.push({
       resource: newMedication,
       search: { mode: "match" },
     });
     patientProviderMedsCache.total = patientProviderMedsCache.entry?.length;
-    return res(ctx.status(200), ctx.json(newMedication));
+
+    return res(ctx.delay(500), ctx.status(200), ctx.json(newMedication));
```
mrosata, line 79: Delay response by 500ms to be a little more realistic. It also makes it where when the user hits "save", the medication doesn't instantly move from bottom table (other provider records) to top table (builder records) because the re-render appears to trigger off this response.

src/fhir/models/medication-statement.ts
```diff
@@ -115,6 +111,21 @@ export class MedicationStatementModel extends FHIRModel<fhir4.MedicationStatemen
...
+  get rxNormCodeableConcept() {
+    const coding = getIdentifyingRxNormCoding(
+      this.resource,
+      this.includedResources
+    );
+
+    return {
+      ...(coding ?? {}),
+      display: coding?.display ?? this.display,
+    };
```
mrosata, line 126: Here we're just returning the RxNorm codeable concept.. however, if the display isn't set we set it to the `MedicationStatementModel.display` value. The form to add a med uses this display attribute as the value for `resource.medicationCodeableConcept.text`.

> It does seem to always be the case _(in every med I've seen)_ that `display` is not set in the codeable concept... but it's valid on a `fhir4.Code` so we might as well support it.

src/fhir/models/medication-statement.ts
```diff
@@ -115,6 +111,21 @@ export class MedicationStatementModel extends FHIRModel<fhir4.MedicationStatemen
     return getIdentifyingRxNormCode(this.resource, this.includedResources);
   }
 
+  /**
+   * Get RxNorm coding with "display" defaulting to this Med-Statement label.
+   */
+  get rxNormCodeableConcept() {
+    const coding = getIdentifyingRxNormCoding(
```
pmadria3, line 118: I see you have this function calling the `getMedicationCodeableConcept` -- there's also a `codeableConceptLabel` you could use potentially?

pmadria3: Sorry for coding you could maybe use: `findCoding`

pmadria3 (approved):
small comment - otherwise LGTM

src/fhir/models/medication-statement.ts
mrosata: The `codeableConceptLabel` and `getMedicationCodeableConcept` are both being called by getting `this.display` https://github.com/zeus-health/ctw-component-library/blob/41c45e53a36dc1def522aec7306395279139721a/src/fhir/models/medication-statement.ts#L68-L72
That handles getting the complete set of medicationCodeableConcept(s) and finding a display/text with fallback.
https://github.com/zeus-health/ctw-component-library/blob/41c45e53a36dc1def522aec7306395279139721a/src/fhir/codeable-concept.ts#L13-L17

The `getIdentifyingRxNormCoding` also calls `getMedicationCodeableConcept`, but then it searches through all the codeable concepts to the one that is rxNorm without any extensions (or at least not without enriched "ActiveIngredient" or "BrandName" extensions.

`findCoding` would presumably work here to find the rxNorm code. I think the reason `getIdentifyingRxNormCodeing` makes more sense is that there must be times when there are multiple rxNorm codes (but some are different code values because they are enriched to show active ingredient or brand name rxnorm codes. That's what I was able to gather anyways. There are a handful of similar functions :)

PR #365, opened 2022-12-27
CTW-627/Dismiss Medications

mrosata:
This PR makes it so other provider medications can be dismissed. There isn't a mechanism to retain a medication that has been dismissed, just a one way street. However, the `useQueryAllPatientMedications` query hook will now return at third collection of meds, `dismissedOtherProviderMedications` in addition to the current two _(`builderMedications`, `otherProviderMedications`)_

Here's a loom.

https://www.loom.com/share/cf001c87f9ca49379025f7fa4414b493

---
> Additionally. It seemed we used to have a `MedicationStatementModel.clinicalStatus` which was orphaned. I added it back in. It's used in the form as the capitalized status doesn't work anymore on the add new medication form (statuses on the form must be lowercase)


mrosata:
It maybe worth noting that in a future PR the action menu with "Add to Record" and "Dismiss" may become buttons instead of a dropdown hamburg menu. For this iteration though we just need to capability to dismiss the records.

mrosata:
https://www.loom.com/share/4b7e38bf5dc7469c93d87dc913a242f2

eric-gunther:
@mrosata see [this Slack thread](https://zushealth.slack.com/archives/C03M2KKTXPZ/p1672329825739469) for explanation as to why I just pushed 2 changes:

1. Cleaned up terminology around "status". Now both condition and medication-statement models have a "displayStatus" that is used to calculate a user friendly status value.
2. Updated to use "archive" as the profile action. This is to maintain consistency with conditions. It looks like we've entered a state where "archive" wins out as the technical term and "dismiss" is the UX term that is subject to further iteration.

eric-gunther (approved):
I tested locally against both conditions and meds (after making my changes) and it all works. We should probably have a storybook test for this but I won't block approval on that, especially since we're following up on this quickly with a larger refactoring of the medications component UX (ie. the tabbed layout).

Since this affects conditions I'd like one of those folks to eyeball it before merging.

src/components/content/medication-drawer.tsx
```diff
@@ -55,7 +57,19 @@ export const MedicationDrawer = ({
         </div>
       </Drawer.Body>
       <Drawer.Footer>
-        <div className="ctw-flex ctw-justify-end">
+        <div className="ctw-flex ctw-justify-end ctw-space-x-2">
+          {isFunction(onDismissal) && (
```
WesleyKapow, line 61: This can just be `onDismissal &&` as the type is `(m: MedicationStatementModel) => void | undefined` .. so if it is defined, it will always be a function.

src/components/content/medication-drawer.tsx
```diff
@@ -55,7 +57,19 @@ export const MedicationDrawer = ({
...
+            <button
+              type="button"
+              className="ctw-btn-primary"
+              onClick={() => {
+                onDismissal(medication);
+                drawerProps.onClose();
+              }}
+            >
+              Dismiss
```
WesleyKapow, line 70: May want to go with `archive` to align with conditions 2.0

src/components/content/medications/medication-actions.tsx
```diff
@@ -0,0 +1,23 @@
+import type { MedicationStatementModel } from "@/fhir/models";
+import { CTWRequestContext } from "@/components/core/ctw-context";
+import { recordProfileAction } from "@/fhir/basic";
+import { QUERY_KEY_OTHER_PROVIDER_MEDICATIONS } from "@/utils/query-keys";
+import { queryClient } from "@/utils/request";
+
+export const handleMedicationDismissal = async (
```
WesleyKapow, line 7: lgtm here, just a note that we'll probably want to DRY this up with the conditions one.. they are basically the same thing.

src/fhir/bundle.ts
```diff
@@ -68,7 +68,7 @@ export function getIncludedResources(bundle: FhirResource): ResourceMap {
 // of associated basic resources.
 // This way we can easily lookup a specific resources associated basic resources.
 export function getIncludedBasics(bundle: FhirResource): ResourceArrayMap {
-  const resources = {} as ResourceArrayMap;
+  const resources = new Map() as ResourceArrayMap;
```
WesleyKapow, line 71: Curious why this change?

src/components/content/medication-drawer.tsx
```diff
@@ -55,7 +57,19 @@ export const MedicationDrawer = ({
         </div>
       </Drawer.Body>
       <Drawer.Footer>
-        <div className="ctw-flex ctw-justify-end">
+        <div className="ctw-flex ctw-justify-end ctw-space-x-2">
```
WesleyKapow, line 60: OH! ya'll have two ways to dismiss it? In the drawer and in the "..." menu?

WesleyKapow (approved):
LGTM, 1 or two minor nitpicks and then some comments/questions.

src/components/content/medication-drawer.tsx
WesleyKapow: NVM, sounds like "dismiss" is the new lingo.

src/components/content/medication-drawer.tsx
mrosata: yeah, I wonder if we'll want to show a popup dialog eventually for the one in the drawer footer _(or both)_. I'm interested to see if people click "dismiss" thinking it's to dismiss the side panel drawer and not the med.

We don't have an undo or "un-dismiss" yet either.

eric-gunther (approved):

src/components/content/medication-drawer.tsx
WesleyKapow: Eh, we opted to not have a confirmation since we can "un-dismiss".. I assume meds will have un-dismiss too for consistency.

src/components/content/medication-drawer.tsx
mrosata: just merged, but adding this change in the branch for my next pr.

src/fhir/bundle.ts
mrosata: Object works fine here, but Map feels more appropriate and is built for these kinds of operations. Figured it won't make much difference here, but it'll codify our use of Map over Object in the future if a dev reuses this code. Mozilla has a good table on the difference between Map and Objecthttps://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps

src/components/content/medication-drawer.tsx
mrosata: I brought it up to UX so it'll be on their radar

src/components/content/medication-drawer.tsx
mrosata: Weird, typescript behaves differently with `onDismissal &&` vs `isFunction(onDismissal)`. Will look deeper in am

PR #369, opened 2022-12-29
Conditions2 refactors

WesleyKapow:
Refactored a bunch of conditions 2.0 code to take advantage of some new providers. This allows us to use hooks and functions to open condition form drawers, conditions history, and delete confirmation modal.

New providers:
* `ModalProvider`
* `DrawerProvider`

Bug fixes / changes:
* Hover actions now respect `readOnly` and won't show up when `true`.
* Condition history now respects `readOnly` and won't show the **Update** button when set to `true`.
* Hide **Remove** action for already "deleted" conditions (when status = `entered-in-error`).
* Default new conditions status to `Active`.
* Fix bug when editing entered-in-error conditions where status form element wasn't getting pre-populated correctly.
* Re-add line clamping for notes data in the table (clamp to 3 lines).
* Interestingly, this refactor allowed cleaning up the timeout used for closing history drawer and then opening the edit form drawer.


src/components/content/condition-history/conditions-history.tsx
```diff
@@ -147,19 +147,14 @@ export function ConditionHistory({
...
-              onEdit={() => {
-                onClose();
-                // TODO: Clean this up when headless ui supports multiple drawers  https://github.com/tailwindlabs/headlessui/discussions/1564.
-                // We setTimeout here because we need to wait till condition history drawer closes.
-                // This fixes a bug where having multiple drawers causes headless ui useScrollLock to become out of sync, which causes overlay: hidden incorrectly persist on the html element.
-                setTimeout(onEdit, 400);
-              }}
-            />
-          )}
+          <Details
```
WesleyKapow, line 150: Note: We now always show `Details` but details will only show the **Update** button if we have an `onEdit` function.

src/components/content/conditions-helper.ts
```diff
@@ -38,3 +43,24 @@ export const onConditionDelete = async (
 
   await queryClient.invalidateQueries([QUERY_KEY_PATIENT_CONDITIONS]);
 };
+
+export const toggleArchive = async (
```
WesleyKapow, line 47: Refactored this out into a helper (used by 2 places). May want to refactor further to DRY things up with meds as well.

src/components/core/modal-confirm-delete.tsx
```diff
@@ -34,17 +34,19 @@ export const ModalConfirmDelete = ({
...
       <div className="ctw-items-left ctw-flex ctw-h-full ctw-flex-col ctw-space-y-2 ctw-overflow-y-auto">
         <span className="ctw-text-left ctw-text-lg ctw-font-medium ctw-text-content-black">
-          {`Remove this ${resource.resourceType.toLowerCase()}?`}
+          {`Remove this ${resourceType}?`}
         </span>
         <span className="ctw-subtext ctw-max-w-sm ctw-text-left ctw-text-content-light">
           This will remove{" "}
           <span className="ctw-font-medium">{resourceName}</span> from this
-          patient&apos;s condition list.
+          patient&apos;s {resourceType} list.
```
WesleyKapow, line 49: Generic component was previously hardcoded with `condition`.

src/components/core/table/table-rows.tsx
```diff
@@ -1,18 +1,18 @@
...
 
-type TableRowsProps<T extends MinRecordItem> = {
+export type TableRowsProps<T extends MinRecordItem> = {
   records: T[];
   columns: TableColumn<T>[];
   isLoading: boolean;
   emptyMessage: string | ReactElement;
   handleRowClick?: (record: T) => void;
-  rowActions?: (record: T) => JSX.Element;
+  RowActions?: ComponentType<{ record: T }>;
```
WesleyKapow, line 15: This was tricky! Switching to a real component here helped fix the following errors:
> Rendered fewer hooks than expected. This may be caused by an accidental early return statement.

> Rendered more hooks than during the previous render.

When # of rows changed (add/removes), we'd end up with a different set of row actions. These row actions now have their own state/hooks (via things like `useAddConditionForm` which in turn call other hooks). The problem was that by using a function, these hooks were getting owned by the `TableRows` component. This meant that different render calls could equate to a different set of hooks for the `TableRows` component which is a big NO NO in react. This is the same reason why you have to define all of your hooks at the top before any conditional returns. I.e. the # of hooks has to be the same and maintain the same order across all render calls.

By switching `RowActions` over to be a real component, react does the right thing and associates those hooks with the `RowAction` component and not `TableRows`. So new row = new `RowAction` component which is of course allowed to have its own hooks.

luthfib (approved):

luthfib:
Just missing a changeset

luthfib (approved):

PR #371, opened 2022-12-30
Upgrade vite and fix providers

WesleyKapow:
The main purpose here is to fix a HMR error by splitting context out of the provider files.

Also:
* Upgraded vite from 3.* to 4.0
* Moved CTW & Patient providers into the `core/providers` directory.


luthfib (approved):
LGTM, just missing a changeset

PR #372, opened 2022-12-30
Ctw 570/support audit trail

luthfib:
Support audit trail in condition history. Only fetch binary document when clicking on the source document button.

image


src/components/content/conditions/story-helpers/mocks/provencanceCondition.ts
```diff
@@ -0,0 +1,674 @@
+export const ProvenanceCondition = {
```
WesleyKapow, line 1: File names should be kebab-case (lot of the CCDA files aren't because they were copied over... fine for now I guess).

src/components/content/conditions/story-helpers/mocks/requests.ts
```diff
@@ -157,3 +160,44 @@ const mockConditionsBasic = rest.post(
...
 );
+
+const mockConditionProvenance = rest.get(
+  "https://api.dev.zusapi.com/fhir/Provenance?target=Condition/:Condition",
+  async (req, res, ctx) => res(ctx.status(200), ctx.json(ProvenanceCondition))
+);
+
+// Mocked post will add the new condition to the
+// patient conditions bundle.
+const mockConditionBundle = rest.post(
```
WesleyKapow, line 171: Ah this is to handle the batch for conditions and history? Maybe spell that out a bit more in the name & comment? Comment says "will add the new condition", but I don't see any add happening.

Any change this can be DRY'd up with `mockConditionGet`?

src/fhir/conditions.ts
```diff
@@ -129,6 +172,7 @@ export function useConditionHistory(condition?: ConditionModel) {
           patientUPID: patient.UPID,
           _include: ["Condition:patient", "Condition:encounter"],
           "_include:iterate": "Patient:organization",
+          // _revinclude: "Provenance:target",
```
WesleyKapow, line 175: Remove?

src/fhir/conditions.ts
```diff
@@ -148,7 +192,31 @@ export function useConditionHistory(condition?: ConditionModel) {
           searchParams
         );
 
-        return { conditions, bundle };
+        const versionHistoryBundle = (await getConditionVersionHistory(
+          requestContext,
+          searchParams
+        )) as fhir4.Bundle | undefined;
```
WesleyKapow, line 198: Shouldn't need the `as` here. Have `getConditonVersionHistory` be explicit of it's return type of `Promise<fhir4.Bundle | undefined>`

WesleyKapow (changes requested):
LGTM, few minor questions and comments.

WesleyKapow (approved):
LGTM

PR #377, opened 2023-01-03
Use basic.code instead of meta.tag for archive/unarchive

WesleyKapow:
This fixes two issues:
1. There could be permission issues around updating meta tags.
2. Meta tags are applied to all versions of the resource. Meaning if we were to archive and then unarchive, we'd lose the history of the archive action.


mrosata (approved):
lgtm

luthfib:
Checking one behavior, before approving

PR #381, opened 2023-01-03
Bugfix/history details not showing

luthfib:
Fix bug where details are not showing when there are not history records present.

image


src/components/content/condition-history/conditions-history.tsx
```diff
@@ -132,86 +132,101 @@ export function ConditionHistory({
...
-          <ConditionHeader condition={condition} />
-          <Details
-            data={conditionData(condition)}
-            readOnly={!onEdit}
-            onEdit={() => {
-              onClose();
-              onEdit?.();
-            }}
-          />
+export type HistoryRecordsProps = {
```
WesleyKapow, line 168: No need to export this as the sub component isn't exported.

WesleyKapow (approved):
Awesome, just one minor nit.

WesleyKapow (approved):

PR #383, opened 2023-01-04
Added storybook tests for <PatientMedications />

mrosata:
This PR has 3 tests:
 - Adding meds from "other provider" records to "provider" records.
 - Adding a brand new medication.
 - Canceling adding a med after starting to fill out the form.

Additionally: The react testing library knows about `data-testid` attributes for selecting elements. This PR adds the name of form fields into test IDs on those fields. The need for this came up for just a single element, but as it worked well I just converted some `.getByLabelText("..")` methods to instead use `.getByTestId("form-field-<name>")`


src/components/content/medications/story-helpers/mocks/forms-data-terminology-dosages.ts
```diff
@@ -0,0 +1,241 @@
+export const searchDosagesALB = [
```
pmadria3, line 1: do we need all these medications or for the sake of mock just use a couple?

package.json
```diff
@@ -38,7 +38,8 @@
     "release": "changeset publish",
     "storybook": "npm run generate:theme-cjs && storybook dev -p 6006",
     "build-storybook": "npm run generate:theme-cjs && storybook build",
-    "test-storybook": "test-storybook"
+    "test-storybook": "test-storybook",
```
pmadria3, line 41: Nothing in package - but we probably want a changeset

pmadria3 (commented):
looks good to me and tests pass - just a couple comments

package.json
WesleyKapow: Changesets are for consumer facing changes. Anything we do that is internal (testing / dev stuff / etc) does not need to be in a changeset as users of the library don't need or care about it ;)

src/components/content/medications/story-helpers/mocks/forms-data-terminology-dosages.ts
mrosata: They are more or less for when users interact with the component so it feels a bit more real, I was going to take 2-3 meds for each letter of the alphabet but then it seemed too much work once I was in postman. I don't mind having less in there, but don't think there's an overly compelling reason not to have them there.

package.json
pmadria3: ah good to know

mrosata: Normally I wouldn't bump the version for just adding tests, but since I did modify a couple files with `data-test-id` it technically is a patch. I'll put a changeset in for that.

src/components/content/medications/story-helpers/mocks/forms-data-terminology-dosages.ts
pmadria3: That's fair enough

pmadria3 (approved):
LGTM

PR #390, opened 2023-01-05
Ctw 570/audit trail fixes

luthfib:
Change deduping logic in history drawer so that enriched records will now no longer cause dupes to show.

image


pmadria3 (approved):
LGTM, could you add screenshot just to verify?

PR #395, opened 2023-01-10
Query Invalidation Helpers

mrosata:
These helper functions can be used internally or externally. The need came up to be able to invalidate medication query caches from code that uses the ctw-component-library. It is possible that a builder app knows a medication has been modified or stopped via some mechanism that isn't a component library component, and thus wishes to helpfully clear a cache.
Rather than have consumers of the library know about cache keys and which query library we use, they can just call a helper like `invalidateOtherProviderMedsQueries()` or `invalidateMedicationHistoryQueries()`.

### DIFF
--- a/.changeset/strong-turkeys-impress.md
+++ b/.changeset/strong-turkeys-impress.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Added query invalidation helpers for both internal and external use
+Add query invalidation helpers for both internal and external use
--- a/.changeset/strong-turkeys-impress.md
+++ b/.changeset/strong-turkeys-impress.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Added query invalidation helpers for both internal and external use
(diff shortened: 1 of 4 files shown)


.changeset/strong-turkeys-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Added query invalidation helpers for both internal and external use
```
pmadria3, line 5:
```suggestion
Added query invalidation helpers for both internal and external use
```
Add query invalidation helpers for both internal and external use.

src/utils/invalidate-queries.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  QUERY_KEY_MEDICATION_HISTORY,
+  QUERY_KEY_OTHER_PROVIDER_MEDICATIONS,
+  QUERY_KEY_PATIENT_BUILDER_MEDICATIONS,
+} from "./query-keys";
+import { queryClient } from "@/utils/request";
+
+export const invalidateQueriesFrom = (queryKey: unknown[]) =>
```
pmadria3, line 8: do we want this type to be unknown?

pmadria3 (approved):
LGTM, left couple comments

src/utils/invalidate-queries.ts
mrosata: yeah that's how it's typed by the tanstack query library. Query keys can be made up of anything. We are using strings mostly though.

pmadria3: Gotcha! 👍

pmadria3 (approved):
LGTM

PR #398, opened 2023-01-11
CTW-594/update-pagination-2.0

pmadria3:
This PR updates pagination for all the embedded component tables. 

Screenshot 2023-01-11 at 12 24 08 PM

Screenshot 2023-01-11 at 12 24 22 PM

Original view for the table:
Screenshot 2023-01-11 at 12 23 37 PM

### DIFF
--- a/.changeset/polite-crews-argue.md
+++ b/.changeset/polite-crews-argue.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Fix text and button styling for pagination.
--- a/.changeset/polite-crews-argue.md
+++ b/.changeset/polite-crews-argue.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": minor
 ---
 
-Fix text and button styling for pagination.
+Update text and button styling for pagination.
--- a/src/components/core/table/table.scss
+++ b/src/components/core/table/table.scss
@@ -92,7 +92,11 @@
 
   .ctw-pagination {
     margin: 8px 0 !important;
-    flex-direction: column-reverse;
+    flex-direction: initial;
+    justify-content: space-between;
+    align-items: center;
+    padding-left: 1rem;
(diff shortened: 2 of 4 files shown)


luthfib (changes requested):
Looking good, however in quick view (Slightly smaller sizes, it is currently looks off... 

image

.changeset/polite-crews-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Fix text and button styling for pagination.
```
luthfib, line 5: We didn't really fix anything. We changed it to be different. This almost makes sound like a patch was done, which is not really the case.

src/components/core/table/table.scss
```diff
@@ -92,7 +92,11 @@
 
   .ctw-pagination {
     margin: 8px 0 !important;
-    flex-direction: column-reverse;
+    flex-direction: initial;
+    justify-content: space-between;
+    align-items: center;
+    padding-left: 1rem;
+    padding-right: 1rem;
     gap: 10px;
```
luthfib, line 100: Can we make these tailwind values? so we could do something like.. 

```
.ctw-pagination {
    @apply items-center etc; 
    margin: 8px 0 !important;
    
```

We shouldnt really add px or rem values whenever possible to stay consistent

src/components/core/table/table.scss
```diff
@@ -91,12 +91,9 @@
...
 
   .ctw-pagination {
+    @apply ctw-items-center ctw-justify-between ctw-pl-4 ctw-pr-4;
     margin: 8px 0 !important;
     flex-direction: initial;
-    justify-content: space-between;
-    align-items: center;
-    padding-left: 1rem;
-    padding-right: 1rem;
     gap: 10px;
```
luthfib, line 97: Same with these values

luthfib (approved):

PR #399, opened 2023-01-11
ActionList supports second action

mrosata:
- `<ActionList/>` supports a secondary action button having the same styling of the primary action.
- Added `useDismissMedication` hook which just creates a dismissMedication function.


eric-gunther (approved):
LGTM. Seems like at some point we could have an actions array instead of primary and secondary actions, but don't feel strongly enough to hold it up.

pmadria3 (approved):
LGTM

PR #402, opened 2023-01-12
Fix status mapping

luthfib:
Display status will now show the correct display status. Add tests to confirm a bunch of cases


src/fhir/models/condition.ts
```diff
@@ -277,31 +277,55 @@ export class ConditionModel extends FHIRModel<fhir4.Condition> {
         case "resolved":
           return "Inactive";
         default:
-          return "Unknown";
+          return "";
       }
     }
 
+    // What to show if patient record resource.
```
pmadria3, line 284: Not sure what this comment means

src/fhir/models/condition.ts
```diff
@@ -277,31 +277,55 @@ export class ConditionModel extends FHIRModel<fhir4.Condition> {
...
+    function byVerificationStatus(code: VerificationStatus | undefined) {
```
pmadria3, line 285: rename function?

pmadria3: maybe whichVerificationStatus? same with the clinicalStatus?

src/fhir/models/condition.ts
luthfib: Like if the record is not other provider records but patient owned resource

pmadria3 (approved):
Couple nits - otherwise LGTM

src/fhir/models/condition.ts
luthfib: I moved it

pmadria3: Gotcha - wasn't sure what the comment meant but makes sense

pmadria3 (approved):

mrosata (approved):

PR #405, opened 2023-01-12
Remove entered in error audit trail

luthfib:
Do not fetch condition history for conditions with verification status of entered-in-error. Also do not remove entered-in-error from conditions with entered-in-error from the audit trail.


pmadria3 (approved):
LGTM - can you add screenshot to PR?

PR #407, opened 2023-01-17
CTW-680/Telemetry and DataDog Logging

mrosata:
CTW-680

## Telemetry
This PR gets us DataDog telemetry and logging. DD request tracing and session recording is disabled all the time. Developers can optionally disable all telemetry with the `disableTelemetry` property on `CTWProvider`. All errors are being logged currently, however, I suspect we will lock that down very soon (maybe even in this PR). There are 2 dataset attributes which can be used on any HTMLElement to achieve and support event logging. These two attributes are `data-zus-telemetry-click` and `data-zus-telemetry-namespace`. Take a look at this contrived markup:
```html
<div data-zus-telemetry-namespace="Backyard">
  <div data-zus-telemetry-namespace="Trampoline">
    <button onClick={someHandler} data-zus-telemetry-click="Jump jump!">
       Jump on Trampoline
    </button>
  </div>
</div>
```
Clicking the above button would yield a log message like `"click event: Backyard > Trampoline > Jump jump!"`. That log message would be apart of a larger JSON event object that contained additional information such as the date-time, Zus user details, builder name and ID, web origin, geoip, browser agent and a few other details.

Additionally we get page load metrics from basic RUM setup. It's unclear how valuable this will be as currently it is measuring page load times which could likely be impacted more by customer code than our own. Similar to the error logging, this might be too much information and we may need to further pare it down. We don't have as much wiggle room with which metrics are posted as we do with the logging.

---
Here are a few screenshots of the logs and errors in DataDog. We can drill into any log message to get granular details, like who the user was that generated the log.
Screen Shot 2023-01-17 at 4 24 41 PM

Screen Shot 2023-01-18 at 2 58 51 PM

The Error logs have stack traces. These are only errors caught by our error boundary though.
Screen Shot 2023-01-18 at 3 01 10 PM

Here is a screenshot of the RUM dashboard.
Screen Shot 2023-01-17 at 4 38 54 PM
You may have noticed in the screenshot above that we get the version of the deployed code with the errors which is neat, it would be a shame to lose that if we can't keep RUM error handling and instead need to roll our own in the DD Logging. We could log the version as well though.


src/utils/telemetry.ts
```diff
@@ -0,0 +1,195 @@
...
+  AUTH_IS_SUPER_ORG,
+  AUTH_PATIENT_ID,
+  AUTH_PRACTITIONER_ID,
+  AUTH_USER_ID,
+  AUTH_USER_TYPE,
+  ZusJWT,
+} from "@/utils/auth";
+
+// Assume environment is production unless localhost
+let defaultEnvironment = "production";
```
luthfib, line 18: debating whether this is good or not. I guess we should theoretically always pass an env in right? But what if we pass in a typo in the env, then we would invartently send logs to prod? We might have fail safes before, but not sure..

luthfib (commented):
Looking good to me, left a question about the defaultEnvironment. Might also be good to add docs to the repo, so that we know we have to add this new datadog data attrs to future components.

src/utils/telemetry.ts
```diff
@@ -0,0 +1,239 @@
...
+    let ns = namespace;
+    if (target.dataset.zusTelemetryNamespace) {
+      const nextNamespace = target.dataset.zusTelemetryNamespace;
+      ns = ns ? `${nextNamespace} > ${ns}` : nextNamespace;
+    }
+    if (target.parentElement && target.parentElement !== window.document.body) {
+      return this.lookupComponentNamespace(target.parentElement, ns);
+    }
+    return ns || "unknown";
+  }
```
WesleyKapow, line 193: Can probably simplify this a bit with https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

src/utils/telemetry.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * Process Click Event - This function takes a target element and an optional
+   * explicitTargetName. If the `explicitTargetName` is NOT passed in, we need to
+   * traverse the DOM and verify whether it corresponds with a zus telemetry
+   * event or not. However, if `explicitTargetName` is passed in, then we can
+   * assume this is an event ready for processing.
+   */
+  static processHTMLEvent(
```
WesleyKapow, line 202: Same thing with this, you can probably simplify with https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

src/components/core/telemetry-boundary.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface State {
+  hasError: boolean;
+  error?: Error;
+  errorInfo?: ErrorInfo;
+}
+
+/**
+ * Error Boundary for internal use.
+ */
+export class TelemetryBoundary extends Component<Props, State> {
```
WesleyKapow, line 20: 1. This is like the only class based component. Favor sticking with hooks!
2. Should this be called `ErrorBoundary` ?
3. How does this look for our different components? When is it triggered?

src/components/core/table/table-head.tsx
```diff
@@ -49,9 +50,16 @@ export const TableHead = <T extends MinRecordItem>({
...
-            column.sortIndices && onSort && onSort(column.title || "")
-          }
+          onClick={(event) => {
+            if (column.sortIndices && onSort) {
+              Telemetry.processHTMLEvent(
+                event.currentTarget,
+                "zusTelemetryClick",
+                `sort=${column.title}`
+              );
+              onSort(column.title || "");
```
WesleyKapow, line 60: Why can't this use `data-zus-telemetry-click`?

src/components/core/providers/ctw-provider.tsx
```diff
@@ -35,11 +36,12 @@ type AuthTokenSpecified = { authToken: string; authTokenURL?: never };
 type AuthTokenURLSpecified = { authToken?: never; authTokenURL: string };
 
 type CTWProviderProps = {
+  builderId?: string;
   children: ReactNode;
+  disableTelemetry?: boolean;
```
WesleyKapow, line 41: We should think if we want this an opt-out or an opt-in. I don't feel strongly either way.

src/components/core/providers/ctw-provider.tsx
```diff
@@ -95,6 +109,16 @@ function CTWProvider({ theme, children, ...ctwState }: CTWProviderProps) {
...
 
+  useEffect(() => {
+    if (!disableTelemetry) {
+      Telemetry.init(ctwState.env);
+      Telemetry.setBuilder(ctwState.builderId);
+      handleAuth()
+        .then((accessToken) => Telemetry.setUser(accessToken))
+        .catch(() => Telemetry.clearUser());
+    }
+  }, [ctwState.builderId, ctwState.env, disableTelemetry, handleAuth, token]);
```
WesleyKapow, line 120: Don't know odds of this happening but if `disableTelemetry` for some reason was initially `false` and then was `true` we'd want to cleanup and disable telemetry. I think I'd feel more comfortable if we did that here!

src/components/content/medications/other-provider-meds-table.tsx
```diff
@@ -54,46 +55,49 @@ export function OtherProviderMedsTable({
...
-        rowMenuActions={(medication) =>
-          compact([
-            {
-              name: "View History",
-              action: async () => {
-                openHistoryDrawer(medication);
+    <TelemetryBoundary name="OtherProviderMedsTable">
+      <div data-zus-telemetry-namespace="OtherProviderMedsTable">
+        <MedicationsTableBase
+          telemetryNamespace="MedicationsTableBase"
```
WesleyKapow, line 61: Weird seeing you passing the name of the component to the component like this. Why is this needed? Can the component set the namespace? Do we need that namespace?

README.md
```diff
@@ -95,6 +95,118 @@ function PatientPortal({ patientID }: PatientPortalProps) {
 }
 ```
 
+## Telemetry
+
+Telemetry is sent to DataDog from CTW components. This telemetry tracks usage so that Zus can continually
+improve the ctw-component-library and fix issues that might otherwise go unnoticed. Currently, the telemetry
+provides a basic logger, a React error boundary to log errors and a few HTML dataset attributes to inform which
+elements have trackable events to listen for and the context of which CTW component the event was trigger from.
```
WesleyKapow, line 103:
```suggestion
elements have trackable events to listen for and the context of which CTW component the event was triggered from.
```

WesleyKapow, line 103: This plus the opt out part are probably the only things that need to be in this README. This README is meant for our customers. You can move the rest to a new file like `TELEMETRY.md` or something more generic like `MAINTAINERS.md`

README.md
```diff
@@ -95,6 +95,118 @@ function PatientPortal({ patientID }: PatientPortalProps) {
...
+> React error boundaries don't catch all errors. For example, they don't catch errors thrown from event handlers.
+> Contributors trying to log errors from event handlers should use the Telemetry logger. Read more about error
+> boundaries and their limitations [here](https://reactjs.org/docs/error-boundaries.html)
+
+### Opting out of telemetry
+
+Zus has a BAA with DataDog and we understand that your applications need to meet HIPAA regulations the same as ours.
+We take this into account when choosing which telemetry data can be tracked, but we also respect the choice to opt-out
+of telemetry collection all together. We encourage you to reach out to the team to discuss which telemetry we collect.
+However, you choose to simply opt-out, simply pass the boolean property `disableTelemetry` to the `CTWProvider`
```
WesleyKapow, line 203:
```suggestion
To opt-out, pass the boolean property `disableTelemetry` to the `CTWProvider`
```

WesleyKapow (changes requested):
This was certainly a triumph! I love that we'll now have error reporting! Great job on all the documentation and comments too!

Minor process feedback -- In future aim for smaller PRs. The first PR here could have been the setup + just a few example uses. That will allow for feedback prior of plumbing this everywhere AND it'll help keep the bulk of the work from touching so many files which could have merge conflicts. Nothing to do for this PR, just a note and reminder for the future.

src/components/core/providers/ctw-provider.tsx
eric-gunther: I think we should go opt-in as that seems to be the most careful option and would require us to get explicit approval from each customer utilizing the component library.

I also wonder if it makes sense to have the service name, token and application ID (ie. all the stuff in the DD config object) be passed in to the `CTWProvider` via something like a `DataDogConfig` prop. That way those values are out of this public repo (not that they are necessarily secrets, but perhaps better not to have them totally out in the open), and customers could potentially change those values if they wanted to write to their DD instance instead.

WesleyKapow: Agreed with opt-in.

As for the tokens and application ID. I'm not sure. Seems like added friction to have to tell customers to set these up... and then they hardcode them? Not sure likelihood of them wanting to point to their own datadog as well.

src/utils/telemetry.ts
mrosata: oh that's nice. I'll make the change

src/components/core/telemetry-boundary.tsx
mrosata: 1. afaik only class components support the `getDerivedStateFromError` and `componentDidCatch`, it's possible that changed in some recent version of React 18.* and I didn't hear about it, but def not supported in 16/17.
2. We could call it `ErrorBoundary` or more aptly `TelemetryErrorBoundary`.
3. This might not have been in the README.md when you reviewed 4 days ago. It's there now with some other info, but essentially it acts as a Fragment around the body of a component and catches errors from that component. When it catches an error we send a log, but additionally it allows us to render something in the errored components place rather than React unmounting every component up the tree until either root or another error boundary. There are some caveats, which is that it doesn't catch errors from event handlers _(which is also why event handler errors don't unmount the component tree)_. Errors that it will catch would be something like trying to render some mis-shaped data  that tries access a property of a nonexistent object.
```tsx
import { TelemetryBoundary } from "@/components/core/telemetry-boundary";

function SomeComponent(props) {
  return (
    <TelemetryBoundary name="SomeComponent">
      <h1>Wow! This is some component!</h1>
      <AnotherComponent />
    </TelemetryBoundary>
  );
}
```

src/components/core/providers/ctw-provider.tsx
mrosata: For the opt-in, I agree, I'll make that change now.

src/components/content/medications/other-provider-meds-table.tsx
mrosata: I tried to get the name of the component inside `TelemetryBoundary` and it besides ts-lint errors it wasn't reliable. It was basically getting the function name also, so in production builds it would be nonsense. There is a way to set names on functional components in react, but it is more awkward than setting a property on the TelemetryBoundary.

The `name` prop isn't required, so it can be omitted, but then errors caught in the error boundary don't have namespace context. The error boundary doesn't benefit from the data-zus-telemetry-namespace as we get an error as context rather than an HTMLElement.

README.md
mrosata: This is outdated, I updated docs to reflect the `opt-in`.

src/utils/telemetry.ts
mrosata: ironically the function grew in size to support `Element.closest(selector)`. It is probably easier to grok though so going to leave it in. The reasons the code requires more lines with closest is just:
1. closest returns an `Element`, so we have to setup a type check for `instanceof HTMLElement` (it's possible we could use `as` but idk the caveats there.
2. closest is inclusive of the current element, so we can't pass it into `this.lookupComponentNamespace` recursively, we need to verify that `element.parentElement` exists and pass that in instead.

---
This was the final result _(already pushed)_
```ts
private static lookupComponentNamespace(
  target: HTMLElement,
  namespace = ""
): string {
  let ns = namespace;
  const closest = target.closest("[data-zus-telemetry-namespace]");
  if (closest instanceof HTMLElement) {
    const nextNamespace = closest.dataset.zusTelemetryNamespace;
    if (typeof nextNamespace === "string") {
      ns = ns ? `${nextNamespace} > ${ns}` : nextNamespace;
    }
    if (closest.parentElement) {
      return this.lookupComponentNamespace(closest.parentElement, ns);
    }
  }
  return ns || "unknown";
}
```

src/utils/telemetry.ts
mrosata: I might leave this as-is because unlike `lookupComponentNamespace`, `processHTMLEvent` doesn't actually know 100% that the HTMLEvent it is processing is a zus-telemetry-event. It could be any click or focus event on the page.  `processEvent` has a max search depth of 5 so it won't search up the entire DOM on every click the user makes.

The reason this short-walk up the tree is needed in the first place is just because if someone clicks the span with id="inner"
```html
<button data-zus-telemetry-click="awesome button">
  <span id="inner">CLICK ME</span>
</button>
```
we would see they clicked a span, but not know they actually clicked the button. So when the user clicks anything on page, we walk up to 5 elements up the tree looking for the data-zus-telemetry-click attribute, if we don't find it then we assume it's not a relevant event and stop walking.

src/components/core/telemetry-boundary.tsx
WesleyKapow: 1. Ah didn't catch that, makes sense.
2. Bikeshedding here but I'd go with `ErrorBoundary`, the fact that it logs telemetry is just a feature of it. It is not an error boundary for telemetry components.
3. Are we going to get duplicate error logs, one from the location of the `throw` and another from this boundary? How can we create a good pattern here where we only ever get one? Can we get it so we can just throw and the error boundary logs / less boilerplate / PIT OF SUCCESS!

src/components/core/table/table-head.tsx
WesleyKapow: This seems very error prone. We could have an element with `data-zus-telemetry-click` and then in the future add an `onClick` and all of a sudden telemetry is broken for that element. That's a footgun (opposite of pit of success) since it is 100% non-obvious to developers editing the code.

See if you can investigate further what is going on here. Maybe the window click is firing but with the wrong target?

I just tried the following which correctly logs all three when clicking on the "rar" button:
```tsx
function onClick1() {
  console.log("on click 1");
}
function onClick2() {
  console.log("on click 2");
}
window.document.body.addEventListener("click", (event) => {
  console.log("window click", event.target);
});

// JSX...
<div onClick={onClick1}>
  <button onClick={onClick2}>rar</button>
  hello world
</div>
```

src/components/content/medications/other-provider-meds-table.tsx
WesleyKapow: Oh also you pass in a different namespace for the same component in provider-meds-table. I guess just change the namespace to something more meaningful like `OtherProviderMedicationsTable`.

src/components/core/providers/ctw-provider.tsx
mrosata: @WesleyKapow this is outdated because the `disableTelemetry` became `enableTelemetry`, but the context of your comment isn't outdated.

I can add a `disableTelemetry()` method.. it's going to be very grey area and possibly misleading. I think we can clear the global context and user context, we can certainly stop sending our own custom logs, but from what I've seen, I don't think we can just flip the off switch after it's been initialized. Switching from enabled `false` to `true` is supported for us atm, but going from enabled to disabled the best thing we can probably do is throw an error because I don't think we can guarantee telemetry had actually 100% stopped.

src/components/content/medications/other-provider-meds-table.tsx
mrosata: namespace and component aren't necessarily the same thing.

src/components/core/table/table-head.tsx
mrosata: I think your point here might be valid, but the example doesn't exemplify any problems for our telemetry.

One example of a problem might be that we can't prevent event handlers from doing `event.stopPropigation()` _(as we had discussed at offsite)_. I don't know how we can circumvent someone using a browser API to stop an event from propagating other than explicitly calling the logger.

Another potential issue, and possibly this exact issue we see are seeing in the sorting columns, could be that the browser doesn't always handle clicks through SVGs the same as the DOM. It's only when clicking the SVG caret in the table head is clicked that the events don't propagate. I've seen people mention this in non-react projects and say `pointer-events: auto` fixes it, but that didn't do the trick for me.

As for the target stuff, this is the reason for the tree traversal that happens in our click handler. When someone clicks "hello world", we get the target in our handler and then the handler asks:
- does this target have `data-zus-telemetry-click` on it?
  - YES? Good, lets start processing the event (getting namespace and actually logging)
  - NO? Ok, let's repeat this inspection on the parent (up to a predetermined depth)

I don't know that we can make perfect the enemy of the good here. I strived to make this as good as I could, with as little hackery as I could, with as much benefit to the company and as little obtrusiveness to the developer as I could. There will be issues and we will have to figure out how to solve them. Right now the only issue is that the SVG clicks don't propagate up to the window, but they do make it to the `onClick` handler, so it doesn't seem too out there to just conditionally call the telemetry from there.

WesleyKapow: This isn't about perfection or anything like that. These types of work arounds are footguns (opposite of pit-of-success). How will future developers know when they need to do this? What if I add an `onClick` to an existing `data-zus-telemtry-click` element? Will I break it? Why is this code doing it this way and not the other way? Any work arounds like this should be verified that they are needed and then well documented to warn other developers about these non-obvious issues.

In this particular case, you had a theory about the issue:
> I believe that the React onClick handler stopped the event from propagating up to window

Sounds a bit odd to me (react shouldn't be breaking such a crucial and standard part of the spec), but ok, let's verify it. My code above proves that the event does reach the listener on window. So what is the issue? Maybe svg's behave weird with clicks. I added some code and verified that wasn't the issue either. So now I'm wondering why the heck the window event listener doesn't fire. I added some console logs and it turns out it does fire! The root of the issue turns out to be a boolean issue in your handler.

```ts
      if (!(isTrusted && target instanceof HTMLElement)) {
        return;
      }
```

This fails for svgs!

Screenshot 2023-01-30 at 4 20 53 PM

Your other point about not being able to stop people from calling `event.stopPropagation();` is a good one. I tested that too and I looks like listeners on window are not affected, so this shouldn't be a problem either. (my guess is listening for any clicks on window does not involve propagation)

Be very mindful of any assumptions you have for work arounds. See if you can quickly hack together an example to verify the issue. And then we should be very careful about any non-obvious issues (foot-guns) and see if there are ways to protect ourselves (turn them into pit-of-success) and if not, add documentation and education.

src/App.tsx
```diff
@@ -50,13 +50,14 @@ const DemoApp = ({ accessToken = "" }) => (
...
     builderId={VITE_BUILDER_ID}
+    enableTelemetry
   >
     <PatientProvider patientID={VITE_PATIENT_ID} systemURL={VITE_SYSTEM_URL}>
       <div className="App">
         <h1>CTW Component Library</h1>
 
-        {demoComponents.map((demo) => (
-          <div className="ctw-space-y-5 ctw-bg-white" key={demo.title}>
+        {demoComponents.map((demo, index) => (
```
WesleyKapow, line 59: ooh first time seeing `demoComponents` refactor, love it!

src/components/core/error-boundary.tsx
```diff
@@ -80,3 +80,12 @@ export class TelemetryBoundary extends Component<Props, State> {
     return props.children;
   }
 }
+
+export function withErrorBoundary<T>(
+  wrappedComponent: (props: T) => ReactNode,
+  name?: string
```
WesleyKapow, line 86: All of these things work out if there's no name? We just won't get `componentName`?

src/utils/telemetry.ts
```diff
@@ -103,17 +103,19 @@ export class Telemetry {
     // nodes of the DOM before knowing whether this event was relevant to us.
     window.document.body.addEventListener("click", (event) => {
       const { target, isTrusted } = event;
-      if (!(isTrusted && target instanceof HTMLElement)) {
+      if (!(isTrusted && target instanceof Element)) {
         return;
       }
-      this.processHTMLEvent(target, "zusTelemetryClick");
+      const htmlElement = this.closestHTMLElement(target);
+      if (htmlElement instanceof HTMLElement) {
```
WesleyKapow, line 110: Why is it important to get an HTMLElement and not just an element?

WesleyKapow (approved):
Changes look great, just left a few questions for ya.

src/components/core/error-boundary.tsx
mrosata: correct

src/utils/telemetry.ts
mrosata: The `HTMLElement` has dataset attributes and `Element` does not. I did quickly try converting the `processEvent` method and using `nextTarget.dataset?[telemetryKey]` but ts still complains that `dataset` does not exist on type Element.

WesleyKapow: oh gross, ok you could go with `HTMLElement | SVGElement` which both have a dataset

WesleyKapow: gross microsoft -- `HTMLOrSVGElement`

https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.htmlorsvgelement.html

PR #408, opened 2023-01-17
Fix radix dropdown z index

luthfib:
Fix z-index with radix dropdown menu. Change background color of condition 2.0 to white, to better reflect design.

### DIFF
--- a/src/components/core/main.scss
+++ b/src/components/core/main.scss
@@ -125,7 +125,7 @@
     @apply ctw-flex ctw-items-start ctw-space-x-2 ctw-p-4 ctw-text-left;
   }
 
-  .ctw-provider [data-radix-popper-content-wrapper] {
+  .ctw-provider > [data-radix-popper-content-wrapper] {
     // We add important here because radix z-index will take presedence if we don't.
     z-index: 10000 !important;
   }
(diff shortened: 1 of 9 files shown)


pmadria3 (dismissed):
LGTM

src/components/core/main.scss
```diff
@@ -124,4 +124,9 @@
   .ctw-alert-bg {
     @apply ctw-flex ctw-items-start ctw-space-x-2 ctw-p-4 ctw-text-left;
   }
+
+  .ctw-provider [data-radix-popper-content-wrapper] {
```
WesleyKapow, line 128: Blocker: As discussed, we should target direct descendant which I believe is `.ctw-provider > [data-radix-popper-content-wrapper]`, that way we won't target any other ones of these if they happened to exist (via consuming app using radix as well).

WesleyKapow (changes requested):
One blocker, otherwise lgtm

src/components/core/main.scss
WesleyKapow: https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator

WesleyKapow (approved):

PR #412, opened 2023-01-18
CTW-619/list-navigation

pmadria3:
[CTW-619
](https://zeushealth.atlassian.net/jira/software/projects/CTW/boards/49?selectedIssue=CTW-619)

This PR changes the layout of the headers for the table.

Selected tab: 
Screenshot 2023-01-19 at 5 05 42 PM

Hover: 
Screenshot 2023-01-19 at 5 06 32 PM


src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -0,0 +1,105 @@
+import { Tab } from "@headlessui/react";
+import cx from "classnames";
+import { useAddConditionForm } from "./condition-hooks";
+import { FilterCollection } from "./patient-conditions-filters";
+import { Badge } from "@/components/core/badge";
+import { Toggle } from "@/components/core/toggle";
+import { ConditionModel } from "@/fhir/models";
+
+export type PatientConditionsActionsProps = {
```
WesleyKapow, line 9: File is `PatientConditionsTabs` .. not actions! Also if it is tabs, then I'd expect it to handle the tabs part of the component and not the next row which has the "actions".

src/components/content/conditions/patient-conditions-actions.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export function PatientConditionsActions({
+  hideAdd,
+  onToggleShowHistoric,
+}: PatientConditionsActionsProps) {
+  const showAddConditionForm = useAddConditionForm();
+
+  return !hideAdd ? (
+    <div className="ctw-flex ctw-items-center ctw-justify-end ctw-space-x-2 ctw-p-4">
+      <Toggle name="historic" text="Historic" onChange={onToggleShowHistoric} />
```
WesleyKapow, line 17: This should be visible regardless of `hideAdd`

src/components/content/conditions/patient-conditions-actions.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <button
+        type="button"
+        className="ctw-btn-primary"
+        onClick={() => showAddConditionForm()}
```
WesleyKapow, line 21: Can you just pass `showAddConditionForm` or do you have to wrap it in the function?

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -34,41 +26,47 @@ export function PatientConditionsActions({
...
+            >
+              {({ selected }) => (
                 <button
                   type="button"
                   className={cx(
                     activeClass("other"),
-                    "ctw-tab-underline ctw-tab ctw-relative ctw-space-x-2"
+                    selected
+                      ? "ctw-tab-underline ctw-tab-underline-hover ctw-tab ctw-relative ctw-space-x-2"
+                      : "ctw-tab-underline ctw-tab-underline-hover:focus-visible ctw-tab ctw-relative ctw-space-x-2"
```
WesleyKapow, line 69: `:foo` are not actual classes and are instead "pseudo" selectors. You don't add them to the classes and instead only use them in CSS. Like you would NOT do `className="foo:hover bar:after"`. So you should be able to remove these `:focus-visible` classes.

src/components/content/conditions/patient-conditions.scss
```diff
@@ -50,13 +50,19 @@
     @apply ctw-cursor-pointer ctw-border-0 ctw-bg-transparent ctw-p-0 ctw-text-base;
   }
 
-  .ctw-tab-underline {
+  .ctw-tab-underline-hover {
```
WesleyKapow, line 53: It's weird to call this hover, what does it have to do with hovering? Is this the class that gets applied when a tab is selected? So should there be a `ctw-tab-selected`?

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -34,41 +26,47 @@ export function PatientConditionsActions({
...
+            >
+              {({ selected }) => (
                 <button
                   type="button"
                   className={cx(
                     activeClass("patient"),
-                    "ctw-tab-underline ctw-tab ctw-relative"
+                    selected
+                      ? "ctw-tab-underline-hover ctw-tab-underline ctw-tab ctw-relative"
+                      : "ctw-tab-underline-hover:focus-visible ctw-tab-underline ctw-tab ctw-relative"
```
WesleyKapow, line 46: With `cx` you can conditionally added classes a few ways, here, I'd split out the classes that are always the same, and then only add the optional ones based on the bool, something like `"ctw-tab-underline ctw-tab ctw-relative", {"ctw-tab-selected": selected}`

WesleyKapow (dismissed):

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -74,32 +74,37 @@ export function PatientConditions({
...
+          otherConditions={otherConditions}
+          collection={filters.collection}
+          onCollectionChange={(collection) => updateFilters({ collection })}
+        />
+        <PatientConditionsActions
+          hideAdd={readOnly || filters.collection === "other"}
+          onToggleShowHistoric={() =>
+            updateFilters({ showHistoric: !filters.showHistoric })
+          }
+        />
```
WesleyKapow, line 91: newline after this, just to help visually break up these large component uses

src/components/content/conditions/patient-conditions-actions.tsx
pmadria3: It has to be in a function

src/components/content/conditions/patient-conditions-actions.tsx
pmadria3: Do we want to show the historic/Add condition button in other provider records? If I remove the hideAdd it will show in that tab as well...right now it hides it

src/components/content/conditions/patient-conditions.scss
```diff
@@ -45,4 +45,28 @@
...
   }
+
+  .ctw-tab {
+    @apply ctw-cursor-pointer ctw-border-0 ctw-bg-transparent ctw-p-0 ctw-text-base;
+    &:focus-visible {
+      @apply ctw-outline-none;
+    }
+  }
+
+  .ctw-tab-underline-hover {
```
luthfib, line 56: This could be changed to `.ctw-tab-underline`

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -0,0 +1,68 @@
...
+            >
+              Condition List
+            </Tab>
+            <Tab
+              className={({ selected }) =>
+                cx(
+                  activeClass("other"),
+                  selected
+                    ? "ctw-tab-underline-selected ctw-tab-underline-hover ctw-tab ctw-relative ctw-space-x-2"
+                    : "ctw-tab-underline-hover ctw-tab ctw-relative ctw-space-x-2"
```
luthfib, line 56: This is a bit confusing. I think this could be improved by always applying the common classes. `ctw-tab-underline-hover ctw-tab ctw-relative`. Im curious as to why `ctw-space-x-2` is needed when selected by not when selected?

luthfib (dismissed):
Looking good, however still some changes around cx

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -25,57 +25,48 @@ export function PatientConditionsTabs({
     (condition) => condition.displayStatus === "Active"
   ).length;
 
+  function blurClicked() {
```
WesleyKapow, line 28: Add a link to the issue that explains why this is needed.

WesleyKapow, line 29: Add a link to the issue that explains why this is needed.

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -25,57 +25,48 @@ export function PatientConditionsTabs({
...
+    if (document.activeElement) {
```
WesleyKapow, line 30:
```suggestion
    // Guard against server side where document isn't defined.
    if (document) {
```

src/components/content/conditions/patient-conditions-tabs.tsx
WesleyKapow: ugh sorry, can you also add a comment summarizing the issue and why this is needed, that way fellow devs will know from just the code and can follow the link if they want to learn or explore the issue more.

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -25,57 +25,49 @@ export function PatientConditionsTabs({
...
+            onCollectionChange(index === 0 ? "patient" : "other");
+          }}
+        >
+          <Tab.List>
             <Tab
-              className={cx(
-                activeClass("patient"),
-                "ctw-tab-underline ctw-tab ctw-relative"
-              )}
+              onClick={() => blurClicked()}
```
WesleyKapow, line 52: No need to wrap this in a function.

WesleyKapow (changes requested):

src/components/content/conditions/patient-conditions-tabs.tsx
pmadria3: Had to add an eslint-disable because it was always truthy in this case

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const activeCount = otherConditions.filter(
+    (condition) => condition.displayStatus === "Active"
+  ).length;
+
+  // Issue: Clicking on tab enables focus-visible and not only via keyboard interactions.
+  // Resolution: https://github.com/tailwindlabs/headlessui/issues/1694
+  function blurClicked() {
+    // Guard against server side where document isn't defined.
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (document) {
```
mrosata, line 33:
```suggestion
    if (typeof document !== "undefined") {
```

src/components/content/conditions/patient-conditions-tabs.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <div className="ctw-space-x-4">
+        <Tab.Group
+          // Keyboard navigation requires onChange instead of onClick.
+          onChange={(index) => {
+            onCollectionChange(index === 0 ? "patient" : "other");
+          }}
+        >
+          <Tab.List>
+            <Tab
+              onClick={blurClicked}
```
mrosata, line 54: Is this going to work with keyboard nav or would it be better just to call `blurClick` above in onChange as well?

pmadria3: Yep so this works with the keyboard nav using `onClick` but `onChange` it doesn't seem to toggle back

WesleyKapow: We specifically only want to call the blur function on click. That way we don't get the outline when you click but you will get it when tabbing (keyboard use).

src/components/content/conditions/patient-conditions.scss
```diff
@@ -47,15 +47,15 @@
...
+    @apply ctw-cursor-pointer ctw-border-transparent ctw-bg-transparent ctw-p-2 ctw-text-base;
     &:focus-visible {
       @apply ctw-outline-2;
     }
   }
 
   .ctw-tab-underline {
     &:after {
-      @apply ctw-absolute ctw-left-0 -ctw-bottom-5 ctw-block ctw-h-0.5 ctw-w-full;
+      @apply ctw-absolute ctw-left-0 -ctw-bottom-0 ctw-block ctw-h-0.5 ctw-w-full;
```
WesleyKapow, line 58: Negative bottom-0 is weird ya?

WesleyKapow (approved):

src/components/content/conditions/patient-conditions.scss
pmadria3: yeah but the bar seems to slid all the way up into the button if I remove it :( 
Screenshot 2023-01-20 at 2 47 19 PM

With -0: 
Screenshot 2023-01-20 at 2 47 44 PM

PR #413, opened 2023-01-18
Debugging vite build output

mrosata:
Created "@/utils/nodash" and "@/utils/nodash/fp" modules to replace all usage of lodash in the project. The reason for this is because vite is unable to properly tree-shake lodash when functions are imported from the global "lodash" package. This change should yield a smaller build and more importantly it won't clobber `window._` with the lodash library.

In the future when importing functions from lodash, instead of doing:
```ts
import { get, pick } from "lodash/fp";
import { reduce } from "lodash";
```
we should instead write
```ts
import { get, pick } from "@/utils/nodash/fp";
import { reduce } from "@/utils/nodash";
```
If a function from `lodash` or `lodash/fp` isn't available in `@/utils/nodash` or `@/utils/nodash/fp` yet, then just go into the appropriate nodash utils file and add the function. That would look something llike this:
```ts
# "./src/utils/nodash/index.ts"
export { reduce as default } from "lodash";
// ... ect ect
```

> The name "nodash" refers to the fact that we should no longer have a `_` lodash in the global namespace... no lodash = nodash. It's cute, don't worry about it.


.changeset/brave-trains-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+do not make regular dependencies external
```
pmadria3, line 5:
```suggestion
Remove showing regular dependencies externally.
```

luthfib, line 5:
```suggestion
Do not mark regular dependencies external.
```

pmadria3 (dismissed):
LGTM

pmadria3 (approved):

luthfib (approved):

PR #417, opened 2023-01-20
CTW-706/Eslint rule to disallow lodash and lodash/fp imports

mrosata:
To avoid accidentally leaking the lodash `_` into the global scope we are disallowing imports directly from base packages "lodash" or "lodash/fp". Preferred usage would be to import and export lodash functions directly within the project modules "@/utils/nodash" and "@/utils/nodash/fp". See #413

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -84,6 +84,8 @@ module.exports = {
             message:
               "`useLayoutEffect` causes a warning in SSR. Use `useIsomorphicLayoutEffect` instead.",
           },
+          { name: "lodash" },
+          { name: "lodash/fp" },
         ],
       },
     ],
--- a/package.json
+++ b/package.json
@@ -122,11 +122,5 @@
   },
   "msw": {
     "workerDirectory": "public"
-  },
-  "overrides": {
-    "react-json-view": {
-      "react": "^16.9.0 || ^17 || ^18",
-      "react-dom": "^16.9.0 || ^17 || ^18"
-    }
   }
 }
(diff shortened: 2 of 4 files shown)


mrosata:
@pmadria3 you can confirm that the react-json-viewer wrapper and overrides config is no longer needed?

package.json
```diff
@@ -122,11 +122,5 @@
   },
   "msw": {
     "workerDirectory": "public"
-  },
-  "overrides": {
-    "react-json-view": {
-      "react": "^16.9.0 || ^17 || ^18",
-      "react-dom": "^16.9.0 || ^17 || ^18"
-    }
```
mrosata, line 130: @pmadria3 you can confirm that the react-json-viewer wrapper and overrides config is no longer needed?

luthfib (approved):
LGTM. Pretty sure that json viewer stuff is no longer needed, but we can get Priyank to 100% confirm

luthfib (commented):
Don't forget to add a changeset

mrosata:
> Don't forget to add a changeset

Thanks, I am going to skip the changeset on this one b/c the eslint rules will take effect without one and the code changes should be a no-op.

package.json
pmadria3: yep we don't need the react-json-view anymore

pmadria3 (approved):

.eslintrc.cjs
```diff
@@ -84,6 +84,8 @@ module.exports = {
             message:
               "`useLayoutEffect` causes a warning in SSR. Use `useIsomorphicLayoutEffect` instead.",
           },
+          { name: "lodash" },
+          { name: "lodash/fp" },
```
WesleyKapow, line 88: Can you please follow up with adding a message here like the other one. Hopefully the message is displayed when you try to import this. That way future-us and future-others won't be like "wtf why is eslint complaining about lodash"

PR #419, opened 2023-01-20
Add handful of form validations to patient request form

WesleyKapow:
* State is now an enum / select dropdown of state codes. We also properly convert state to the code when pre-populating.
* Date of birth has to be > 1900 (fixes issue where user could enter 2 digits for year)
* Zip codes are now pattern validated.
* Email is now pattern validated.
* Phone numbers are now pattern validated.
* Updated form-helpers to no longer add the "pattern" attribute on input fields based on schema. This fixes an issue where the browser would end up doing the validation and not us!

Screenshot 2023-01-20 at 4 31 04 PM
Screenshot 2023-01-20 at 4 30 52 PM
Screenshot 2023-01-20 at 4 30 40 PM
Screenshot 2023-01-20 at 4 30 35 PM
Screenshot 2023-01-20 at 4 30 31 PM


pmadria3 (approved):
LGTM

PR #421, opened 2023-01-24
CTW-653/timeline document viewer

WesleyKapow:
CTW-653

**NOTE TO REVIEWER: Best to review one commit at a time.**

* Refactor provenance & binary fetching (code cleanup, generalized for any resource type, & uses react query for caching).
* Refactor CCDA modal to use new `useModal` hook.
* Setup patient timeline to show source document.

Screenshot 2023-01-24 at 11 53 50 AM


src/components/content/condition-history/conditions-history.tsx
```diff
@@ -77,15 +54,15 @@ export function ConditionHistory({
...
+        const provenances = await searchProvenances(requestContext, [
+          condition,
+          ...conditionsDataDeduped,
+        ]);
 
         let binaryId;
         const conditionsToDisplay = conditionsDataDeduped.map(
           (dedupdedCondition) => {
-            binaryId = getBinaryId(provenanceBundles, dedupdedCondition.id);
+            binaryId = getBinaryId(provenances, dedupdedCondition.id);
```
mrosata, line 65: Won't block on this, as the temporary closure created by `let binaryId;` was there previously. It makes more sense to just declare the variable inside the mapped function.
```suggestion
        const conditionsToDisplay = conditionsDataDeduped.map(
          (dedupdedCondition) => {
            const binaryId = getBinaryId(provenances, dedupdedCondition.id);
```

src/fhir/binaries.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const hasTarget = some(provenance.target, (t) =>
+      t.reference?.includes(targetId)
+    );
+
+    if (hasTarget) {
+      const source = find(provenance.entity, { role: "source" });
+      if (source) {
+        // Return the ID portion of the reference.
+        return source.what.reference?.split("/")[1];
```
mrosata, line 22: Is it possible that `source` is defined but `source.what.reference` is undefined? TypeScript seems to think so, but if it's a case of we know better than the TS engine feel free to resolve this convo.
```suggestion
      if (source?.what.reference) {
        // Return the ID portion of the reference.
        return source.what.reference?.split("/")[1];
```

src/components/content/condition-history/conditions-history.tsx
```diff
@@ -132,51 +132,38 @@ const HistoryRecords = ({
...
+                entry.title ?? "Condition"
+              )
+            }
+            text="Source Document"
+          />
+        );
+      }
+      return entry;
+    });
+  }
```
mrosata, line 153:
```suggestion
  function maybeAddDocumentButtons(entries: CollapsibleDataListStackEntries) {
    return entries.map((entry) => {
      if (!entry.binaryId) {
        return entry;
      }
      return {
        ...entry,
        documentButton: (
          <DocumentButton
            onClick={() =>
              openCCDAModal(
                entry.binaryId as string,
                entry.title ?? "Condition"
              )
            }
            text="Source Document"
          />
        ),
      };
    });
  }
```

Or, if importing `{ set }` "@/utils/nodash/fp" _(would need to be "fp" to avoid mutation)_ we could refactor to not use the spread operator.
```suggestion
  function maybeAddDocumentButtons(entries: CollapsibleDataListStackEntries) {
    return entries.map((entry) => {
      if (!entry.binaryId) {
        return entry;
      }
      return set(
        "documentButton",
        <DocumentButton
          onClick={() =>
            openCCDAModal(entry.binaryId as string, entry.title ?? "Condition")
          }
          text="Source Document"
        />,
        entry
      );
    });
  }
```

> Note: I don't actually have a preference if the `return entry;` line is at top or bottom of function. I moved it in my suggested examples in the hopes the 2 extra spaces would allow for better formatting. Which it did not, the 4 extra white spaces in the second suggestion did allow for the `openCCDAModal(entry.binaryId as string, entry.title ?? "Condition")` to fit on one line and not break formatting.

src/components/content/condition-history/conditions-history.tsx
WesleyKapow: Thanks! Done!

src/fhir/binaries.ts
WesleyKapow: Yes that's better! TY! This way we'll also keep looping to try and find a binary with a valid `what.reference`.

src/components/content/condition-history/conditions-history.tsx
WesleyKapow: Good suggestion, helps me remove the bad eslint hack. `set` seems superfluous as it doesn't really simplify things.

luthfib (approved):

src/components/core/main.scss
```diff
@@ -79,6 +79,8 @@
 
   .ctw-link {
     @apply ctw-cursor-pointer ctw-text-xs ctw-font-medium ctw-text-primary-dark hover:ctw-text-primary-main;
+    // Also handle any icons (svg fill).
+    @apply ctw-fill-primary-dark hover:ctw-fill-primary-main;
```
pmadria3, line 83: was this done for organization? Maybe we could just have it in the same apply statement?

pmadria3 (dismissed):
LGTM - nice rework with reactQuery as well : ) 

One small comment.

pmadria3 (approved):

PR #422, opened 2023-01-24
CTW-629/Tabbed Patient Medications

mrosata:
This PR creates an initial version of Medications 2.0. It doesn't have filters. The tables are the same as meds 1.0 with the addition of a "Historical" tab that shows inactive and dismissed medications _(NOTE: currently no inactive medications show up in the "Historical" tab because our API calls only return active meds, however if they ever do return inactive meds, the `splitMedications` function has been refactored to handle it)_.

https://www.loom.com/share/16b40d526f40484c98eabaf376e659a2


src/components/content/medications/patient-medications.scss
```diff
@@ -8,4 +8,22 @@
...
+    @apply ctw-relative ctw-mr-5 ctw-cursor-pointer ctw-border-0 ctw-border-transparent ctw-bg-transparent ctw-py-4 ctw-px-1;
+    &:after {
+      @apply ctw-absolute ctw-left-0 ctw-bottom-0 ctw-block ctw-h-0.5 ctw-w-full;
+      content: "";
+    }
+  }
+
+  .ctw-tab-list {
+    width: 312px;
+  }
```
eric-gunther, line 28: Seems like these drive some discrepancy between the condition and medication components - I'm assuming these were intentional, but it may be worth checking with Tai & Jesse on how the tab styling should officially look.

src/components/content/medications/provider-inactive-medications-table.tsx
```diff
@@ -0,0 +1,66 @@
+import { useEffect, useState } from "react";
+import { MedicationDrawer } from "@/components/content/medication-drawer";
+import { MedicationsTableBase } from "@/components/content/medications-table-base";
+import { MedicationStatementModel } from "@/fhir/models/medication-statement";
+import { useQueryAllPatientMedications } from "@/hooks/use-medications";
+import { get, pipe, toLower } from "@/utils/nodash/fp";
+import { sort } from "@/utils/sort";
+
+export type HistoricalMedRecordsTableProps = {
```
eric-gunther, line 9:
```suggestion
export type InactiveMedRecordsTableProps = {
```

src/components/content/medications/provider-inactive-medications-table.tsx
```diff
@@ -0,0 +1,66 @@
...
+  className?: string;
+};
+
+/**
+ * Displays a table of historical medical records. These are medications from
+ * other providers that have been dismissed by the builder viewing the table.
```
eric-gunther, line 15:
```suggestion
 * Displays a table of inactive medication records. These are builder-owned medications
 * that have a status other than "active". These are not to be confused with "dismissed" or "archived"
 * medication records pulled from the network of other providers.
```

src/components/content/medications/provider-inactive-medications-table.tsx
```diff
@@ -0,0 +1,66 @@
...
+ */
+export function ProviderInactiveMedicationsTable({
+  className,
+}: HistoricalMedRecordsTableProps) {
```
eric-gunther, line 19:
```suggestion
}: InactiveMedRecordsTableProps) {
```

eric-gunther (approved):
Tested locally using sandbox test patient. Had some questions, largely around maintaining consistency across conditions and meds components, but nothing blocking. I'm sure we'll need a later pass in order to get consistency across the components. but I'm not sure if that's really the highest priority right now anyway.

https://user-images.githubusercontent.com/97455910/215634828-f07009f6-900d-424b-985a-2020ff4d9c63.mov

src/components/content/medications/patient-medications.scss
mrosata: Did you see a discrepancy or just that the scss file itself is different?

eric-gunther: Yeah I think it's really just a matter of font-size and line-height, which are explicitly specified in this scss file and not in the conditions one. TBH I think the meds one looks better, but I'd defer to whatever design folks say. That said, I'm not blocking on this.

image

image

eric-gunther (approved):

PR #434, opened 2023-01-26
Allergies updates and dedupe

pmadria3:
This PR dedupes allergies and adds the manifestation column. Additionally, it dedupes immunizations as well.

Screenshot 2023-01-26 at 2 42 47 PM


luthfib (dismissed):
Fix esilnt issues

luthfib (approved):

PR #436, opened 2023-01-30
Allergies table add heading and better card view, FHIR button icon

mrosata:
Made the card view of the Allergies component look a little cleaner:
 - Not showing status on card view.
 - Not showing FHIR button as a column on card view (because it's vertical).
 - FHIR button on card view is a row action (on hover it shows up).
 - Updated Header to match other common components.

Updated the FHIR View component:
 - Use `ctw-text-sm` for smaller font size so the record looks cleaner and fits in view better.
 - Added a small fire icon next to the button and renamed it just "FHIR" for size reasons.

Screen Shot 2023-01-30 at 10 59 08 AM

---
These images compare the previous font size of FHIR resource JSON to the updated font size in this  PR (`ctw-text-sm`).

Previous/current font size:
Screen Shot 2023-01-30 at 11 00 54 AM

Updated font size _(this PR)_:
Screen Shot 2023-01-30 at 11 03 57 AM

### DIFF
--- a/src/components/content/allergies/patient-allergies-column.tsx
+++ b/src/components/content/allergies/patient-allergies-column.tsx
@@ -1,35 +1,39 @@
 import { TableColumn } from "@/components/core/table/table-helpers";
 import { ViewFHIR } from "@/components/core/view-fhir";
 import { AllergyModel } from "@/fhir/models/allergies";
+import { compact } from "@/utils/nodash";
 
-export const patientAllergiesColumns: TableColumn<AllergyModel>[] = [
-  {
-    title: "Description",
-    dataIndex: "display",
-  },
-  {
-    title: "Type",
-    dataIndex: "type",
-  },
-  {
-    title: "Status",
-    dataIndex: "clinicalStatus",
-  },
-  {
-    title: "Category",
-    dataIndex: "categories",
-  },
-  {
-    title: "Onset",
-    dataIndex: "onset",
-  },
-  {
-    title: "Manifestations",
-    dataIndex: "manifestations",
-  },
-  {
-    render: (allergy) => (
-      <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
-    ),
-  },
-];
+export const patientAllergiesColumns = (
+  cardView = false
+): TableColumn<AllergyModel>[] =>
+  compact([
+    {
+      title: "Description",
+      dataIndex: "display",
+    },
+    {
+      title: "Type",
+      dataIndex: "type",
+    },
+    cardView
+      ? null
+      : {
+          title: "Status",
+          dataIndex: "clinicalStatus",
+        },
+    {
+      title: "Category",
+      dataIndex: "categories",
+    },
+    {
+      title: "Onset",
+      dataIndex: "onset",
+    },
+    cardView
+      ? null
+      : {
+          render: (allergy) => (
+            <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
+          ),
+        },
+  ]);
--- a/src/components/content/allergies/patient-allergies-column.tsx
+++ b/src/components/content/allergies/patient-allergies-column.tsx
@@ -24,6 +24,10 @@ export const patientAllergiesColumns: TableColumn<AllergyModel>[] = [
     dataIndex: "onset",
   },
   {
+    title: "Manifestations",
+    dataIndex: "manifestations",
+  },
+  {
     render: (allergy) => (
       <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
     ),
--- a/src/components/content/allergies/patient-allergies-column.tsx
+++ b/src/components/content/allergies/patient-allergies-column.tsx
@@ -3,6 +3,10 @@ import { AllergyModel } from "@/fhir/models/allergies";
 
 export const patientAllergiesColumns: TableColumn<AllergyModel>[] = [
   {
+    title: "Onset",
+    dataIndex: "onset",
+  },
+  {
     title: "Description",
     dataIndex: "display",
   },
@@ -13,9 +17,5 @@ export const patientAllergiesColumns: TableColumn<AllergyModel>[] = [
   {
     title: "Category",
     dataIndex: "categories",
-  },
-  {
-    title: "Onset",
-    dataIndex: "onset",
   },
 ];
--- a/src/components/content/allergies/patient-allergies-column.tsx
+++ b/src/components/content/allergies/patient-allergies-column.tsx
@@ -1,33 +1,21 @@
 import { TableColumn } from "@/components/core/table/table-helpers";
-import { ViewFHIR } from "@/components/core/view-fhir";
 import { AllergyModel } from "@/fhir/models/allergies";
-import { compact } from "@/utils/nodash";
 
-export const patientAllergiesColumns = (
-  isStackedView = false
-): TableColumn<AllergyModel>[] =>
-  compact([
-    {
-      title: "Description",
-      dataIndex: "display",
-    },
-    {
-      title: "Type",
-      dataIndex: "type",
-    },
-    {
-      title: "Category",
-      dataIndex: "categories",
-    },
-    {
-      title: "Onset",
-      dataIndex: "onset",
-    },
-    isStackedView
-      ? null
-      : {
-          render: (allergy) => (
-            <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
-          ),
-        },
-  ]);
+export const patientAllergiesColumns: TableColumn<AllergyModel>[] = [
+  {
+    title: "Description",
+    dataIndex: "display",
+  },
+  {
+    title: "Type",
+    dataIndex: "type",
+  },
+  {
+    title: "Category",
+    dataIndex: "categories",
+  },
+  {
+    title: "Onset",
+    dataIndex: "onset",
+  },
+];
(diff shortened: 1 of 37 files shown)


src/components/content/allergies/patient-allergies-column.tsx
```diff
@@ -1,35 +1,39 @@
...
+      title: "Type",
+      dataIndex: "type",
+    },
+    cardView
+      ? null
+      : {
+          title: "Status",
+          dataIndex: "clinicalStatus",
+        },
+    {
```
luthfib, line 24: Whats card view and why dont we show status in card view?

pmadria3:
does the button look a little offset? Or is it just the screenshot?

src/components/content/allergies/patient-allergies-column.tsx
```diff
@@ -1,35 +1,39 @@
...
-    dataIndex: "manifestations",
-  },
-  {
-    render: (allergy) => (
-      <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
-    ),
-  },
-];
+export const patientAllergiesColumns = (
+  cardView = false
```
pmadria3, line 7: Maybe rename this? I think you are referring to when we have it reduced by responsiveness to the small view right?

pmadria3: isStackedView is perfect!

src/components/content/allergies/patient-allergies-column.tsx
mrosata: See the comment above from @pmadria3. Card is probably more aptly "stacked".
Screen Shot 2023-01-30 at 11 44 59 AM

We don't want to show the status in that view because it looks too much like a list, and more than likely the status will always be "Active" so it degrades readability for no additional information _(in the stacked view at least)_. I didn't want to rip it from the table row view because it fit in the row ok, if down the road we learn that it's information we need, then we can add it back in.

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -19,28 +26,27 @@ export function PatientAllergies({ className }: PatientAllergiesProps) {
...
+        className={cx(
+          "ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white",
+          className,
+          {
+            "ctw-stacked": breakpoints.sm,
+          }
+        )}
+      >
         <Table
+          RowActions={breakpoints.sm ? viewRecordFHIR : undefined}
```
WesleyKapow, line 42: Clicking the row does what here? All of our other components show the details/history drawer. I recommend changing allergies to do the same.

mrosata:
> does the button look a little offset? Or is it just the screenshot? The column values look lifted compared to button maybe center the values?

Yeah, idk if you saw my similar comment in slack. It was like that in the off-site version as well _(see images)_. I haven't checked pixels, but don't think the button is offset incorrectly, the button is aligned the same as the row text, but then the text inside the button is aligned according to the padding of the button. A smaller font in the button might fix it, I didn't see any small button styles in ctw, and while not opposed to doing so, was a bit hesitant to add a unique button style.

Screen Shot 2023-01-30 at 12 02 38 PM

Screen Shot 2023-01-30 at 11 55 22 AM

src/components/content/allergies/patient-allergies-column.tsx
```diff
@@ -1,35 +1,39 @@
...
-  },
-  {
-    render: (allergy) => (
-      <ViewFHIR name="Allergy Resource" resource={allergy.resource} />
-    ),
-  },
-];
+export const patientAllergiesColumns = (
+  isStackedView = false
+): TableColumn<AllergyModel>[] =>
```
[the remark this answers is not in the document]
mrosata: Yeah we can just remove "Status" completely until we know if we need it or not. With the test data we've used it's always set to active, I put a screenshot of all columns being rendered in stack view below, it's a screenshot of my entire viewport, a card with 4 allergies would likely take up the whole vertical height of my MacBook.

It's possible that this is the only place we're doing this dynamically now, but we did this on meds before where columns were hidden based on size from requirements passed down from UX Design. It was Canvas driven iirc and shortly after making that change we migrated Canvas to that `ActionList` component which only displays in a stacked view.

Screen Shot 2023-01-30 at 12 26 14 PM

src/components/core/view-fhir.tsx
```diff
@@ -55,14 +60,16 @@ function FHIRDrawer({
...
       title={name}
       isOpen={isOpen}
       onClose={onClose}
       showCloseFooter
     >
       <Drawer.Body>
-        <ReactJason value={resource} theme={github} />
+        <div className="ctw-text-sm">
+          <ReactJason value={resource} theme={github} />
+        </div>
```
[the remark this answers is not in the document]
mrosata: That's what I tried first, they don't take `className` as a property to their component.

eric-gunther:
Made some changes to bring immunizations and allergies into closer alignment, @WesleyKapow this included some changes to the columns definition that I'm unsure if you have strong opinions on.

src/components/content/immunizations/patient-immunizations.tsx
```diff
@@ -17,15 +28,28 @@ export function PatientImmunizations({ className }: PatientImmunizationsProps) {
...
-      <Table
-        className="-ctw-mx-px !ctw-rounded-none"
-        showTableHead={false}
-        isLoading={patientImmunizationsQuery.isLoading}
-        records={patientImmunizationsQuery.data ?? []}
-        columns={patientImmunizationsColumns}
-        handleRowClick={handleRowClick}
-      />
-    </div>
+    <StackedWrapper className={cx("ctw-patient-immunizations", className)}>
```
WesleyKapow, line 31: I don't think this is needed. It's setting up a `ctw-stacked` but we do that below as well and we need to use breakpoints in `Table` props.

src/components/content/immunizations/patient-immunizations-columns.tsx
```diff
@@ -1,29 +1,32 @@
...
-  },
-];
+export const patientImmunizationsColumns = (
+  isStackedView = false
+): TableColumn<ImmunizationModel>[] =>
+  compact([
+    {
+      title: "Date",
+      dataIndex: "occurrence",
+    },
```
WesleyKapow, line 13: Why hide fhir button when stacked?

Why remove the column sizing? We'll want column sizing down the road. Without it, the table will set the column widths based on the content being rendered. This means that as you page or filter or sort, you may get different sized columns.

src/components/content/immunizations/patient-immunizations.tsx
```diff
@@ -17,15 +28,28 @@ export function PatientImmunizations({ className }: PatientImmunizationsProps) {
...
+        className={cx(
+          "ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white",
+          className,
+          {
+            "ctw-stacked": breakpoints.sm,
+          }
+        )}
+      >
+        <Table
+          RowActions={breakpoints.sm ? viewRecordFHIR : undefined}
```
WesleyKapow, line 44: Why not remove view fhir from columns and have it always be a row action?

mrosata: Was just trying to keep the main table as-is, but I think that would be fine. I'll make the change.

mrosata: @WesleyKapow actually I remember why I did this. I don't really like the RowActions being rendered on the big table.. we can leave it for now as I already pushed the change, but the horizontal space is taken up regardless of how it's added which only feels like a half victory.. it's nice we don't have to see all the buttons, but it feels weird to have the invisible column on a smushed table. Which is to say on some tables it looks fine, on others it looks strange. Immunizations look ok, allergies look awkward.

WesleyKapow: There shouldn't be an invisible column. We should use column width settings in the column definitions to set the size of the different columns.

In general I think we have been moving to a world where we have row actions on hover. This is what we do for conditions 2.0 where you can add/delete/archive.

src/components/content/immunizations/patient-immunizations.tsx
mrosata: I'd like to take the time someday to figure out the proper way to forward ref to or from `StackedWrapper` in such a way that we don't need to copy these 10 lines into every stacked component.
```
      <div
        ref={containerRef}
        className={cx(
          "ctw-border ctw-border-solid ctw-border-divider-light ctw-bg-white",
          className,
          {
            "ctw-stacked": breakpoints.sm,
          }
        )}
      >
```

WesleyKapow: Are these `ctw-stacked` classes even doing anything?

WesleyKapow: I made it responsive in storybook and if I toggle `ctw-stacked` or `ctw-patient-immunizations`, I see no change.

src/components/content/immunizations/patient-immunizations-columns.tsx
mrosata: We can add it in widths down the line. This probably isn't worth spending time on when we aren't even sure which columns are needed at the moment.

src/components/content/immunizations/patient-immunizations.tsx
mrosata: Unless something had changed, I believe the table is adding that `.ctw-stacked` class and the outer one doesn't effect it either way just due to how the class works. If the StackedWrapper could do more than just add a class though is what my hopeful vision of it would be, ie: it's able to tell your component that it's a small breakpoint, and what your component does with that knowledge is its business, ie: component A might just need that class to target, component B might need the explicit knowledge that it is or isn't small so it can inform a child component (like the table). Either way, it's down the road stuff and I don't want to congest these threads too much more.

WesleyKapow (approved):

PR #438, opened 2023-01-30
Ctw 624/filter in table

luthfib:
_Awaiting Mike's PR so that we can consolidate the work, in this case the filter pills._ 

The only issue is one type error in filters file.


luthfib:
Closing and making new pr to resolve feature

PR #441, opened 2023-01-30
Ctw 474/in progress info

pmadria3:
CTW-474

This PR adds in-progress/initialize loading to the form drawer. 
Screenshot 2023-01-30 at 4 52 50 PM

[CTW-474]: https://zeushealth.atlassian.net/browse/CTW-474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/content/patient-history/patient-history-message-status.tsx
```diff
@@ -0,0 +1,42 @@
...
+            className="ctw-h-5 ctw-flex-none ctw-fill-caution-icon"
+          >
+            <path
+              d="M304 48c0-26.5-21.5-48-48-48s-48 21.5-48 48s21.5 48 48 48s48-21.5 48-48zm0 
+            416c0-26.5-21.5-48-48-48s-48 21.5-48 48s21.5 48 48 48s48-21.5 48-48zM48 304c26.5 0 48-21.5 48-48s-21.5-48-48-48s-48
+             21.5-48 48s21.5 48 48 48zm464-48c0-26.5-21.5-48-48-48s-48 21.5-48 48s21.5 48 48 48s48-21.5 48-48zM142.9 437c18.7-18.7
+              18.7-49.1 0-67.9s-49.1-18.7-67.9 0s-18.7 49.1 0 67.9s49.1 18.7 67.9 0zm0-294.2c18.7-18.7 18.7-49.1 0-67.9S93.7 56.2 
+              75 75s-18.7 49.1 0 67.9s49.1 18.7 67.9 0zM369.1 437c18.7 18.7 49.1 18.7 67.9 0s18.7-49.1 0-67.9s-49.1-18.7-67.9 0s-18.7 49.1 0 67.9z"
+            />
+          </svg>
```
luthfib, line 29: Can this go into its own icon file?

src/fhir/formatters.ts
```diff
@@ -108,3 +108,14 @@ export function formatAge(age: fhir4.Age): string {
   const { value, unit } = age;
   return compact([value, unit]).join(" ");
 }
+
+// Formats a string ISO date to MM/DD/YYY.
+export function formatStringISOtoDate(dateStr?: string): string | undefined {
```
luthfib, line 113: Dont we already like this with `formatStringToDate`?

pmadria3: So we do but this format is different - that one formats a string yyyymmdd or yyyyMMddHHmmss into MM/DD/YYYY. This one is in ISO form with dashes.

src/services/patient-history/patient-history.ts
```diff
@@ -33,18 +33,55 @@ export async function getPatientRefreshHistoryMessages(
...
+        patientHistoryExists: true,
+        status: "error",
+        dateCreated: messages[0]._createdAt,
+      };
+    default:
+      return {
+        patientHistoryExists: false,
+        status: "",
+        dateCreated: messages[0]._createdAt,
+      };
```
luthfib, line 85: This switch seems redundant. you dont really need `patientHistoryExists` as you are always returning true as long as it has a status and you are always setting status to `messages[0].status`

luthfib (commented):
Looking, good but left some comments

src/services/patient-history/patient-history.ts
```diff
@@ -33,18 +33,56 @@ export async function getPatientRefreshHistoryMessages(
...
+        ...lastPatientHistoryRequest,
+      };
+    case "initialize":
+      return {
+        patientHistoryExists: false,
+        ...lastPatientHistoryRequest,
+      };
+    case "error":
+      return {
+        patientHistoryExists: true,
```
WesleyKapow, line 78: true for error?

src/services/patient-history/patient-history.ts
```diff
@@ -33,18 +33,56 @@ export async function getPatientRefreshHistoryMessages(
...
+    case "in_progress":
+      return {
+        patientHistoryExists: false,
+        ...lastPatientHistoryRequest,
+      };
+    case "initialize":
+      return {
+        patientHistoryExists: false,
+        ...lastPatientHistoryRequest,
+      };
```
WesleyKapow, line 75: These are the same ya? You can take advantage of switch fallthrough:

```
case "in_progress":
case "initialize":
  return ...
```

luthfib (dismissed):
LGTM

src/services/patient-history/patient-history.ts
```diff
@@ -33,18 +33,56 @@ export async function getPatientRefreshHistoryMessages(
...
+
+  switch (messages[0].status) {
+    case "done":
+      return {
+        patientHistoryExists: true,
+        ...lastPatientHistoryRequest,
+      };
+    case "in_progress":
+      return {
+        patientHistoryExists: false,
```
WesleyKapow, line 68: How is `patientHistoryExists` used? Couldn't there be another message that has it as `done`?

.changeset/soft-trees-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Add progress bar in patient history drawer.
```
luthfib, line 5:
```suggestion
Add  in-progress state in patient history drawer.
```

src/fhir/formatters.ts
WesleyKapow, line 113: nit: name suggestion to follow existing pattern: `formatISODateStringToDate`

.changeset/soft-trees-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Add  in-progress state in patient history drawer.
```
luthfib, line 5:
```suggestion
Add in-progress banner in patient history drawer.
```

src/components/content/patient-history/patient-history-message-status.tsx
```diff
@@ -0,0 +1,31 @@
...
+            <div className="ctw-text-caution-message">
+              In Progress - request received {formatStringISOtoDate(date)}
+            </div>
+          </div>
+        </div>
+      );
+    case "done":
+    case "error":
+    default:
+      return <div />;
```
WesleyKapow, line 29: Do we actually want an empty div or could this return `null` instead?

src/components/content/patient-history/patient-history-message-status.tsx
```diff
@@ -0,0 +1,31 @@
...
+  status,
+  date,
+}: PatientHistoryStatusProps) => {
+  switch (status) {
+    case "initialize":
+    case "in_progress":
+      return (
+        <div className="ctw-alert-bg ctw-rounded-md ctw-bg-caution-bg">
+          <LoadingIcon className="ctw-h-5 ctw-flex-none ctw-fill-caution-icon" />
+          <div className="ctw-space-y-2">
```
WesleyKapow, line 19: Spacing class only works when there are 2 or more children. I think you can remove this div.

luthfib (dismissed):

src/components/content/conditions.tsx
```diff
@@ -351,10 +362,16 @@ export function Conditions({ className, readOnly = false }: ConditionsProps) {
       {patientResponse.data && (
         <PatientHistoryRequestDrawer
           header={
-            <div className="ctw-pt-0 ctw-text-base">
-              Request patient clinical history from 70K+ providers across the
-              nation. No changes will be made to your patient record.
-            </div>
+            <>
```
WesleyKapow, line 365: Should there be a set space between the two child divs? If so, you would set this to a div and add `space-y-*`.

WesleyKapow (dismissed):

src/services/patient-history/patient-history.ts
pmadria3: Yeah so there's a weird thing where there is a master status and then statuses of the two connectors (Surescripts and Commonwell) and so if one of them is an error but the other passes -> the master status is error. I met with michelle to see if we wanted to create some sort of map to make see what happens and she said to just focus on the master status and work off of that. So you can have a state where there is an error message that shows up but the data still shows in the table as one one connector passes and the other fails (as in the case with this test patient).

src/services/patient-history/patient-history.ts
pmadria3: They are the same but the `status` will be different for each of them depending on the case

pmadria3: one of the fields in the `lastPatientHistoryRequest`

src/services/patient-history/patient-history.ts
pmadria3: Not sure what this means but I use patientHistoryExists in the conditions to deal with null states and if it exists and when to show in `conditions.tsx`. status is what contains the `done` and is passed down. (In the future there will be more banners so we will need the date and status)

src/components/core/loading-icon.tsx
```diff
@@ -0,0 +1,21 @@
+export type LoadingIconProps = {
```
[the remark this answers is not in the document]
pmadria3: Confirmed from UX to have 2 separate ones and renaming to `ProgressIcon`

PR #447, opened 2023-02-01
CTW-473/status refresh

pmadria3:
CTW-473

This PR adds status message and Last Retrieved date (if there is one) to the other providers table. 
Screenshot 2023-02-01 at 12 44 32 PM

### DIFF
--- a/src/components/content/conditions.tsx
+++ b/src/components/content/conditions.tsx
@@ -289,7 +289,7 @@ export const Conditions = withErrorBoundary(
           </div>
           <div className="ctw-space-y-3">
             {(patientHistoryInfo?.status === "in_progress" ||
-              patientHistoryInfo?.status === "error") && (
+              patientHistoryInfo?.status === "initialize") && (
               <PatientHistoryStatus
                 status={patientHistoryInfo.status}
                 date={patientHistoryInfo.dateCreated}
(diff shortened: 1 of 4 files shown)


mrosata (approved):
lgtm

src/components/content/conditions.tsx
```diff
@@ -287,17 +288,32 @@ export const Conditions = withErrorBoundary(
             />
           </div>
           <div className="ctw-space-y-3">
+            {(patientHistoryInfo?.status === "in_progress" ||
+              patientHistoryInfo?.status === "initialize") && (
+              <PatientHistoryStatus
+                status={patientHistoryInfo.status}
+                date={patientHistoryInfo.dateCreated}
+              />
+            )}
```
WesleyKapow, line 297: Why is this guard needed, won't `PatientHistoryStatus` take care of rendering or not rendering something for us?

pmadria3: ah yes you are right that was the whole point of the switch lol

src/components/content/conditions.tsx
```diff
@@ -293,33 +293,32 @@ export const Conditions = withErrorBoundary(
...
-                  )}
+              {shouldShowClinicalHistoryArea &&
+                !readOnly &&
+                !hideRequestRecords && (
+                  <div className="ctw-flex ctw-items-baseline ctw-space-x-2">
+                    {patientHistoryInfo?.lastRetrievedAt && (
+                      <div className="ctw-text-sm ctw-italic ctw-text-black">
+                        Last Retrieved {patientHistoryInfo.lastRetrievedAt}
+                      </div>
+                    )}
```
WesleyKapow, line 310: Eh this isn't quite right, we can show "Last Retrieved" even if it is read only and even if we are hiding the request records link.

WesleyKapow (changes requested):

src/components/content/conditions.tsx
pmadria3: Ah good catch - moved it outside

luthfib (approved):

WesleyKapow (approved):

PR #448, opened 2023-02-01
CTW-629/Allow custom "Add to Record" handler

mrosata:
For overriding the opening of drawer to add new med when using "Add to Record" just pass a function to prop`handleAddToRecord` on the `PatientMedicationsTabbed` component.

Med tables:
- RowActions instead of ellipsis
- Utilizing onRowClick to show history
- Lowercase tabs and buttons (but capitalized using tailwind ctw-capitalize)
- Stacked view only shows medication name and dosage
https://www.loom.com/share/efa7a54acf5f412c981b63bb2a666220

### DIFF
--- a/.changeset/new-books-impress.md
+++ b/.changeset/new-books-impress.md
@@ -1,5 +1,0 @@
----
-"@zus-health/ctw-component-library": patch
----
-
-Add new usePatientHistory hook which can be used to open the patient history request drawer from anywhere.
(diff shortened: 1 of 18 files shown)


eric-gunther (approved):
Tested locally, works great. Per our discussion, we believe it may be better to just implement these changes in the `PatientMedications` component as opposed introduced a new `PatientMedicationsTabbed` component. Canvas is the only consumer of `PatientMedications` and we have control over when that gets upgraded. Long term we don't want to have 2 different medications, I think we want to push this design as the default starting point.

PR #451, opened 2023-02-02
Simplify authTokenURL usage

WesleyKapow:
We now decode the access token to compute expiresAt. Also updated README and comments to encourage use of authTokenURL and not accessToken.


luthfib (approved):
LGTM.

PR #455, opened 2023-02-02
CTW-632/FilterBox

mrosata:
Screen Shot 2023-02-06 at 4 01 22 PM

This PR adds the `FilterBar` component. It also sets that component up for medications table. The image above shows the table with dismissed medications. The row grayed out is the dismissed row.

--- 
> Note: no need to use capital letters on any button text or options text, they will capitalize from CSS. This is because not all customers want title case on buttons so we should just start controlling casing via CSS for everything (`.ctw-capitalize`).


src/components/content/medications/patient-medications-tabbed.tsx
```diff
@@ -19,30 +20,99 @@ export type PatientMedicationsTabbedProps = {
...
+    }) => (
       <>
-        <OtherProviderMedsTable handleAddToRecord={handleAddToRecord} />
+        <FilterBar
+          className="-ctw-mt-2"
+          filters={[
+            {
+              key: "dismissed",
+              type: "tag",
+              icon: "eye",
```
luthfib, line 79: I wonder if instead of having a switch we could just render jsx here?

src/fhir/medications.ts
```diff
@@ -179,12 +179,8 @@ export function splitMedications(
...
   const otherProviderMedications = summarizedMedications.filter(
     (medication) =>
-      !(
-        medication.isArchived ||
-        medication.status !== "active" ||
-        builderOwnedMedications.some(
-          (builderMed) => builderMed.rxNorm === medication.rxNorm
-        )
+      !builderOwnedMedications.some(
+        (builderMed) => builderMed.rxNorm === medication.rxNorm
```
mrosata, line 183: This is so we can instead filter these meds out in the component rather than from the query.

src/fhir/models/medication-statement.ts
```diff
@@ -141,6 +141,10 @@ export class MedicationStatementModel extends FHIRModel<fhir4.MedicationStatemen
     return !!this.getBasicResourceByAction("archive");
   }
 
+  get isInactive(): boolean {
+    return !["active", "intended", "unknown"].includes(this.status);
```
mrosata, line 145: This seems reasonable to me but would hear other suggestions or advice.

src/components/content/medications/other-provider-meds-table.tsx
```diff
@@ -123,15 +137,16 @@ export const OtherProviderMedsTable = withErrorBoundary(
...
 
 export const BadgeOtherProviderMedCount = () => {
-  const { otherProviderMedications } = useQueryAllPatientMedications();
-  if (
-    isArray(otherProviderMedications) &&
-    otherProviderMedications.length > 0
-  ) {
+  const { otherProviderMedications = [] } = useQueryAllPatientMedications();
+  const activeUnarchivedMedications = otherProviderMedications.filter(
+    (medication) => !(medication.isArchived || medication.isInactive)
```
luthfib, line 142: Not sure if this is going to be updated or the criteria will be but sometimes I think its nice to have this a const somewhere, probably above that we can easily update without having try to to figure out where the logic is

luthfib: Not a blocker or anything though, just suggestion

src/fhir/models/medication-statement.ts
luthfib: This is probably fine, not sure how many states you have in meds. I know we have a map for all the level 1 and level 2 statuses, and I kinda liked how we did it there, but if its not as complicated on the meds side then this is fine

src/components/core/filter-bar/filter-bar.tsx
```diff
@@ -0,0 +1,188 @@
...
+  }, [filters]);
+
+  const resetAllFilters = () => {
+    setActiveFilterKeys([]);
+    setActiveFilterValues({});
+    handleOnChange({});
+  };
+
+  // Add or remove a filter from the activated filters list
+  const addRemoveFilter = (key: string, remove = false) => {
```
luthfib, line 84: nit: rename to `addOrRemoveFilter`, saw this used elsewhere and thought it was something else

src/components/core/filter-bar/filter-bar-pills/select-pill.tsx
```diff
@@ -0,0 +1,75 @@
...
+            item === clearButton,
+        }),
+      }))}
+      onChange={(index, item) => {
+        if (item.key === "reset") {
+          onRemove();
+        } else if (item.key === "clear") {
+          onChange("", false);
+        } else {
+          onChange(item.key, true);
```
luthfib, line 70: This is good, however eventually I think it would be nicer if we could just pass an onChange to this component instead of defining it here

src/components/core/filter-bar/filter-bar.tsx
```diff
@@ -0,0 +1,188 @@
...
+      </div>
+
+      <ListBox
+        useBasicStyles
+        btnClassName="ctw-capitalize ctw-bg-transparent ctw-rounded ctw-text-content-light ctw-my-2 ctw-py-2 ctw-px-3 ctw-flex"
+        optionsClassName="ctw-capitalize"
+        items={inactiveFilterMenuItems}
+        onChange={(index, item) => {
+          if (item.key === "_clear") {
+            resetAllFilters();
```
luthfib, line 178: Should this be clear all filters as we already have the reset lingo meaning something else?

.changeset/kind-peaches-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
```
luthfib, line 2: This normally would definitely be a minor instead of a patch, but not a blocker

src/fhir/models/medication-statement.ts
mrosata: For meds we typically do `medication.status === "active"` for determining if is active. The full list of statuses are
```ts
('active'|'completed'|'entered-in-error'|'intended'|'stopped'|'on-hold'|'unknown'|'not-taken');
```
so I was thinking, since we are looking at the latest medication statement for that med:

| status | active | inactive |
|-------|-------|---------|
| active | true | false |
| completed | false | true |
| entered-in-error | false | true |
| intended | ? | ? | 
| stopped | false | true |
| on-hold | false | true |
| unknown | ? | ? | 
| not-taken | false | true |

Assumption is `active` is always active but never inactive, `unknown` is neither active or inactive and `intended` is likewise neither because there is a plan to go active with it but it's not there. I'll tag @eric-gunther on this also, does this logic make sense for `isInactive`?

src/components/core/filter-bar/filter-bar.tsx
mrosata: made the change

.changeset/kind-peaches-walk.md
mrosata: semver pre version `1.0.0` majors are minors and minors/patches are both generally treated as minors.

Sometimes I forget this though and treat a minor as a minor all the same, but I've noticed our minor number growing a lot so I would like to try to keep minors and patches as patches until we release `1.0.0`. Can discuss it more in PL maybe.

src/fhir/models/medication-statement.ts
luthfib: Is there only Inactive and Active as master statuses? I assume we wont show the end user `intended`. I know for Conditions we have quite a few like 'Pending', looks like:

image

luthfib: Just to double check what the end users sees.

mrosata: yeah I think we're just going to need to know inactive for filtering inactive records for the inactive meds table (or for an inactive filter)

eric-gunther:
Local testing looks great. Only issue I see is that when you click on a dismissed medication you still see the "Dismiss" button in the drawer. I know "un-dismiss" wasn't in scope for this PR, and the extra dismiss button doesn't do any harm (though I didn't check if it's doing nothing or creating a redundant Basic resource indicating another dismiss), but if it's feasible I would just remove it from the drawer for dismissed meds.

https://user-images.githubusercontent.com/97455910/217101660-c3581a8f-8fa6-4c29-9ec7-e1f13c46d569.mov

mrosata:
> Local testing looks great. Only issue I see is that when you click on a dismissed medication you still see the "Dismiss" button in the drawer. I know "un-dismiss" wasn't in scope for this PR, and the extra dismiss button doesn't do any harm (though I didn't check if it's doing nothing or creating a redundant Basic resource indicating another dismiss), but if it's feasible I would just remove it from the drawer for dismissed meds.
> 
>  Screen.Recording.2023-02-06.at.5.25.15.PM.mov

all set

src/fhir/models/medication-statement.ts
eric-gunther: Our intention is to show builders their "active" meds and their meds that "are not active" (hence "inactive"). So this implies that "unknown" and "intended" should count as "inactive". While those specifically seem like they could be an argument, I think exposing them in the inactive list is better than hiding them altogether.

eric-gunther: My initial reaction was the "inactive" == "all meds that are not active", but I actually like your logic better. I think showing "intended" and "unknown" alongside the active meds is a pretty reasonable assumption.

I think you should merge this PR with this assumption, but I also think you should poke Aaron and let him know this is an assumption you made. He may have a different suggestion that you can implement in a separate PR.

eric-gunther (approved):

PR #456, opened 2023-02-03
CTW-737 Adding managing org to med statement in history

eric-gunther:
Added patient managing organization to med statements in history. The majority of our med statements are produced either via CommonWell or Canvas. This will show the CommonWell org that we pulled the data from or "Firsthand" in the case of Canvas (though I need a follow-up task to ensure that managing org is set on firsthand patients, I'm not confident that's true). 

image

### DIFF
--- a/.changeset/chilly-lies-obey.md
+++ b/.changeset/chilly-lies-obey.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Adding patient managing organization to medication statement cards in history
+Add patient managing organization to medication statement cards in history.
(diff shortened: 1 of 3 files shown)


.changeset/chilly-lies-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Adding patient managing organization to medication statement cards in history
```
luthfib, line 5:
```suggestion
Add patient managing organization to medication statement cards in history.
```

luthfib (approved):

PR #457, opened 2023-02-03
lower required version of node

eric-gunther:
Customer running into a problem with our component library requiring node >= 16. The dev doing the work was able to install locally but not in the container for the test env they are setting up. They are running node 12.22. They could potentially upgrade but I'm not sure it's really necessary.


luthfib:
We need a changset

luthfib (approved):

pmadria3 (approved):

PR #469, opened 2023-02-08
Ctw 624/filter my patient record other records

luthfib:
Adds filters to conditions.

image


luthfib:
Not too sure why I get conflicts here but not locally

PR #471, opened 2023-02-08
CTW-756 - Fix FilterBar types, language and edge cases

mrosata:
Some finishing touches/fixes on FilterBar
- "Clear all filters" is now "Remove all filters" and has an icon.
Screen Shot 2023-02-08 at 2 55 13 PM
- Individual active filters have reset and remove actions with icons now. 
Screen Shot 2023-02-08 at 2 56 01 PM
- Modified some types to support the above
- Improved some types on the FilterChangeEvent
- Added prescriber medication to filters
Screen Shot 2023-02-08 at 5 06 43 PM

### DIFF
--- a/.changeset/hot-needles-swim.md
+++ b/.changeset/hot-needles-swim.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Fixed types and text on the filter bar. Also added icons where missing.
--- a/.changeset/hot-needles-swim.md
+++ b/.changeset/hot-needles-swim.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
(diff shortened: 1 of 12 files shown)


eric-gunther (approved):
Changed grammar on changeset to align with how folks have been doing it (not usually something I care too much about but was literally just talking to Luthfi about it).

Otherwise looks reasonable to me and filters still work great locally.

pmadria3:
Looks good to me - what edge cases did you handle?

mrosata:
> Looks good to me - what edge cases did you handle?

It turned into less of an edge case and more a bad type that was fixed and making the value of the FilterBar change event a little smarter so the developer adding FilterBar to their component doesn't need to do as many type checks.

luthfib (approved):

PR #474, opened 2023-02-08
CTW-733 Styling fixes for medications v2

eric-gunther:
Addressing some vertical padding that made Canvas look awkward and fixing issue with scrollbar always appearing in stacked table.

Kind of along/awkward video, but just showing how changing these styles when the component is in Canvas fixes the issues.

https://user-images.githubusercontent.com/97455910/217655876-f0d139c8-3057-43ae-a18f-16a306abdd94.mov

### DIFF
--- a/src/components/core/main.scss
+++ b/src/components/core/main.scss
@@ -20,7 +20,7 @@
   }
 
   .ctw-body-container {
-    @apply ctw-m-3 ctw-space-y-5 ctw-bg-bg-white;
+    @apply ctw-mx-3 ctw-space-y-5 ctw-bg-bg-white;
   }
 
   .ctw-border-default {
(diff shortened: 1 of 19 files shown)


src/components/core/table/table.scss
```diff
@@ -115,10 +115,6 @@
         td:not(.ctw-table-full-length-row):not(.ctw-table-row-actions) {
           @apply ctw-block ctw-w-full ctw-p-0 ctw-pr-7;
         }
-
-        td.ctw-table-row-actions {
-          height: inherit;
-        }
```
eric-gunther, line 121: This fixes the issue with a scrollbar always appearing in tables that reach the breakpoint triggering a stacked table and contain action buttons. Removing this does not seem to cause any unintended consequences and it definitely fixes the problem.

src/components/core/main.scss
```diff
@@ -20,7 +20,7 @@
   }
 
   .ctw-body-container {
-    @apply ctw-m-3 ctw-space-y-5 ctw-bg-bg-white;
```
eric-gunther, line 23: Only adding horizontal margin so that vertical spacing looks better.

luthfib (approved):

PR #489, opened 2023-02-10
Fix/prevent tree shaking CSS

mrosata:
SideEffects set to prevent css from being tree-shaken by webpack.
Upgraded storybook to 7.0.0-beta.45


mrosata:
fyi, built and ran storybook tests locally with success

luthfib (approved):

PR #492, opened 2023-02-13
624/filter my patient record other records

luthfib:
Add filters to conditions. 

image

### DIFF
--- a/src/components/content/conditions/patient-conditions-actions.tsx
+++ b/src/components/content/conditions/patient-conditions-actions.tsx
@@ -36,7 +36,7 @@ export function PatientConditionsActions({
   const patientHistory = usePatientHistory();
 
   return (
-    <div className="ctw-flex ctw-items-center ctw-justify-between ctw-space-x-2 ctw-p-3">
+    <div className="ctw-flex ctw-items-center ctw-justify-between ctw-space-x-2 ctw-p-3 sm:ctw-pt-1.5">
       <div className="ctw-flex">
         <SortButton
           options={sortOptions}
--- a/src/components/content/conditions/patient-conditions-actions.tsx
+++ b/src/components/content/conditions/patient-conditions-actions.tsx
@@ -1,20 +1,18 @@
 import { useAddConditionForm } from "./condition-hooks";
-import { Toggle } from "@/components/core/toggle";
+import { FilterBar } from "@/components/core/filter-bar/filter-bar";
 
 export type PatientConditionsActionsProps = {
   hideAdd: boolean;
-  onToggleShowHistoric: () => void;
 };
 
 export function PatientConditionsActions({
   hideAdd,
-  onToggleShowHistoric,
 }: PatientConditionsActionsProps) {
   const showAddConditionForm = useAddConditionForm();
 
   return (
     <div className="ctw-flex ctw-items-center ctw-justify-end ctw-space-x-2 ctw-p-3">
-      <Toggle name="historic" text="Historic" onChange={onToggleShowHistoric} />
+      <FilterBar filters={filterItems} handleOnChange={setFilters} />
       {!hideAdd && (
         <button
           type="button"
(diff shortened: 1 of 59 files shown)


src/components/content/conditions/patient-conditions-actions.tsx
```diff
@@ -5,60 +5,67 @@ import { useAddConditionForm } from "./condition-hooks";
...
   return (
-    <div className="ctw-flex ctw-items-center ctw-justify-between ctw-space-x-2 ctw-p-3 sm:ctw-pt-1.5">
-      {patientHistory.lastRetrievedAt && hideAdd && (
-        <PatientHistoryTableHeaderMessage
-          patientHistory={patientHistory}
-          message="Last Retrieved"
-        />
-      )}
-      <div>
+    <div className="ctw-flex ctw-items-center ctw-justify-between ctw-space-x-2 ctw-p-3">
```
mrosata, line 39: The `sm:ctw-pt-1.5` class attempts to make the x-space between tabs -> actions -> content even on larger displays
```suggestion
    <div className="ctw-flex ctw-items-center ctw-justify-between ctw-space-x-2 ctw-p-3 sm:ctw-pt-1.5">
```

With:
Screen Shot 2023-02-13 at 12 47 20 PM

Without _(extra vertical white space above actions):
Screen Shot 2023-02-13 at 12 47 54 PM

src/components/content/conditions/patient-conditions-filters.tsx
```diff
@@ -1,20 +1,66 @@
...
 
   function updateFilters(newFilters: Partial<Filters>) {
-    setFilters(merge(cloneDeep(filters), newFilters));
+    setFilters(() => ({
+      ...filters,
+      [collection]: {
+        ...newFilters,
+      },
+    }));
+  }
```
mrosata, line 45: Not a blocker, can leave this as-is. This looks like a good candidate for set
```suggestion
    setFilters(set(collection, newFilters, filters));
```
where set is from lodash fp _(non-fp set is mutative)_,`import { set } from "@/utils/nodash/fp"`.

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -89,10 +91,16 @@ export const PatientConditions = withErrorBoundary(
...
-            updateFilters({ showHistoric: !filters.showHistoric })
-          }
           currentSorts={currentSorts[collection]}
+          filterItems={availableFilters(
+            getUnfilteredCollection(
+              patientConditions,
+              otherConditions,
+              collection
+            )
+          )}
```
mrosata, line 101: Also not a blocker, but isn't the whole `getUnfilteredCollection` abstraction just doing a ternary statement? 
```suggestion
          filterItems={availableFilters(
            collection === "patient" ? patientConditions : otherConditions
          )}
```
In which case we could also remove the entire `getFilteredCollection` function at bottom of file as well.

src/components/core/filter-bar/filter-bar-pills/checkbox-pill.tsx
```diff
@@ -63,11 +56,18 @@ export function FilterBarCheckboxPill({
       buttonClassName={cx(filter.className, buttonClassName)}
       onItemSelect={(item) => onChange(item.key, item.value)}
     >
-      <span className="ctw-font-medium ctw-text-content-black">
+      <div className="ctw-font-medium ctw-text-content-black">
```
mrosata, line 59: Because this is in a button and div is a block level element, it should remain a span. If we need the span to behave like a div then we could use `.ctw-block` or `.ctw-inline-block`. Sometimes browsers or linters give warnings if div is in a button.

luthfib: Good callout

mrosata (approved):

mrosata (approved):

PR #495, opened 2023-02-14
CTW-765 pushable issue

eric-gunther:
In some EHRs we have this issue where opening a CTW drawer causes the entire chart to shift. We had previously gotten around this by adding logic to temporarily remove this `pushable` class from the chart just while the drawer is opening. Now that we've reduced our footprint by simply dropping in our `PatientMedicationsTabbed` component we no longer have the flexibility to add this kind of workaround when opening our drawer.

This PR exposes some props on the medications component that can be used to execute a function when opening the history drawer. 

Longer term we should find a way to more generally expose these kinds of functions to run upon opening a drawer because they are obviously not specific to medications.

https://user-images.githubusercontent.com/97455910/218780622-83e3853c-ab07-4544-b63f-9fba17beefac.mov

### DIFF
--- a/src/components/content/medications-table-base.tsx
+++ b/src/components/content/medications-table-base.tsx
@@ -8,6 +8,11 @@ import {
 } from "@/components/core/table/table-helpers";
 import { useBreakpoints } from "@/hooks/use-breakpoints";
 
+export type MedsHistoryTempProps = Partial<{
+  onOpenHistoryDrawer: () => void;
+  onAfterOpenHistoryDrawer: () => void;
+}>;
+
 export type MedicationsTableBaseProps<T extends MinRecordItem> = {
   children?: ReactNode;
   className?: string;
--- a/src/components/content/medications/other-provider-meds-table.tsx
+++ b/src/components/content/medications/other-provider-meds-table.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from "react";
-import { MedicationDrawer } from "@/components/content/medication-drawer";
+import { useMedicationHistory } from "./medication-history-drawer"
 import { MedicationsTableBase } from "@/components/content/medications-table-base";
 import { AddNewMedDrawer } from "@/components/content/medications/add-new-med-drawer";
 import { Badge } from "@/components/core/badge";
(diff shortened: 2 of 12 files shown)


src/components/content/medications-table-base.tsx
```diff
@@ -8,6 +8,11 @@ import {
 } from "@/components/core/table/table-helpers";
 import { useBreakpoints } from "@/hooks/use-breakpoints";
 
+export type MedsHistoryTempProps = Partial<{
+  onOpenHistoryDrawer: () => void;
+  onAfterOpenHistoryDrawer: () => void;
+}>;
+
```
luthfib, line 15: Not blocking or anything, just interesting use of `Partial`. Because you are defining a new type essentially and then using Partial on it.

src/components/content/medications/other-provider-meds-table.tsx
```diff
@@ -51,8 +56,19 @@ export const OtherProviderMedsTable = withErrorBoundary(
       useQueryAllPatientMedications();
 
     function openHistoryDrawer(row: MedicationStatementModel) {
+      // Temp - onOpen and onAfterOpen should be side-effect free as
+      // they may be called after component unmounts. We added
+      // this to support a bug-fix workaround in canvas.
+      if (isFunction(onOpenHistoryDrawer)) {
+        onOpenHistoryDrawer();
+      }
```
luthfib, line 64: Is this check just an extra safety net? Since `onOpenHistoryDrawer` can only be a function

luthfib (approved):
LGTM, nit picks but you can ignore if and we can address later

luthfib:
Might want to add a changeset though

src/components/content/medications-table-base.tsx
mrosata: I may have created that initial typing to hit at anyone modifying that type that they should not add any required properties.

mrosata: We can move this type to the `src/components/content/medications/other-provider-meds-table.tsx` file now though.

src/components/content/medications/other-provider-meds-table.tsx
```diff
@@ -19,7 +22,7 @@ export type OtherProviderMedsTableProps = {
   sortColumn?: keyof MedicationStatementModel;
   sortOrder?: SortDir;
   records?: MedicationStatementModel[];
-};
+} & MedsHistoryTempProps;
```
mrosata, line 25: can move that type
```ts
export type MedsHistoryTempProps = Partial<{
  onOpenHistoryDrawer: () => void;
  onAfterOpenHistoryDrawer: () => void;
}>;
```
into this file/type as it's only used here now.

src/components/content/medications/other-provider-meds-table.tsx
mrosata: yeah typescript didn't like `openHistoryDrawer?.()`. Some types with optional properties typescript requires the extra type-check.

mrosata (approved):

src/components/content/medications/other-provider-meds-table.tsx
eric-gunther: Per comment above, using it in the other two tables that appear in the tabbed component.

PR #503, opened 2023-02-15
Datadog/environment names

mrosata:
1. Use Datadog prod account everywhere except from local development.
2. Use normalized environment names in Datadog (prod, dev).

### DIFF
--- a/src/utils/telemetry.ts
+++ b/src/utils/telemetry.ts
@@ -17,19 +17,23 @@ import {
 
 type TelemetryEventKey = "zusTelemetryClick" | "zusTelemetryFocus";
 
-const prodAccountConfig = {
+const prodDatadogConfig = {
   service: "ctw-component-library",
   clientToken: "pub7f1b01887ceb412fd989f5e08cf60d9a",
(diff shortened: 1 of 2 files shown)


src/utils/telemetry.ts
```diff
@@ -17,19 +17,23 @@ import {
...
 };
 
-const origin = typeof window !== "undefined" ? window.location.origin : "";
-const body = typeof window !== "undefined" ? window.document.body : undefined;
+let origin = "";
+let body: HTMLElement | undefined;
+if (typeof window !== "undefined") {
+  origin = window.location.origin;
+  body = window.document.body;
+}
```
mrosata, line 36: We haven't seen any issues surrounding this bit here, but essentially the point of the original
```ts
const origin = typeof window !== "undefined" ? window.location.origin : "";
const body = typeof window !== "undefined" ? window.document.body : undefined;
```
was to not access properties of window in non-browser environments. However, ternary statements are eagerly evaluated, hence the refactor.

luthfib (approved):

PR #511, opened 2023-02-16
Setup up Document component

pmadria3:
CTW-753

This PR adds the Document Component. It also dedupes the entries based on a variety of columns and sorted by date. Storybook tests are also available.

https://user-images.githubusercontent.com/98342697/220460308-c081b7d0-93a6-4613-aee0-6c1690884c5b.mov


src/fhir/search-helpers.ts
```diff
@@ -33,6 +33,40 @@ const UPI_TAGS = [`${SYSTEM_ZUS_UPI_RECORD_TYPE}|universal`];
 
 const SUMMARY_TAGS = [`${SYSTEM_SUMMARY}|Common`];
 
+// Returns the needed search params to filter resources down to those
+// pertaining to our patientUPID.
+function patientSearchParams(
```
WesleyKapow, line 38: Why move this up? I was going with a pattern of exports on top, helper functions on bottom.

src/fhir/models/document.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  get dateCreated(): string | undefined {
+    return this.resource.content[0].attachment.creation;
+  }
+
+  get managingOrganization(): string | undefined {
+    return this.resource.custodian?.display;
+  }
+
+  get SectionDisplays(): string[] | undefined {
```
WesleyKapow, line 34: Should not be capitlized.

src/fhir/models/document.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  get title(): string | undefined {
+    return this.resource.content[0].attachment.title;
+  }
+
+  get dateCreated(): string | undefined {
+    return this.resource.content[0].attachment.creation;
+  }
+
+  get managingOrganization(): string | undefined {
```
WesleyKapow, line 30: Maybe change to `custodian` ?

src/components/content/document/story-helpers/mocks/requests.ts
```diff
@@ -0,0 +1,36 @@
...
+    // less than the default testing-library timeout of 1000ms.
+    (req, res, ctx) => res(ctx.delay(750), ctx.status(200), ctx.json(patient))
+  );
+
+  const mockDocumentGet = rest.get(
+    "https://api.dev.zusapi.com/fhir/DocumentReference",
+    (req, res, ctx) => res(ctx.status(200), ctx.json(documents))
+  );
+
+  const mockConditionBinaryDocumentinDocumentReference = rest.get(
```
WesleyKapow, line 19: Check name here.

src/components/content/document/patient-document-columns.tsx
```diff
@@ -0,0 +1,40 @@
...
+      widthPercent: 30,
+      minWidth: 150,
+      title: "Managing Organization",
+      render: (document) => <div>{document.managingOrganization}</div>,
+    },
+  ];
+  if (includeViewFhirResource) {
+    documentColumns.push({
+      widthPercent: 10,
+      minWidth: 200,
```
WesleyKapow, line 32: This `minWidth` seems very wrong. This column is taking 10 % but requires a minimum of 200px. MinWidth's should be whatever is the minimum for the column to not look bad. It's a bit arbitrary, but it should roughly align (be proportional) with widthPercents.

src/components/content/document/patient-document-columns.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { DocumentModel } from "@/fhir/models/document";
+
+export const patientDocumentColumns = (includeViewFhirResource = true) => {
+  const documentColumns: TableColumn<DocumentModel>[] = [
+    {
+      widthPercent: 20,
+      minWidth: 150,
+      title: "Date Created",
+      render: (document) => (
+        <div>{formatISODateStringToDate(document.dateCreated)}</div>
```
WesleyKapow, line 13: We typically push formatting into the model itself. That way we don't need to format it everywhere we want to display it (like here and in the details). You can then use the raw date off of the resource for sorting purposes.

src/components/content/document/document-details-drawer.tsx
```diff
@@ -0,0 +1,103 @@
...
+    async function load() {
+      setIsLoading(true);
+      if (document.binaryID) {
+        setBinaryId(document.binaryID);
+      }
+      setIsLoading(false);
+    }
+
+    void load();
+  }, [binaryId, document, getRequestContext]);
```
WesleyKapow, line 53: Not sure what this is doing. You aren't using `getRequestConext` anywhere and you aren't loading anything.

WesleyKapow (changes requested):
From the list of things to do for every resource, there's three items remaining, the first of which is a MUST for this PR:

* MUST: Remember to export the new component!
* Unleash feature toggle
* CTW uses feature toggle and adds link to patient sidebar / nav

src/fhir/search-helpers.ts
pmadria3: Ah I moved it up because I couldn't find it and realized it was all the way at the bottom. I'll move it back down

pmadria3:
Created the feature toggle for Unleash, exported the new component, and working on PR for the CTW sidebar

src/components/content/document/patient-document-columns.tsx
```diff
@@ -3,33 +3,31 @@ import { ViewFHIR } from "@/components/core/view-fhir";
...
       title: "Managing Organization",
-      render: (document) => <div>{document.managingOrganization}</div>,
+      render: (document) => <div>{document.custodian}</div>,
     },
   ];
   if (includeViewFhirResource) {
     documentColumns.push({
-      widthPercent: 10,
-      minWidth: 200,
+      widthPercent: 20,
```
WesleyKapow, line 28: With this column added, we'll have `widthPercent` total = `120`, does that work out ok? Does it force scroll?

src/components/content/document/story-helpers/mocks/requests.ts
```diff
@@ -16,14 +16,14 @@ function mockDocumentRequests() {
     (req, res, ctx) => res(ctx.status(200), ctx.json(documents))
   );
 
-  const mockConditionBinaryDocumentinDocumentReference = rest.get(
+  const mockBinaryDocumentinDocumentReference = rest.get(
```
WesleyKapow, line 19: DocumentinDocument?

OH! `BinaryDocument IN DocumentReference`? The `i` should be capitalized then.

How about simply, `mockBinary`? This is only used here in this file, no need to get super verbose.

src/fhir/document.ts
```diff
@@ -19,7 +20,7 @@ export function usePatientDocument() {
         );
         return orderBy(
           applyDocumentFilters(documents),
-          [(document) => document.dateCreated || ""],
+          [(document) => formatDateLocalToISO(document.dateCreated) || ""],
```
WesleyKapow, line 23: Be explicit here and go with `document.resource.content[0].attachment.creation`. That decouples the sort code from the `dateCreated` code. As you have it now, if we changed `dateCreated`, sorting would break.

WesleyKapow (changes requested):

src/components/content/document/document-details-drawer.tsx
WesleyKapow: Ping.

WesleyKapow: Note: Make sure to not resolve comments unless you made a change or replied to the comment as to why nothing is needed.

src/components/content/document/document-details-drawer.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const documentData = (document: DocumentModel) => [
+  { label: "status", value: document.status },
+  { label: "docStatus", value: document.docStatus },
+  { label: "Managing Organization", value: document.custodian },
+  {
+    label: "Section Display",
+    value: document.sectionDisplays && (
+      <SimpleMoreList
+        items={document.sectionDisplays}
+        limit={30}
```
WesleyKapow, line 97: THIRTY!? That seems like a large limit.

pmadria3: So not entirely sure as there could be many sections (varies by document - not sure what the limit is so I set it super high)

WesleyKapow: The more list will limit the display of the array to `limit` and then have a `+ # more` text (not a link/button). Check with UX on what to display here!

src/components/content/document/document-details-drawer.tsx
pmadria3: Ah thanks for catching that - accidentally closed this one.

src/components/content/document/patient-document-columns.tsx
WesleyKapow: ^^ ?? You still have a set of columns that total more than 100 %.. .is that a problem? Please try it and report back.

WesleyKapow (changes requested):

src/components/content/document/document-details-drawer.tsx
```diff
@@ -35,11 +35,7 @@ export function DocumentDetailsDrawer({
   const [binaryId, setBinaryId] = useState<string | undefined>();
 
   useEffect(() => {
-    function load() {
-      setBinaryId(document.binaryID);
-    }
-
-    void load();
+    setBinaryId(document.binaryID);
   }, [binaryId, document]);
```
WesleyKapow, line 39: What is the purpose of `binaryId` state? Do we need that?

WesleyKapow (changes requested):

src/components/content/document/patient-document-columns.tsx
```diff
@@ -11,13 +11,13 @@ export const patientDocumentColumns = (includeViewFhirResource = true) => {
...
     {
-      widthPercent: 40,
+      widthPercent: 30,
       minWidth: 200,
       title: "Title",
       render: (document) => <div>{document.title}</div>,
     },
     {
-      widthPercent: 40,
+      widthPercent: 30,
```
WesleyKapow, line 20: OK now the concern is what happens when the set of column's don't add up to 100%. I tried this myself and it seems like it mostly works as you'd hope in that each column proportionally expands to take up the entire 100% width.

Little context here:
We set each column to have the `width: {widthPercent}; min-width: {minWidth}`. This way, each column grows with window size and takes the correct proportion (percent) of the horizontal space. The min-width, guarantees that the column doesn't look bad / ugly-wrapping, when the window size shrinks by causing the table to be scrollable (prior to changing to stack view).

src/components/content/document/patient-document-columns.tsx
```diff
@@ -0,0 +1,37 @@
...
+      minWidth: 200,
+      title: "Managing Organization",
+      render: (document) => <div>{document.custodian}</div>,
+    },
+  ];
+  if (includeViewFhirResource) {
+    documentColumns.push({
+      widthPercent: 20,
+      minWidth: 120,
```
WesleyKapow, line 29: `120` allows wrapping which is ugly:
Screenshot 2023-02-22 at 4 17 23 PM

Set this to `150` and that way we'll never get this wrapped button!

WesleyKapow (changes requested):

PR #517, opened 2023-02-21
Timeline updates

WesleyKapow:
CTW-741
CTW-740

* Added timeline table headers.
* Adjusted columns.
* Tweaked how encounter type is computed / what gets displayed.
* Bumped # of participants to display from 3 -> 6. This should handle the majority of cases. This was done to avoid the `+ # more` text, which will now appear above encounter location.

### Storybook
Screenshot 2023-02-21 at 1 37 54 PM

### DIFF
--- a/src/fhir/models/encounter.ts
+++ b/src/fhir/models/encounter.ts
@@ -73,7 +73,7 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
     // first type coding.
     const isNullFlavor = classSystem === SYSTEM_NULL_FLAVOR;
     const code = isNullFlavor ? typeCode : classCode;
-    const display = isNullFlavor ? typeText : classDisplay;
+    let display = isNullFlavor ? typeText : classDisplay;
 
     // Optionally add expanded version of the code
     // if code is AMB, IMP, EMER and display isn't already the expanded version.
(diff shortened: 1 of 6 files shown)


luthfib (dismissed):

src/fhir/models/encounter.ts
```diff
@@ -63,10 +63,42 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
...
+    // first type coding.
+    let code;
+    let display;
+    if (classSystem !== SYSTEM_NULL_FLAVOR) {
+      code = classCode;
+      display = classDisplay;
+    } else {
+      code = typeCode;
+      display = typeText;
+    }
```
mrosata, line 82: nonblocking, just an option to reduce the body of function:
```suggestion
    const isNullFlavor = classSystem === SYSTEM_NULL_FLAVOR;
    const code = isNullFlavor ? typeCode : classCode;
    const display = isNullFlavor ? typeText : classDisplay;
```

mrosata (approved):

src/fhir/models/encounter.ts
WesleyKapow: Done! Thanks for the suggestion!

mrosata (approved):

PR #519, opened 2023-02-21
Storybook Preperation for ZusAggregatedProfile

mrosata:
As I started the ZusAggregatedProfile I realized storybook would need refactoring. That snowballed into a good place, but also lots of files touched. So this PR I removed the ZusAggregatedProfile component and left in storybook and component changes to support the mega component.

- MSW mocks now cleanup appropriately when changing to different components and reset completely before a new story begins.
- Medications and Conditions both mock `Basic` resource, which is fine, now it's double fine as both mocks can exist at same time (as they will need to in ZusAggregatedProfile story). Having multiple mocks for the same URL is fine, returning undefined from a handler allows other handlers to run. So the conditions mock of "Basic" now returns undefined if the resource sent in the request isn't a condition (and vice-versa for medications component mocks)
- Removed header from allergies component.
- Made padding in conditions actions even on top and bottom (so meds and conditions will look the same in ZAP component and match figma).

There will be more style changes when the second ZAP component PR comes in. For now I wanted to get the storybook changes in as they are legion.

### DIFF
--- a/src/components/content/conditions/patient-conditions.tsx
+++ b/src/components/content/conditions/patient-conditions.tsx
@@ -1,7 +1,6 @@
 import cx from "classnames";
 import { useRef, useState } from "react";
 import { useConditionHistory } from "../condition-history/conditions-history-drawer";
-import { filterOtherConditions } from "./helpers";
 import { PatientConditionsActions } from "./patient-conditions-actions";
 import { patientConditionsColumns } from "./patient-conditions-columns";
 import {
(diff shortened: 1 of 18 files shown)


src/components/content/conditions/patient-conditions.tsx
```diff
@@ -82,17 +79,16 @@ export const PatientConditions = withErrorBoundary(
       <div
         ref={containerRef}
         className={cx(
-          "ctw-patient-conditions ctw-items-center ctw-justify-between ctw-py-5",
+          "ctw-patient-conditions ctw-items-center ctw-justify-between ctw-bg-white",
```
mrosata, line 82: @eric-gunther

mrosata:
Even though the requests are being cleaned up. I'm not 100% sure that the tests will always pass. Especially stories with lots of tests since despite this change, the mocks are setup when a story loads (not a test). Ideally, before each test mocks are cleared and recreated.

pmadria3 (approved):
Looks good and I changed my document story to match your changes as discussed

PR #523, opened 2023-02-23
Uniform tables and telemetry

mrosata:
Wraps tables in `.ctw-overflow-hidden` because the -1px x margin on tables will cause a bold lefthand border if the table touches the sides of its container (and the container has a border). This happened even if the left and right borders of the table were off.

Telemetry and error boundary wrapper on Allergies, Immunizations and Documents.

Removed these two class names `ctw-items-center ctw-justify-between` from component containers (besides old meds and old conditions) as the classes don't affect the layout at all. Must have been a relic of when more content was in the body of components.

Removed the title bar from Documents and Immunizations as the V2 components don't need titles. It's becoming more of a CTW pattern than a CTW Component Library pattern.

Removed the padding on Patients Table along with its tables left and right borders. See the image below, it is CTW, the nested borders are different styles and don't look like the rest of the site:
Screen Shot 2023-02-23 at 2 15 57 PM
So now they will look like this:
Screen Shot 2023-02-23 at 2 18 33 PM

For completeness, here are how the other components that were changed look:

https://www.loom.com/share/5044ddd4c0fa4fffb7e499da506cdc29

### DIFF
--- a/src/components/content/allergies/patient-allergies.tsx
+++ b/src/components/content/allergies/patient-allergies.tsx
@@ -1,5 +1,6 @@
 import { useRef } from "react";
 import { patientAllergiesColumns } from "@/components/content/allergies/patient-allergies-column";
+import { withErrorBoundary } from "@/components/core/error-boundary";
 import { Table } from "@/components/core/table/table";
 import { usePatientAllergies } from "@/fhir/allergies";
 import { useBreakpoints } from "@/hooks/use-breakpoints";
@@ -9,7 +10,7 @@ export type PatientAllergiesProps = {
   enableFqs?: boolean;
 };
 
-export function PatientAllergies({
+function PatientAllergiesComponent({
   className,
   enableFqs,
 }: PatientAllergiesProps) {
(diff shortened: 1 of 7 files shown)


src/components/content/allergies/patient-allergies.tsx
```diff
@@ -9,7 +10,7 @@ export type PatientAllergiesProps = {
   enableFqs?: boolean;
 };
 
-export function PatientAllergies({
+function PatientAllergiesComponent({
```
luthfib, line 13: Just curious behind the reasoning of adding "Component" to the end of the title

src/components/content/document/patient-documents.tsx
```diff
@@ -32,23 +32,27 @@ export function PatientDocuments({
...
-      <Heading title="Documents" />
-      <Table
-        stacked={breakpoints.sm}
-        className="-ctw-mx-px !ctw-rounded-none"
-        isLoading={isLoading}
-        records={document}
-        columns={patientDocumentColumns(includeViewFhirResource)}
-        handleRowClick={handleRowClick}
-      />
+      <div className="ctw-overflow-hidden">
```
luthfib, line 40: I guess doesnt have to be part of this pr, but feels odd that we have to add this class everywhere instead of it being an abstraction. Maybe part of a table, not sure.

luthfib (approved):
Some nit pick questions but otherwise looks good.

PR #524, opened 2023-02-23
Ctw 793/resource details drawer

WesleyKapow:
CTW-793

Add `ResourceDetailsDrawer` & refactor a bunch of resource components. This is a start towards setting up more re-usable, generic, resource components to use.

Drawer and associated hook provide a way to specify:
* header (displayed in big above details box)
* Optional subHeader (displayed under header)
* details (the contents of the details box)
* getSourceDocument (boolean to indicate if we should try and fetch source document, used by patient timeline).

Refactored resources that do not show history:
* Allergies
* Documents
* Immunizations
* Timeline / Encounters

Additionally:
* Added `resourceTypeTitle` to `FHIRModel`, this way all models have a default to use but it can be overwritten. `DocumentModel` overwrites it in order to return `Document` instead of the default `Document Reference`.
* Setup Allergy table to show details on row click.
* Fixed patient timeline in storybook to fetch provenance and binary. This was preventing the details from showing up.
* Fixed patient document table in storybook to show a source document from the details view

**Screenshots for reference**
Screenshot 2023-02-23 at 2 25 15 PM
Screenshot 2023-02-23 at 12 36 07 PM
Screenshot 2023-02-23 at 12 35 37 PM
Screenshot 2023-02-23 at 2 36 38 PM
Screenshot 2023-02-23 at 2 36 45 PM


pmadria3 (dismissed):
LGTM

src/fhir/models/document.ts
```diff
@@ -20,6 +20,11 @@ export class DocumentModel extends FHIRModel<fhir4.DocumentReference> {
     return this.resource.docStatus;
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  get resourceTypeTitle(): string {
+    return "Document";
+  }
```
pmadria3, line 26: is the thought process here just to have all in the model?

WesleyKapow: Mostly yes. It's a balance though. Like I didn't put `header` and `subHeader` into the model because that felt very specific to the details drawer.

pmadria3 (approved):

PR #531, opened 2023-02-27
CTW-776/ZusAggregatedProfile component

mrosata:
Add `ZusAggregatedProfile` component.

https://www.loom.com/share/0d86808594ef473c982648404397a49b


src/components/content/story-helpers/mocks/requests.ts
```diff
@@ -1,7 +1,8 @@
 import { rest } from "msw";
 import { binary } from "./binary";
 
-export const mockBinaryGet = rest.get(
-  "https://api.dev.zusapi.com/fhir/Binary/:BinaryId",
-  async (_, res, ctx) => res(ctx.status(200), ctx.json(binary))
-);
+export const mockBinaryGet = () =>
+  rest.get(
```
pmadria3, line 5: why make it a function? Was there something wrong with just doing mockBinaryGet = rest...?

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {
+  ProviderMedsTable,
+  ProviderMedsTableProps,
+} from "@/components/content/medications/provider-meds-table";
+import { ZAPResourceName } from "@/components/content/zus-aggregated-profile/zus-aggregated-profile";
+import { TabGroupItem } from "@/components/core/tab-group/tab-group";
+
+export type ZusAggregatedProfileTabs = Record<
+  ZAPResourceName,
+  (props: object) => TabGroupItem<unknown>
```
pmadria3, line 32: why is this unknown?

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export type ZAPResourceName =
+  | "allergies"
+  | "conditions"
+  | "conditions-outside"
+  | "documents"
+  | "immunizations"
+  | "medications"
+  | "medications-outside";
+
```
pmadria3, line 24: Potential Blocker :octagonal_sign: 

I don't see encounters anywhere - the ticket mentions `conditions, medications, allergies, immunizations, encounters, and documents.`

pmadria3:
Also while you are on it - could you also add CareTeams to the ZAP (last one 🥲 )

src/components/content/story-helpers/mocks/requests.ts
mrosata: Kinda just trying to follow the guidance of setup mocks when they are needed and cleanup when they aren't (then repeat). In a function we have to choice to mock whenever and however many times we like. Every-time someone calls that function it gets mocked. Otherwise, if we were to  reset mocks and then call `setup*Mocks` for a story that also uses this mock, it wouldn't be included.

I don't think this would likely cause issues, but most of the other mocks have migrated to this pattern so it's best not to start adding outliers.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+  args: {
+    resources: [
+      "allergies",
+      "conditions",
+      "conditions-outside",
+      "documents",
+      "immunizations",
+      "medications",
+      "medications-outside",
+      "timelines",
```
pmadria3, line 104: Last one but would you be able to add CareTeam as well? That should wrap up the ZAP.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
mrosata: Actually I think I can refactor that generic type out of TabGroupItem completely. Let me try that quick. Originally this was needed because TabGroupItem had a prop that was a function which could be like `(c: ConditionModel) => JSX` or `(m: MedicationModel) => JSX` but I think it's not needed anymore because we created `BadgeOtherProviderConditionsCount` and `BadgeOtherProviderMedicationsCount` components which eliminated the need for tabs to know anything about the underlying resource

src/components/content/zus-aggregated-profile/zus-aggregated-profile.stories.tsx
mrosata: sure

src/components/content/story-helpers/mocks/requests.ts
mrosata: Oh @pmadria3 ... the number one reason this is great.. is that we no longer need to restart storybook every time we add a mock.. I should have led with that.

luthfib (approved):

pmadria3 (dismissed):
LGTM

pmadria3 (approved):

PR #535, opened 2023-02-28
Refactor condition filters and simplify FilterBar

WesleyKapow:
Creating this first PR to keep the generic component work smallish.

FilterBar will now keep track fo the "recentlyAdded" filter and use that for setting `isOpen`. This way the component handles it and the passed in filters don't need to specify it.

Renamed FilterBar's `handleOnChange` to `onChange` to align with react conventions.

Refactored condition filters into filter-options.ts


luthfib (changes requested):
Getting the following error locally when going to patient conditions in storybook. 

image

luthfib (approved):

PR #538, opened 2023-03-01
Add request records in condition 2.0 standalone component

pmadria3:
This PR adds the Request Records button in conditions 2.0. 

https://user-images.githubusercontent.com/98342697/222229457-a2c66623-e6f3-4e45-aa1b-235c26197fd7.mov

### DIFF
--- a/src/components/content/conditions/patient-conditions-actions.tsx
+++ b/src/components/content/conditions/patient-conditions-actions.tsx
@@ -21,6 +21,7 @@ export type PatientConditionsActionsProps = {
   filterItems: FilterItem[];
   setFilters: (filterChangeEvent: FilterChangeEvent) => void;
   filters: FilterChangeEvent;
+  hideRequestRecords: boolean;
 };
 
 export function PatientConditionsActions({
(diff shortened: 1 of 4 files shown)


luthfib (approved):

src/components/content/conditions/patient-conditions-actions.tsx
```diff
@@ -51,13 +53,24 @@ export function PatientConditionsActions({
...
-        />
-      )}
+      <div className="ctw-flex ctw-space-x-2 ctw-p-1">
+        {patientHistory.lastRetrievedAt && hideAdd && (
+          <PatientHistoryTableHeaderMessage
+            patientHistory={patientHistory}
+            message="Last Retrieved"
+          />
+        )}
+        {hideAdd && !hideRequestRecords && (
```
WesleyKapow, line 63: Why only show when `hideAdd` is true?

pmadria3: So the hideAdd is making sure that the Request Records only shows up in the OtherProvider section otherwise it would show up in Conditionlist as well

PR #540, opened 2023-03-01
Fix request records logic 

pmadria3:
Fixing the logic to show up by default unless the prop is provided

### DIFF
--- a/src/components/content/conditions/patient-conditions.tsx
+++ b/src/components/content/conditions/patient-conditions.tsx
@@ -101,7 +101,7 @@ export const PatientConditions = withErrorBoundary(
           )}
           setFilters={updateFilters}
           filters={filters[collection]}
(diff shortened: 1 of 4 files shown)


WesleyKapow (approved):
Beautiful!

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -101,7 +101,7 @@ export const PatientConditions = withErrorBoundary(
           )}
           setFilters={updateFilters}
           filters={filters[collection]}
-          hideRequestRecords={hideRequestRecords}
+          hideRequestRecords={hideRequestRecords || collection === "patient"}
```
mrosata, line 104: Just verifying that this means we never allow hiding request records on collection "patient"?

PR #541, opened 2023-03-01
Patient search component

luthfib:
Add patient search component.

image

### DIFF
--- a/src/components/core/list-box/list-box.tsx
+++ b/src/components/core/list-box/list-box.tsx
@@ -1,7 +1,7 @@
 import { Listbox } from "@headlessui/react";
 import { CheckIcon, ChevronDownIcon } from "@heroicons/react/solid";
 import cx from "classnames";
-import React, { Fragment, useState } from "react";
+import { Fragment, useState } from "react";
 import type { ReactNode } from "react";
 import { FilterOptionSelect } from "@/components/core/filter-bar/filter-bar-types";
 import { isFunction } from "@/utils/nodash";
@@ -20,6 +20,7 @@ export type ListBoxProps<T> = {
   onChange: (index: number, item: T) => void;
   optionsClassName?: cx.Argument;
   useBasicStyles?: boolean;
+  isOpen?: boolean;
 };
 
 export type ListBoxOptionStatus = {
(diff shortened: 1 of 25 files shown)


eric-gunther (approved):
Looks good to me, in the interest of urgency for this issue will approve now, but I have 3 follow-up thoughts:

1. We still need to remove the unnecessary `_tag` parameter in searches, but perhaps you have a separate branch for that.
2. Should include in storybook.
3. I think we should show search results with DOB and any non-Zus identifiers, and we should support search by identifier (ie. just updating getBuildersPatientList to search with both the `name` and `identifier` search parameters). I think this is something UX will want to weigh in on so going to put this in Slack.

src/components/core/list-box/list-box.tsx
```diff
@@ -52,58 +54,70 @@ export function ListBox<T extends MinListBoxItem>({
...
+          <>
+            <Listbox.Button
+              className={cx(
+                btnClassName,
+                !useBasicStyles && "after:ctw-bg-content-black",
+                !useBasicStyles &&
+                  "focus-visible:ctw-outline-primary-dark focus-visible:after:ctw-bg-transparent",
+                "ctw-relative ctw-mr-2 ctw-cursor-pointer ctw-border-0 ctw-border-transparent"
+              )}
+            >
```
mrosata, line 67: Does our abstraction of ListBox not work here style-wise? If the styles are diff and it doesn't, that's fine, but if it's the same styles already as the abstraction we should just use`src/components/core/list-box/list-box.tsx`

mrosata: otherwise pr looks good

pmadria3 (approved):
LGTM

luthfib:
> Looks good to me, in the interest of urgency for this issue will approve now, but I have 3 follow-up thoughts:
> 
> 1. We still need to remove the unnecessary `_tag` parameter in searches, but perhaps you have a separate branch for that.
> 2. Should include in storybook.
> 3. I think we should show search results with DOB and any non-Zus identifiers, and we should support search by identifier (ie. just updating getBuildersPatientList to search with both the `name` and `identifier` search parameters). I think this is something UX will want to weigh in on so going to put this in Slack.

1. Yes I have separate pr for that.
2. Agreed. 
3. My idea was to get this in and quickly follow-up with a pr to add that. Only reason being I have to add a prop to add a custom render to the each dropdown field.

eric-gunther (approved):

PR #546, opened 2023-03-02
Add more info patient search

luthfib:
Add more info the options in patient search. Expose prop needed to make patient search work.

image

Will be adding identifiers once more questions are answered.


eric-gunther (approved):
Clever to search by identifier when there's a number else search by name. I'm sure we'll get feedback on this search mechanism and how we present the results, but I think this is good enough to replace the default loading of our patient list for the time being.

PR #549, opened 2023-03-02
Ctw 803/generic resources components

WesleyKapow:
**TODO: Figure out how to persist filters when switching tabs in ZAP!**

This is a fairly large refactor of condition components to use new generics and new component breakdown and naming conventions.

Generic helpers:
* Add `ResourceTable` as a small wrapper around `Table`.
* Add `ResourceTableActions` to setup sort, filter, and additional actions above a resource table.

Condition refactors:
* Moved condition helpers into a directory, hopefully making it clear which condition components are meant to be more widely used.
* `PatientConditions` — sort/filter/actions/table for patient conditions on the patient/builder record.
* `PatientConditionsOutside` — same but for other provider records.
* `PatientConditionsBase` — helper component used by the above two.

Other note-able changes:
* Added table column headings to conditions.
* Simplified `Table` styling.. it now always uses none-rounded y borders and not x borders.
* Reduced length of some long sort names for conditions & medications
* Added ability to hide entire title row in ZAP.

Screenshot 2023-03-02 at 4 52 28 PM


src/components/content/conditions/patient-conditions-outside-badge.tsx
```diff
@@ -0,0 +1,20 @@
+import { Badge } from "@/components/core/badge";
+import { usePatientConditionsOutside } from "@/fhir/conditions";
+
+export const PatientConditionsOutsideBadge = () => {
```
WesleyKapow, line 4: Refactored this out of now removed `patient-conditions-tabs`. Didn't change anything outside of moving the code.

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -1,130 +1,75 @@
-import cx from "classnames";
-import { useRef, useState } from "react";
-import { useConditionHistory } from "../condition-history/conditions-history-drawer";
-import { conditionFilters } from "./filter-options";
-import { PatientConditionsActions } from "./patient-conditions-actions";
-import { patientConditionsColumns } from "./patient-conditions-columns";
 import {
```
WesleyKapow, line 1: Basically removed old file and created new one. This now holds the patient/builder records! Consumers of the old component will want to switch to using `PatientConditionsProfile`.

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -1,130 +1,75 @@
...
+      readOnly={readOnly}
+      rowActions={readOnly ? undefined : RowActions}
+    />
+  );
+};
 
-    // Get our conditions.
-    const patientConditions = patientConditionsQuery.data ?? [];
-    const otherConditions = otherConditionsQuery.data;
+const RowActions = ({ record }: RowActionsProps<ConditionModel>) => {
```
WesleyKapow, line 44: Moved row actions out of previous helper to be local to their single usage.

src/components/content/conditions/patient-conditions-outside.tsx
```diff
@@ -0,0 +1,66 @@
...
+      action={action}
+      className={className}
+      query={query}
+      readOnly={readOnly}
+      rowActions={readOnly ? undefined : RowActions}
+    />
+  );
+};
+
+const RowActions = ({ record }: RowActionsProps<ConditionModel>) => {
```
WesleyKapow, line 37: Moved row actions out of previous helper to be local to their single usage.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -86,16 +83,19 @@ const zusAggregatedProfile = ({
...
-        <h3 className="ctw-m-0 ctw-inline-block ctw-p-0 ctw-pb-3 ctw-text-lg ctw-font-medium">
-          {title}{" "}
-          {!removeBranding && (
-            <span className="ctw-text-sm ctw-font-light ctw-italic ctw-text-content-light">
-              Powered by <img src={ZusSVG} alt="Zus" className="-ctw-mb-1.5" />
-            </span>
-          )}
-        </h3>
-      </Title>
+      {!hideTitle && (
```
WesleyKapow, line 86: Hiding by `hideTitle` is only change here, rest is whitespace.

src/fhir/conditions.ts
```diff
@@ -356,3 +355,38 @@ export const deleteCondition = async (
...
   await queryClient.invalidateQueries([QUERY_KEY_PATIENT_CONDITIONS]);
 };
+
+// Filter out other conditions where:
+//  1. Condition is archived and includeArchived is false.
+//  2. CCS Category code starts with FAC or XXX.
+//  3. There is an existing patient condition with a matching known code.
+//     AND The other condition is older than the patient condition OR they
+//     have the same status.
+export const filterOtherConditions = (
```
WesleyKapow, line 365: Refactored from content/conditions/helper. Ideally our FHIR code should not depend on any component code, so for now I put this here.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -64,6 +60,7 @@ const zusAggregatedProfile = ({
   timelineProps,
   resources,
   title = "Outside Records",
+  hideTitle = false,
   removeBranding = false,
```
luthfib, line 64: I'm wondering if we should removeBranding by default right now.

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -1,130 +1,68 @@
...
-    const containerRef = useRef<HTMLDivElement>(null);
-    const breakpoints = useBreakpoints(containerRef);
-
-    // Drawer helpers.
-    const showConditionHistory = useConditionHistory();
-
-    // Data fetching.
-    const patientConditionsQuery = usePatientConditions();
-    const otherConditionsQuery = useOtherProviderConditionsDeduped();
+export const PatientConditions = ({ className }: PatientConditionsProps) => {
```
mrosata: can we put the telemetry back on this? Dealers choice, either
```tsx
const patientConditions = ({ className }: PatientConditionsProps) => {
  // code code code
 }
 
 export const PatientConditions = withErrorBoundary(patientConditions, "PatientConditions");
 ```
 
 or as it had been
 ```tsx
 export const PatientConditions = withErrorBoundary(({ className }: PatientConditionsProps) => {
   // code code code
 }, "PatientConditions");
 ```

> Personally I like the top version as all the arguments to `withErrorBoundary` fit on a single line that way.

File: src/components/content/conditions/patient-conditions.tsx [hunk 9]
```diff
@@ -1,130 +1,68 @@
...
-        ? PatientConditionHoverActions
-        : OtherProviderConditionHoverActions;
+  return (
+    <div className="ctw-flex ctw-space-x-2">
+      {!record.isDeleted && (
+        <button
+          type="button"
+          className="ctw-btn-default"
+          onClick={(event) => {
+            event.stopPropagation();
```
mrosata: It may make more sense to use this pattern
```tsx
<div
  onClick={(event) => event.stopPropagation()}
  onKeyDown={(event) => event.stopPropagation()}
>
  <button onClick={() => thingA(record)}>
    Action A
  </button>
  <button onClick={() => thingB(record)}>
    Action B
  </button>
</div>
```
instead of stopping propagation on the buttons. I do recall @luthfib and yourself doing some work to make sure clicks on action row buttons don't propagate up to the row itself, assuming this code is for that. Since the same applies for the wrapping element (clicking that div shouldn't propagate to the row), putting the `stopPropagation` just on the div solves the problem for the div and any children it has.

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -1,130 +1,68 @@
...
-        className={cx("ctw-patient-conditions ctw-bg-white", className, {
-          "ctw-patient-conditions-stacked": breakpoints.sm,
-        })}
+      <button
+        type="button"
+        className="ctw-btn-primary"
+        onClick={(event) => {
+          event.stopPropagation();
+          showEditConditionForm(record);
+        }}
```
mrosata: If my last comment about the action buttons event propagation is put into code then this also would become
```suggestion
        onClick={() => showEditConditionForm(record)}
```

src/components/content/conditions/patient-conditions.tsx
WesleyKapow: I think you're viewing an outdated diff. I put this on the `td` for the `RowActions` so that way consumers of `Table` don't have to worry about this.

WesleyKapow: https://github.com/zeus-health/ctw-component-library/pull/549/files#diff-f7717460aa725b63301805216365f488a26019582889ffe0ff22832fdcbcdcd2R85

src/components/content/patient-history/request-records-button.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+  return (
+    <button
+      type="button"
+      className="ctw-btn-clear ctw-link"
+      onClick={openHistoryRequestDrawer}
+      data-zus-telemetry-click="Request records"
+    >
+      Request Records
+    </button>
```
mrosata, line 14: **non-blocker** Canvas has noted to us that they don't use title case on buttons and they ask us to do the same for their components. So I've been utilizing `.ctw-capitalize` on all buttons and letting CSS capitalize the text for us and when we drop components in they can override button.ctw-capitalize to not have text-decoration.

Eventually as you pointed out, it would be nice to setup i18n, in the meantime letting CSS do the work seems fair.
```suggestion
    <button
      type="button"
      className="ctw-btn-clear ctw-link ctw-capitalize"
      onClick={openHistoryRequestDrawer}
      data-zus-telemetry-click="Request records"
    >
      Request records
    </button>
```

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
mrosata: Do you have an elevator pitch to why we should remove branding by default? I can imagine some EHRs asking us to remove branding but unlikely they would request that we turn on branding.

src/fhir/conditions.ts
mrosata: **non-blocker**
agreed. I actually stumbled on one of these earlier today myself https://github.com/zeus-health/ctw-component-library/blob/f543cd02402f1985221e776a45d94f136583c03d/src/fhir/document.ts#L3 -- not sure if you want to include refactoring that one as well, if you do, please leave a comment similar to this one
> Refactored from content/conditions/helper. Ideally our FHIR code should not depend on any component code, so for now I put this here.

for ease of review.

mrosata (approved):

src/fhir/conditions.ts
WesleyKapow: Yes, good spot! Leaving as is for now.. This PR is certainly big enough ;)

src/components/content/patient-history/request-records-button.tsx
WesleyKapow: I'm gonna punt on this for now. We really want to move toward a repeatable solution for this type of thing.

The problem with using classes like `ctw-capitalize` is that we may use it all over the place. In some cases consumers may want to overwrite it and in other places they won't. This creates a very unstable situation (land mine / foot gun / opposite of pit-of-success) where a developer comes along and adds `ctw-capitalize ` to capitalize some text and now the consumers that overwrote `ctw-capitalize` don't have the right formatting for that text.

src/components/content/conditions/patient-conditions.tsx
WesleyKapow: Done!

This was added to `PatientConditions` and `PatientConditionsOutside` but not `PatientConditionsProfile` as that is a dumb wrapper around `ZusAggregatedProfile`.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
luthfib: I dont have an elevator pitch, I just know that this is technically any that upgrades the version and is using this will see by default and will try to remove it. Not a big deal, but just have to remember to set removeBranding to true. I guess since we are not 1.0 not a big deal, but when adding features that expose potentially unwanted behavior for all I tend to default to false, but if we strongly about then sure we can leave it on by default.

WesleyKapow: I don't feel confident in weighing in here. Maybe let's punt on deciding this default for now and see how things shake out? Also note, components that use ZAP that also `hideTitle` will affectively hide the branding too.

mrosata (approved):

PR #553, opened 2023-03-03
lock headless ui version

luthfib:
Locks headless ui version.


pmadria3 (approved):

WesleyKapow:
@luthfib This sets up a trap for future us. What happens when future dev wants to upgrade headlessui version? How can we mitigate this risk?

PR #555, opened 2023-03-03
CTW-681/fine scoped metrics for dashboard

mrosata:
Add new method to `Telemetry.reportMetric` that will send metrics with tags to CTW /report/metric endpoint.
Add async wrapper `withTimerMetric` that will report the performance of an async function.
Storybook mocks are structured now as a nested object with arrays instead of just an array to support namespacing for mocks which allows for global mocking and better merging of msw mocks. 

---
If you want to know more about the change to `msw` mocks object, see this example, _(note the difference between this and `parameters.msw = [...handlers]`)_
```ts
// preview.jsx
export parameters = {
  msw: {
    handlers: {
      metrics: [
        rest.post("*/report/metric", (_, res, ctx) =>
          res(ctx.status(200), ctx.json({}))
        ),
      ],
    },
  },
};
```

```ts
// ../some-component/storybook-helpers/requests.ts
export parameters = {
  msw: {
    handlers: {
      mocks: [
        rest.post("*/api/peace/love", (_, res, ctx) =>
          res(ctx.status(200), ctx.json({}))
        ),
      ],
    },
  },
};
```
Storybook will properly merge the global `parameters.msw.handlers` with some-components `parameters.msw.handlers`. Because they have different keys (`metrics` and `mocks`), the merge doesn't clobber either of them and they will live and play in peace and harmony ☮️ 
This was required to have a global metrics mocked endpoint. Currently all story mocks live at `parameters.msw.handlers.mocks` but you might envision a world where each story uses its own key `parameters.msw.handlers.some-component`


src/fhir/action-helper.ts
```diff
@@ -18,28 +18,34 @@ export async function createOrEditFhirResource(
...
       }
-      return response;
-    }
-    const response = await fhirClient.create({
-      resourceType: resource.resourceType,
-      body: omitEmptyArrays(resource),
-    });
-    if (!isFhirError(response)) {
-      resourceModified.id = response.id;
+      Telemetry.reportActionFailure(`${resource.resourceType}.${action}`);
```
luthfib, line 35: Isn't this getting hit everytime?

src/fhir/practitioner.ts
```diff
@@ -53,3 +54,8 @@ export async function getUsersPractitionerReference(
     display: claimsAuthEmail(requestContext.authToken),
   };
 }
+
+export const getUsersPractitionerReference = withTimerMetric(
+  getUsersPractitionerReferenceReq,
+  "req.get_users_practitioner_reference"
```
luthfib, line 60: If we aren't going to export it why not just wrap the initial function with `withTimerMetric`?

luthfib: sorry it didn't highlight everything I wanted highlighted.. reference to `getUsersPractitionerReferenceReq`

src/fhir/action-helper.ts
mrosata: yeah that's not even supposed to be there, already in the catch block. Good eye

src/fhir/practitioner.ts
mrosata: Two fold, firstly to make the PR smaller b/c wrapping would indent outward the whole body of the function. Second reason was for readability so that a human could see all the arguments to `withTimerMetric`. I realized this after wrapping so many components with `withErrorBoundary` that it looks better in this style... with exceptions to every rule of course.

luthfib (approved):

src/fhir/practitioner.ts
WesleyKapow: NOTE: Don't worry too much about indent = large diff. You can always note that in the PR description and reviewers can use github's "ignore whitespace" feature.

PR #558, opened 2023-03-06
Use react-query for authTokenURL requests

WesleyKapow:
This fixes a race condition where if multiple components need data, and thus need an auth token, we could end up sending multiple simultaneous requests to the atuhTokenURL. We now wrap our requests  react-query's `fetchQuery` which will yield the cached promise for `staleTime` (currently 30s). This way multiple requests only trigger a single actual fetch.

I verified this by running CTW standalone against my local component library. Loading the conditions page previously yielded two auth token requests where now it only yields one. I also added some console logs to see the action (Note: the second "call fetchQuery" is not followed by a "fetch the token" since `fetchQuery` returns a cached promise).
```
call fetchQuery
fetch the token
call fetchQuery
```


luthfib (approved):
LGTM. You mentioned that you added console.logs but I dont see any added, just wanted to call that out

WesleyKapow:
@luthfib added console logs for testing purposes and then removed.

PR #559, opened 2023-03-07
CTW-782 Adjust columns for conditions component

eric-gunther:
- Making display status not bold (so only the diagnosis string is bold)
- Added "Last Updated:" before the recorded date (which I verified is the latest recorded date for other provider records, so I think "Last Updated" is reasonable here).
- Added recorder name in parentheses, if available
- Removed Managing Org display from Other Provider Records tab. I don't think this makes sense to display at this level because we won't have a managing org on a summary resource like this, since the diagnosis could've been summarized from any number of outside records from different managing orgs that information makes more sense in the history (where it's already displayed).

[Verifying with UX in this Slack thread](https://zushealth.slack.com/archives/C03M2KKTXPZ/p1678200670412839?thread_ts=1678125980.304119&cid=C03M2KKTXPZ).

*Before*:
Pasted Graphic 3
Pasted Graphic 4

*After*:
Pasted Graphic 5
>

### DIFF
--- a/.changeset/sixty-cherries-rest.md
+++ b/.changeset/sixty-cherries-rest.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Adjust display values for columns in conditions component.
--- a/src/components/content/conditions/patient-conditions-columns.tsx
+++ b/src/components/content/conditions/patient-conditions-columns.tsx
@@ -40,19 +40,13 @@ export const patientConditionsColumns: TableColumn<ConditionModel>[] = [
         </div>
         <div className="ctw-pc-status-and-extra">
           <div className="ctw-pc-status">{condition.displayStatus}</div>
-
-          {condition.isSummaryResource ? (
-            <div>
-              {compact([
-                condition.patient?.organization?.name,
-                condition.recordedDate,
-              ]).join(" ")}
-            </div>
-          ) : (
-            <div>
-              {compact([condition.recorder, condition.recordedDate]).join(" ")}
-            </div>
-          )}
+          <div>
+            Last Updated:{" "}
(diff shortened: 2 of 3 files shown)


src/components/content/conditions/helpers/patient-conditions.scss
```diff
@@ -31,10 +31,6 @@
     @apply ctw-text-base;
   }
 
-  .ctw-pc-status {
-    @apply ctw-text-sm ctw-font-medium ctw-capitalize ctw-text-content-black;
-  }
-
```
luthfib, line 37: I see you removed this class but I see it used in `[src/components/content/conditions/helpers/columns.tsx](https://github.com/zeus-health/ctw-component-library/pull/559/files#diff-7abc8f27e8ba602ddbab8498c89295bd1a38b9d9f85821dfae1d093f901b2ffd)`

WesleyKapow (commented):
LGTM outside of:
1. Luthfi's comment -- you can. remove `ctw-pc-status` class from the div.
2. Where's "Removed Managing Org display from Other Provider Records tab"? Oh looks like you removed it from both tabs, not just other providers. Is that right?

eric-gunther:
> LGTM outside of:
> 
> 1. Luthfi's comment -- you can. remove `ctw-pc-status` class from the div.
> 2. Where's "Removed Managing Org display from Other Provider Records tab"? Oh looks like you removed it from both tabs, not just other providers. Is that right?

There was previously a ternary so that managing org was only displayed for other provider records, but it doesn't make sense to ever show it here and the logic changes have made it so conditions list and other provider records don't need different behavior anymore.

WesleyKapow (approved):

luthfib (approved):

PR #566, opened 2023-03-08
Support different word for "condition" (i18next)

WesleyKapow:
Add i18next library and setup pattern for internationalization.

Add glossary for condition and setup all text containing "condition" to use i18next and the glossary.

Add `locals` prop to `CTWProvider`, allowing consumers to specify their own translations. Currently this allows consumers to specify their own word for "condition", e.g. "problem" as shown below:

```jsx
<CTWProvider
  locals={{
    en: {
      glossary: {
        condition_one: "problem",
        condition_other: "problems",
      },
    },
  }}
/>
  //...
</CTWProvider>
```

Screenshot 2023-03-07 at 5 40 47 PM
Screenshot 2023-03-08 at 12 04 04 PM
Screenshot 2023-03-08 at 12 02 27 PM
Screenshot 2023-03-08 at 12 26 24 PM
Screenshot 2023-03-08 at 12 29 12 PM
Screenshot 2023-03-08 at 1 08 09 PM
Screenshot 2023-03-08 at 1 12 16 PM
Screenshot 2023-03-08 at 1 08 17 PM


src/components/content/condition-history/conditions-history-drawer.tsx
```diff
@@ -42,10 +43,14 @@ export function ConditionHistoryDrawer({
...
   onEdit,
 }: ConditionHistoryDrawerProps) {
+  const { t } = useTranslation();
+
   return (
     <Drawer
       className={className}
-      title="Condition History"
+      title={t("resource.history.heading", {
+        resource: t("glossary:condition_one"),
```
luthfib, line 52: Why is it called `condition_one`?

WesleyKapow: Standard way to handle pluralization. This allows for things like `You have {{count}} $t(glossary:condition, {count})`. The interesting part is that some languages have more than two forms for plurals, so you have keys like `key_two`, `key_few` and `key_many`.

See https://www.i18next.com/translation-function/plurals#singular-plural for details.

pmadria3 (approved):
Nice work! LGTM

PR #569, opened 2023-03-08
disable metrics from localhost 3000

mrosata:
Disable sending metrics to CTW from localhost:3000

### DIFF
--- a/.changeset/tough-countries-melt.md
+++ b/.changeset/tough-countries-melt.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Disable metrics to ctw from localhost:3000.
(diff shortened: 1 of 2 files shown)


luthfib (approved):

luthfib:
It seems you have a telemetry test you need to disable

luthfib (changes requested):
Seems you have a failing test around telemetry

luthfib (approved):

WesleyKapow:
Why not prevent for all localhost? why specifically that port? Does storybook cause reporting? What if we change the port?

mrosata:
> Why not prevent for all localhost? why specifically that port? Does storybook cause reporting? What if we change the port?

Storybook has this request globally mocked out. If we change the port test will fail, also auth won't work anymore b/c currently 3000 is the only allowed localhost port

PR #570, opened 2023-03-09
Dedupe logic condition history

luthfib:
Couple of decisions that had to be made:

- We sort by hasEnrichment as well because we want enriched records to be preferred in uniqWith function.
- Sort by recordedDate and then version id because recorded date is can be a flat date and it was not sorting by the most recent record if it was on the same date. Not sure maybe instead of versionId maybe lastUpdated time from the meta tag?

Screenshot 2023-03-08 at 11 16 09 PM

### DIFF
--- a/src/components/content/condition-history/condition-history-filters.tsx
+++ b/src/components/content/condition-history/condition-history-filters.tsx
@@ -35,9 +35,5 @@ const valuesToDedupeOn = (condition: ConditionModel) => [
   condition.onset,
   condition.abatement,
   condition.encounter,
-  condition.knownCodings.map((coding) => [
-    coding.display,
-    coding.system,
-    coding.code,
-  ]),
+  condition.knownCodings.map((coding) => [coding.system, coding.code]),
 ];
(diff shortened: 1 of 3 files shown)


src/components/content/condition-history/condition-history-filters.tsx
```diff
@@ -12,17 +12,23 @@ export const applyConditionHistoryFilters = (
     (c) => new ConditionModel(c, includedResources)
   );
 
+  // We sort by hasEnrichment as well because we want enriched records to be preferred in uniqWith function.
   const sortedConditions = orderBy(
     conditionModels,
-    (c) => c.resource.recordedDate ?? "",
+    [(c) => c.resource.recordedDate ?? "", (c) => c.hasEnrichment],
```
mrosata, line 18: not a blocker
```suggestion
    ["resource.recordedDate", "hasEnrichment"],
```

mrosata (approved):
lgtm, left 1 suggestion non-blocker

mrosata (approved):

PR #577, opened 2023-03-09
Ctw-809/care-team-updates

pmadria3:
CTW-809

This PR updates care team with each row being an individual practitioner and then the drawer having details about the practitioner. 

https://user-images.githubusercontent.com/98342697/225120459-e7366fba-0aa5-464c-8dc5-e59f9657f727.mov


src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,29 +1,52 @@
...
+        participant.member?.reference as string
+      );
+      if (practitioner?.id) {
+        if (participant.member?.reference) {
+          const careTeamPractitioner = new CareTeamPractitionerModel(
+            careTeamObject,
+            practitioner
+          );
+          if (careTeamMemberMap.has(practitioner.id)) {
+            const latestPractitioner = careTeamMemberMap.get(practitioner.id);
```
luthfib, line 31: Why is this latestPractitioner? Just seems like you are getting any practitioner from the map?

src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,29 +1,52 @@
...
-  const careTeamData = uniqWith(currentCareTeam, (a, b) =>
-    isEqual(valuesToDedupeOn(a), valuesToDedupeOn(b))
-  );
 
-  return careTeamData;
-};
+  const careTeamMemberMap = new Map<
+    string,
+    CareTeamPractitionerModel | undefined
+  >();
```
mrosata, line 17: Highlighting these 4 lines, but the comment applies to the whole function. This function is named `applyCareTeamFilters` but it's doing a lot more than just applying filters. The main argument type and return type of a filter function should be the same. IE: `const filterCats: (cats: Cat[]) => Cat[]`.

I would suggest either pulling your logic out in a new function (likebuildPractitionersFromCareTeams) and doing something more like
```ts
const careTeamFiltered = applyCareTeamFilters(careTeams);
const careTeamPractitioners = buildPractitionersFromCareTeams(careTeamFiltered);
```

or you could just rename applyCareTeamFilters. Choice is yours

src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,29 +1,52 @@
...
-  careteam.practitionerQualification,
-  careteam.includedPerformer,
-];
+  careTeamModels.forEach((careTeamObject) => {
+    careTeamObject.resource.participant?.forEach((participant) => {
+      const practitioner = careTeamObject.getPractitionerByID(
+        participant.member?.reference as string
+      );
+      if (practitioner?.id) {
+        if (participant.member?.reference) {
```
luthfib, line 25: nested if's seem unnecessary, should just be able to do  `practitioner?.id && participant.member?.reference`

src/fhir/models/careteam-practitioner.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class CareTeamPractitionerModel extends FHIRModel<fhir4.Practitioner> {
+  constructor(careTeam: CareTeamModel, practitioner: fhir4.Practitioner) {
+    super(practitioner);
+    this.careTeam = careTeam;
+  }
+
+  careTeam: CareTeamModel;
+
+  get PractitionerName() {
```
luthfib, line 13: should be `practitionerName`

src/components/content/care-team/patient-careteam-filters.tsx
luthfib: Also this function is complex enough to where some comments describing why you are doing what you are doing would be helpful

luthfib:
Once you address Mikes comment about splitting up, then LGTM

src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,52 +1,41 @@
...
-  });
+    })
+  );
 
-  return compact([...careTeamMemberMap.values()]);
+  const sortedCareTeamPractitionerModels = orderBy(
+    careTeamPractitionerModels,
+    [(c) => c.effectiveStartDate, (c) => c.id],
+    ["desc", "desc"]
+  );
```
mrosata: These orderBy's feel easier to quickly read over when the property getters are just key names instead of functions
```suggestion
  const sortedCareTeamPractitionerModels = orderBy(
    careTeamPractitionerModels,
    ["effectiveStartDate", "id"],
    ["desc", "desc"]
  );
```

mrosata: are we going to get weird sorting if `effectiveStartDate`s have varied formats (ie like the date sort issues in the table). If so might need to instead make the array of property names/getter functions to normalize the date
```
const sortedCareTeamPractitionerModels = orderBy(
    careTeamPractitionerModels,
    [(c) => normalizeDateSomehow(c.effectiveStartDate), "id"],
    ["desc", "desc"]
  );
```

src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,52 +1,41 @@
...
+
+  return uniqWith(sortedCareTeamPractitionerModels, (a, b) =>
+    isEqual(a.id, b.id)
+  );
```
mrosata, line 40: Should be able to use uniqBy here instead 
```ts
return uniqBy(sortedCareTeamPractitionerModels, "id")
```

src/utils/nodash/index.ts
```diff
@@ -36,3 +36,4 @@ export { default as startCase } from "lodash/startCase";
 export { default as trim } from "lodash/trim";
 export { default as uniq } from "lodash/uniq";
 export { default as uniqWith } from "lodash/uniqWith";
+export { default as uniqBy } from "lodash/uniqBy";
```
mrosata, line 39: non-blocker nit, all the exports in this file are alphabetical order besides this added one.. it's not enforced by linting, no big deal, just something to keep in mind in future

src/components/content/care-team/patient-careteam-filters.tsx
```diff
@@ -1,29 +1,39 @@
...
+import { orderBy, uniqBy } from "@/utils/nodash";
 
 export const applyCareTeamFilters = (
   data: fhir4.CareTeam[],
   includedResources: ResourceMap
 ) => {
-  const careTeamModel = data.map(
+  // 1. Make a list of CareTeamPractitionerModel and filter out any practioners that dont have an id.
+  // 2. Sort effectiveStartDate and practitionerId to ensure we remove older records.
+  // 3. UniqWith of the practioners to remove dupe records.
```
mrosata, line 12: Non-blocker This style of commenting is notorious for growing out of sync with code. It would be better if the code didn't need comments, but if does need comments
```
// Make a list of CareTeamPractitionerModel and filter out any practitioners that dont have an id.
.. write code that gets list

// Sort effectiveStartDate and practitionerId to ensure we remove older records.
.. write code that sorts

UniqWith of the practitioners to remove dupe records.
... write code that does this
```
We can already see just in this PR process it's started to drift as `uniqWith` changed to `uniqBy` but where the changed code is so far away from the comments they didn't update in sync.

mrosata: The only part of this code that might need commenting is actually the sort and uniqBy at the end of function

src/components/content/care-team/patient-careteam.tsx
```diff
@@ -56,10 +53,21 @@ export function PatientCareTeam({
...
+export const careTeamData = (
+  careTeamPractitioner: CareTeamPractitionerModel
+) => [
+  { label: "Organization", value: careTeamPractitioner.managingOrganization },
+  {
+    label: "CareTeam Telecom",
+    value: careTeamPractitioner.telecom && (
+      <div>
+        {careTeamPractitioner.telecom.map((item, index) => (
+          // eslint-disable-next-line react/no-array-index-key
```
mrosata, line 65: non-blocker.. we need to add this rule exception to eslint b/c it's not a rule we follow so no point in us getting yelled at by the linter all the time and having to add these ignores

mrosata (approved):

mrosata (approved):

src/components/content/care-team/patient-careteam-filters.tsx
pmadria3: Removed the comment - unnecessary and the code is pretty self-explanatory

PR #578, opened 2023-03-13
Clean up sort & filter UI

WesleyKapow:
See CTW-798 for list of changes.

Much of the work was tweaking styles to get things consistent across select & checkbox dropdowns. Font-awesome react library was added to make using FA icons easier.

# Screenshots

### Before
Screenshot 2023-03-13 at 11 38 49 AM
Screenshot 2023-03-13 at 11 38 11 AM

### After
Screenshot 2023-03-13 at 11 39 52 AM
Screenshot 2023-03-13 at 11 32 19 AM

### Before
Screenshot 2023-03-13 at 11 38 24 AM

### After
Screenshot 2023-03-13 at 11 32 31 AM

### Before
Screenshot 2023-03-13 at 11 37 50 AM
Screenshot 2023-03-13 at 11 37 33 AM

### After
Screenshot 2023-03-13 at 11 31 24 AM
Screenshot 2023-03-13 at 11 49 35 AM

## Select Filter
NOTE: I believe this is not currently used. Here, I modified status filter to use it for purposes of testing.
### Before
Screenshot 2023-03-13 at 11 48 45 AM

### After
Screenshot 2023-03-13 at 11 33 17 AM

## Tag Filter
Not much changed here, mostly just the icon.
### Before
Screenshot 2023-03-13 at 11 39 04 AM

### After
Screenshot 2023-03-13 at 11 40 13 AM


src/components/core/filter-bar/filter-bar-utils.tsx
```diff
@@ -1,66 +1,13 @@
...
-import cx from "classnames";
-import { RotateIcon } from "../rotate-icon";
+import type { FilterItem } from "./filter-bar-types";
+import { ListBoxOptionStatus } from "../list-box/list-box";
 import { FilterChangeEvent, FilterValuesRecord } from "./filter-bar-types";
 import { isFunction, set } from "@/utils/nodash/fp";
 
-const iconClassNames = "ctw-text-content-light ctw-h-3.5 ctw-mr-1 ctw-my-auto";
-
-export function getIcon(icon: string) {
```
WesleyKapow, line 22: NOTE: Far better to allow consumers to pass in whatever icon they want instead of a string that the component needs to know how to turn into an icon. This reduces coupling! As before, the component was coupled to every consumer.

luthfib (dismissed):

src/components/core/list-box/list-box.tsx
```diff
@@ -86,17 +85,24 @@ export function ListBox<T extends MinListBoxItem>({
...
-                  <span className="ctw-inline-flex ctw-align-middle">
+                  <span
+                    className={cx("ctw-inline-flex ctw-align-middle", {
+                      "ctw-font-semibold": selected && !useBasicStyles,
+                    })}
+                  >
                     {renderDisplay(item, { active, selected, listView: true })}
                   </span>
-                  {!useBasicStyles && selected && (
+                  {!useBasicStyles && selected && !item.key.startsWith("_") && (
```
mrosata, line 100: non-blocker but I'm torn on this a bit, because we're saying that any consumers of `<ListBox />` won't be able to use item keys that begin with `"_"` in the same way as keys that don't begin with `"_"` despite that more or less being a constraint of how the filter bar wants to use list-box.

mrosata (approved):

PR #580, opened 2023-03-13
Ctw 836/filter builder id locally

luthfib:
Filter builder tagged records client side in bid to improve performance. NOTE review pr carefully as this will change most requests in ctw component library.


src/fhir/search-helpers.ts
```diff
@@ -89,15 +89,12 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
...
-  const builderTag = `${SYSTEM_ZUS_OWNER}|builder/${requestContext.builderId}`;
   type FirstPartyParams = {
     _tag?: string[];
     firstparty?: boolean;
     "_tag:not"?: string[];
   };
-  const firstPartyParams: FirstPartyParams = {
-    _tag: [builderTag],
-  };
+  const firstPartyParams: FirstPartyParams = {};
```
luthfib, line 97: I believe not adding builder tag here should be fine since we are doing firstParty filter, but correct me if I am wrong

pmadria3 (approved):
LGTM

src/fhir/search-helpers.ts
mrosata: Sounds right. @pmadria3 is also doing some PR work around these tags. Might be good to check in with him to see if there's any overlap

mrosata (approved):

src/utils/nodash/index.ts
```diff
@@ -37,3 +37,4 @@ export { default as trim } from "lodash/trim";
 export { default as uniq } from "lodash/uniq";
 export { default as uniqBy } from "lodash/uniqBy";
 export { default as uniqWith } from "lodash/uniqWith";
+export { default as filter } from "lodash/filter";
```
mrosata, line 40: nit, could you try to alphabetize new entries in future

PR #582, opened 2023-03-14
CTW-667/condition unknown status

WesleyKapow:
Fixed up status filtering for conditions.

1. Show all status options independent of data.
2. Include "Unknown" in list of status options, with ability to filter unknown statuses.
3. Remove "Reset Filter" button from individual filters.
4. Apply filters that have 0 things checked. Luckily when you first add a filter, we don't applyFilters yet, giving users a chance to make some selections. This will also get reworked in the near future to default new checkbox filters to have an "All" selected.

Not implemented, just showing future "All" functionality.
Screenshot 2023-03-14 at 12 28 04 PM

New status filters (implemented here!)
Screenshot 2023-03-14 at 12 11 29 PM

### DIFF
--- a/src/components/content/conditions/helpers/filters.ts
+++ b/src/components/content/conditions/helpers/filters.ts
@@ -7,6 +7,10 @@ import {
   FilterItem,
 } from "@/components/core/filter-bar/filter-bar-types";
 import { ConditionModel } from "@/fhir/models";
+import {
+  conditionStatuses,
+  outsideConditionStatuses,
+} from "@/fhir/models/condition";
 import { uniqueValues } from "@/utils/filters";
 
 export function conditionFilters(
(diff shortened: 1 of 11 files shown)


src/components/content/conditions/helpers/filters.ts
```diff
@@ -9,14 +9,26 @@ import {
...
-      values: uniqueValues(conditions, "displayStatus"),
+      values: outside
+        ? ["Active", "Inactive", "Unknown"]
+        : [
+            "Active",
+            "Inactive",
+            "Entered in Error",
+            "Pending",
+            "Refuted",
+            "Unknown",
```
luthfib, line 30: Hmmm, not a fan of hardcoding values here, I think at the very least these should be typed, and a constant that we can reuse?

luthfib: Also are we aware that these means we will now have options that will lead to empty results? I would verify with UX this is desired behavior

WesleyKapow: Yes that is the intention for status as it has been confusing why other statuses aren't available.

luthfib: You are right. I thought we had a type had a type for `displayStatus`. In reality `displayStatus` shouldn't return a string but a type, which could be `zusStatus` or something like that, that way we could make sure they are both of that type.

luthfib (approved):

PR #583, opened 2023-03-14
Ctw 840/condition 2.0 patient history fixes

luthfib:
Request records now appears in empty table state when patient has never made request to patient history api. Adds ability to pass loading state to PatientConditionsBase. Export last retrieved date so we can use it outside of conditions 2.0. 

Screenshot 2023-03-14 at 2 08 42 PM


src/components/content/patient-history/patient-history-action.tsx
```diff
@@ -9,18 +9,15 @@ export type PatientHistoryActionProps = {
...
   return (
-    <div>
-      {patientHistory.lastRetrievedAt && (
-        <PatientHistoryTableHeaderMessage
-          patientHistory={patientHistory}
-          message="Last Retrieved"
-        />
+    <div className="ctw-flex ctw-space-x-1">
+      <PatientHistoryLastRetrieved />
+      {!hideRequestRecords && !showRequestRecordsRequestButtonQuery.data && (
```
WesleyKapow, line 18: I think we always want to show the request records button unless `hideRequestRecords`. As in, we shouldn't hide it when there's data. We want to provide a way for them to request it again.

src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -143,3 +144,16 @@ function usePatientHistoryDetails() {
...
   );
 }
+
+export function useShowPatientHistoryRequestButton() {
+  const query = usePatientConditionsOutside();
+  const patientHistoryQuery = usePatientHistoryDetails();
+  const otherRecordsExist = query.data.length === 0;
+
+  return {
+    data:
```
WesleyKapow, line 154: This looks more like a bool than a set of data. What should this be named?

src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -143,3 +144,16 @@ function usePatientHistoryDetails() {
     }
   );
 }
+
+export function useShowPatientHistoryRequestButton() {
```
WesleyKapow, line 148: The name is tied to its use but what is this really? What information is this providing / what do we want it to provide?

WesleyKapow (changes requested):
Looks like `usePatientHistory` should already provide the `lastRetrievedAt`. Could you simplify this by just using that existing hook and maybe add an `isLoading` to it?

I don't think we need `PatientHistoryLastRetrieved` as that just wraps the header-message with a check.. we can do that ourselves.

`useShowPatientHistoryRequestButton` is probably not needed either as we want to construct our empty message with the request records button IF there is no `lastRetrievedAt` which we can get from `usePatientHistory`.

src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -143,3 +145,15 @@ function usePatientHistoryDetails() {
     }
   );
 }
+
+export function useHasOtherRecordData() {
+  const query = usePatientConditionsOutside();
```
WesleyKapow, line 150: This makes it conditions specific and thus should not be in the general `patient-history` area.

src/components/content/patient-history/patient-history-action.tsx
```diff
@@ -9,18 +9,15 @@ export type PatientHistoryActionProps = {
...
-          patientHistory={patientHistory}
-          message="Last Retrieved"
-        />
-      )}
-
-      {!hideRequestRecords && <RequestRecordsButton />}
+    <div className="ctw-flex ctw-space-x-1">
+      <PatientHistoryLastRetrieved />
+      {!hideRequestRecords &&
+        !showRequestRecordsRequestButtonQuery.hasNoOutsideData && (
```
WesleyKapow, line 18: What if instead, you rely on consumers to specify `hideRequestRecords` depending if there's data or not?

src/components/content/conditions/patient-conditions-outside.tsx
```diff
@@ -19,6 +21,14 @@ const PatientConditionsOutsideComponent = ({
   readOnly = false,
 }: PatientConditionsOutsideProps) => {
   const query = usePatientConditionsOutside();
+  const showRequestRecordsRequestButtonQuery = useHasOtherRecordData();
+
+  const emptyMessage = showRequestRecordsRequestButtonQuery.hasNoOutsideData ? (
```
WesleyKapow, line 26: Can this instead be based on `usePatientHistory` somehow? We know we won't have any outside records because this is for `emptyMessage`, so we only want to customize the empty message if we have never requested patient history before (do I have that right?).

src/components/content/patient-history/patient-history-last-retrieved.tsx
```diff
@@ -0,0 +1,17 @@
+import { PatientHistoryTableHeaderMessage } from "./patient-history-table-header";
+import { usePatientHistory } from "./use-patient-history";
+
+export const PatientHistoryLastRetrieved = () => {
```
WesleyKapow, line 4: This is only used in one place, is it really needed?

WesleyKapow (changes requested):

src/components/content/patient-history/patient-history-last-retrieved.tsx
luthfib: We need to export this because this will be used by consumers?

src/components/content/patient-history/patient-history-action.tsx
luthfib: Not sure I follow here... There are two scenarios. One if consumers just want to flat out hide request records and the other is if we want to hide it conditionally based on the logic above. Not sure how that fits both use cases, but maybe I am missing something.

src/components/content/patient-history/patient-history-last-retrieved.tsx
WesleyKapow: Why not have consumer use `usePatientHistory` and then `{patientHistory.lastRetrievedAt && <div>Last Retrieved {patientHistory.lastRetrievedAt}</div>`.. guess there's `formatISODateStringToDate` being used by `PatientHistoryTableHeaderMessage`, but that's simple enough too. We probably don't need this component nor `PatientHistoryTableHeaderMessage` (which looks outdated.. it has a switch on status and renders the same thing regardless of the status).

WesleyKapow: (I'm trying to keep things simple.. `usePatientHistory` is a hook that already provides patient history details that consumers should need. We don't then need simple components that show parts of these details.)

luthfib: Sounds more complicated to me, and more code for the consumer. But if we want to avoid exporting something else I can go that route.

WesleyKapow: I think you can roll `formatISODateStringToDate` into `usePatientHistory` so that `lastRetrievedAt` uses that formatted date. I don't think we use the date as a date elsewhere.

I think it keeps things simpler.. sure it's a little more code for the consumer.. but it's one conditional.

That said, we do have the `RequestRecordsButton`.. it's doing a little bit more but also fairly simple. I'm ok with keeping this lastRetrieved component but maybe kill off the `PatientHistoryTableHeaderMessage` component?

luthfib: Sounds good.

src/components/content/conditions/patient-conditions-outside.tsx
```diff
@@ -31,7 +34,13 @@ const PatientConditionsOutsideComponent = ({
   ) : undefined;
 
   const action = (
-    <PatientHistoryAction hideRequestRecords={hideRequestRecords || readOnly} />
+    <PatientHistoryAction
+      hideRequestRecords={
+        (hideRequestRecords &&
```
WesleyKapow, line 39: Should probably be an `||`. Hide it under any of these circumstances.

WesleyKapow (approved):

PR #592, opened 2023-03-17
Ctw 848/zap scrolling on overflow

mrosata:
#### Scrollbars

Added classes `.ctw-scrollable-pass-through-height` and `.ctw-scrollable-content`

https://www.loom.com/share/8187df91f0b34ddb967849fe538cf32c

Problem: While adding scrollbars to an element is fairly simple, adding scrollbars to a nest container can be very nuanced. This is because scrollbars are only added when an element grows past a dimension of its parent. If we have nested elements `A -> B -> C` and want `C` to scroll when a dimension of `B + C > A`, then we need 
- To have a fixed height of `A`
- To have `B + C` able to grow to fill the space of `A`
- To add `overflow: scrollbar` to `C`

Now with more nesting `A -> B1 -> B2 -> BN -> C`, we need to similarly for CSS to be able to know when the nested content has outgrown `A` so it can put scrollbars on `C`. Here's an image with all the nested containers from `A -> ... -> C` in the zus-aggregated-profile. Notice, the outlines are only on elements that both descendants of `A` and ancestors of `C`, we don't care about children of `B` that aren't also descendants of `C`.

Screen Shot 2023-03-17 at 10 41 24 AM

Now in that image above we are saying to ZAP, that collectively all its content shouldn't outgrow 500px, and the inner table grows to the max size it can within that bound before growing scrollbars. All we need to do is set height on the `.ctw-zus-aggregated-profile`.
```tsx
  <style>
    .ctw-zus-aggregated-profile {
      height: 500px;
    }
  </style>
  <ZusAggregatedProfile
    resources={["observations-outside", "observations"]}
    title="Observations"
  />
```
## New implementation
For adding scrolling to an element, you will need to add the `.ctw-scrollable-content` class to the element you'd like to scroll and make sure to place the class `.ctw-scrollable-pass-through-height` on every ancestor of that scrollable element.

```tsx
<div className="ctw-scrollable-pass-through-height">
  <div>
    <h1>I am a title</h2>
  </div>
  <div className="ctw-scrollable-pass-through-height">
    <div>Random content</div>
    
    <div className="ctw-scrollable-content">
       <table>
         {/* A large table */}
       </table>
    </div>
    
    <div>Random content</div>
  </div>
</div> 
```


luthfib (approved):

WesleyKapow:
Please hold off on merging until we can discuss options on Monday!

src/components/content/zus-aggregated-profile/zus-aggregated-profile.stories.tsx
WesleyKapow, line 1: This scrolling functionality is very fragile as it depends on all divs having the right class. Therefore, we really need a test to ensure we don't break it!

src/components/content/observations/patient-observations-outside.tsx
```diff
@@ -11,8 +11,16 @@ const Component = ({ className }: PatientObservationsOutsideProps) => {
   const diagnosticReports = usePatientDiagnosticReportsOutside();
 
   return (
-    <div className={cx(className, "ctw-patient-observations")}>
-      <PatientObservationsBase query={diagnosticReports} />
+    <div
```
WesleyKapow, line 14: I don't think this wrapper is needed and it deviates from how conditions is setup.

src/components/core/main.scss
```diff
@@ -96,6 +96,25 @@
...
+    }
+    /* all ascendants of "ctw-scrollable-content" (without explicit height) need .ctw-scrollable-pass-through-height */
+    &-pass-through-height {
+      display: flex;
+      flex-direction: column;
+      flex-grow: 1;
+      /* Need to keep min-height for Firefox */
+      min-height: 0;
+    }
+  }
```
WesleyKapow, line 116: nit: Make these not nested. Typically nesting is used for nested selectors and not as a way to DRY up class names. The main issue is it makes these class names harder to grep/find.

WesleyKapow (changes requested):

src/components/content/zus-aggregated-profile/zus-aggregated-profile.stories.tsx
```diff
@@ -106,3 +106,38 @@ export const Everything: StoryObj<Props> = {
...
+      "care-team",
+      "timelines",
+    ],
+  },
+  decorators: [
+    (Story, { args }) => (
+      <div className="ctw-border-solid ctw-border-divider-light ctw-p-2">
+        <h3>Fixed height container</h3>
+        <pre>{"// CSS\n.ctw-zus-aggregated-profile { height: 450px }"}</pre>
+        <style>{".ctw-zus-aggregated-profile { height: 450px }"}</style>
```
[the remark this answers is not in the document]
mrosata: This is for customers/consumers using our Storybook. To illustrate they need to set a height on that component I wanted to use a language everyone understands (CSS) rather than tailwind.

PR #595, opened 2023-03-17
CTW-854 Changing allergy tab display and adding option to override outside medications tab name

eric-gunther:


src/App.tsx
```diff
@@ -38,6 +38,20 @@ const theme = {
   },
 };
 
+// Glossary for translation
```
eric-gunther, line 41: Updating App.tsx to better match what's going in to Elation, if only so our local dev/troubleshooting looks closer to what they're using.

pmadria3 (approved):

PR #601, opened 2023-03-21
Fix story decorators for scroll story

WesleyKapow:
The issue was setupZusAggregatedProfileMocks is returning decorators but then we were overwriting it with our custom decorator. This caused a bug where on a fresh load certain resources would fail to fetch any data as the resource cache wasn't being initialized in the decorators as it should be.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@zus-health/ctw-component-library",
-  "version": "0.50.1",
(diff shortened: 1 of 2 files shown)


mrosata (approved):

mrosata:
There are probably a bunch of other stories doing similar (spreading instead of reusing the values). Probably be good to refactor them all

PR #603, opened 2023-03-22
Ctw 826/patient history index

luthfib:
Create POC table of patients who have made patient history requests. 

image


src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -110,6 +115,70 @@ async function getPatientRefreshHistoryMessages(
...
+    return await response.json();
+  } catch (err) {
+    throw errorResponse(
+      "Failed fetching patient refresh history messages",
+      err
+    );
+  }
+}
+
+export function useBuilderPatientHistoryList(
```
WesleyKapow, line 147: Maybe pull this and helper function out into new file `use-builder-patient-history-list.ts`.

Also in general I try to keep helper functions below the exports. That way it's easier to jump in to a file and see what the main responsibilities are.

src/components/content/patient-history/patient-history-table.tsx
```diff
@@ -0,0 +1,120 @@
...
+              currentPage={currentPage}
+              pageSize={pageSize}
+              isLoading={isFetching}
+            />
+          </Table>
+        </div>
+      </CTWBox.StackedWrapper>
+    );
+  },
+  "PatientsTable"
```
mrosata, line 92:
```suggestion
  "PatientsHistoryTable"
```

src/fhir/models/patient-history.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  get lastRetrievedAt() {
+    // eslint-disable-next-line no-underscore-dangle
+    return formatISODateStringToDate(this.historyInfo?._lastUpdated);
+  }
+
+  get retrievedStatus() {
+    return this.historyInfo?.status;
+  }
+}
```
eric-gunther, line 30: It feels strange to me that this model would extend the FHIR model. Patient history requests are a Zus specific thing that really have nothing to do with FHIR. I get that you're extending `FHIRModel<PatientModel>` in order to acquire the name getters, but I wonder if it makes more sense to have a simpler model here since you're constructing these models in `use-builder-patient-history-list.ts` anyway. I don't feel strongly about this, curious if you have thoughts to justify either way.

I just imagine as we build more admin functionality it's unlikely that everything is going to keep extending these fairly large FHIR models.

eric-gunther (changes requested):
Very excited about this feature! Concerned about pagination and performance of the patient history API, I don't think we need to get it perfect on the first try but at least would be good to have some safety checks in place to avoid requesting too much data.

src/fhir/models/patient-history.ts
luthfib: Good point. I think it would make more sense to have this perhaps take a patient model and the history info

bryanjknight (commented):
LGTM but probably need a original dev to approve. Also a good exercise for me to see what I missed in a given PR

src/fhir/models/patient-history.ts
```diff
@@ -1,29 +1,45 @@
...
-import { FHIRModel } from "./fhir-model";
 import { PatientModel } from "./patient";
-import { PatientHistoryResponse } from "@/components/content/patient-history/use-patient-history";
+import { formatDate, formatISODateStringToDate } from "../formatters";
 import { PatientRefreshHistoryMessage } from "@/services/patient-history/patient-history-types";
 
-export class PatientHistoryPatient extends FHIRModel<PatientModel> {
-  kind = "PatientHistoryPatient" as const;
+export class PatientHistorytModel {
+  kind = "PatientHistorytModel" as const;
```
eric-gunther, line 6: I'm assuming this is a typo, right?

eric-gunther (dismissed):

src/services/patient-history/patient-history-types.ts
```diff
@@ -1,12 +1,19 @@
+export type PatientHistoryStatus =
+  | "initialize"
+  | "in_progress"
+  | "done"
+  | "error";
```
pmadria3, line 5: also there is a "done-with-error" state as well right?

src/fhir/models/patient-history.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.historyInfo = historyInfo as PatientRefreshHistoryMessage;
+  }
+
+  get messages() {
+    // eslint-disable-next-line no-underscore-dangle
+    return this.historyInfo?._messages;
+  }
+
+  get key() {
+    // eslint-disable-next-line no-underscore-dangle
```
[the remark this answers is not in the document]
luthfib: We do, the table uses it

src/components/content/conditions/story-helpers/mocks/patient-history-message.ts
```diff
@@ -20,7 +20,6 @@ export const patientHistoryMessage = {
       ],
       _errors: [],
       _createdAt: "2022-11-28T14:29:25.509Z",
-      _updatedAt: "2022-11-28T14:29:27.436Z",
```
pmadria3, line 23: was this meant to be deleted?

pmadria3 (changes requested):
Looking good - just left a few comments 

Also do we need tests or followup?

src/fhir/formatters.ts
```diff
@@ -121,3 +121,12 @@ export function formatISODateStringToDate(
 
   return `${month}/${day}/${year}`;
 }
+
```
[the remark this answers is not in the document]
luthfib: Yeah, lets leave this one. The other one didnt quite work for the format I was using and this new one is a bit more flexible. If anything we could remove the old one by use by use case.

src/components/content/conditions/story-helpers/mocks/patient-history-message.ts
luthfib: Yes, it is not a field anymore

src/fhir/models/patient-history.ts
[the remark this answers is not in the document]
luthfib: We do need it. Or else eslint complains

pmadria3 (approved):

PR #604, opened 2023-03-22
CTW-802/show hide dismiss

WesleyKapow:
CTW-802

1. Get "Show Dismissed Records" showing up below the divider.
2. Get "Hide Dismissed Records" showing up when show is active.

Steps to get this working:
1. Simplified how divider is added, now `ListBox` can take a `{divider: true}` item and put a divider in that spot.
2. Renamed `belowTheFold` to `belowDivider` and added that to the newly refactored `dismissFilter`.
3. Added new `toggleDisplay` and `toggleIcon` options which will be displayed for active filters.
4. Changed `addOrRemoveFilter` to `toggleFilter` so now it knows whether this filter should be added or removed based on the current status.

Screenshot 2023-03-22 at 11 56 21 AM

Screenshot 2023-03-22 at 11 57 03 AM

### DIFF
--- a/.changeset/grumpy-chicken-move.md
+++ b/.changeset/grumpy-chicken-move.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Tweak "Show Dismissed Records" filter to be below divider and to then show up as "Hide Dismissed Records" when it has been applied.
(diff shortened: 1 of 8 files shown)


luthfib (approved):
LGTM. Works as expected.

PR #620, opened 2023-03-31
Ctw 748/allergies update

pmadria3:
CTW-748

This PR adds updates to allergies table, sorting/filtering, and updating the details drawer. 

Screenshot 2023-04-03 at 1 51 42 PM


src/components/content/allergies/allergies-filter.tsx
```diff
@@ -11,10 +17,8 @@ export const applyAllergyFilters = (data: fhir4.AllergyIntolerance[]) => {
...
 const valuesToDedupeOn = (allergy: AllergyModel) => [
-  allergy.categories,
-  allergy.clinicalStatus,
   allergy.display,
-  allergy.manifestations,
-  allergy.onset,
-  allergy.type,
+  allergy.recordedDate,
+  allergy.managingOrganization,
+  allergy.knownCodings,
```
eric-gunther, line 23: The ticket AC says to de-dupe by code, preferring RxNorm over NDC over SNOMED, using the display value. So I think simply de-duping on `allergy.display` was correct, but we'll want to adjust that getter (will provide that feedback in a comment on that area of the code).

src/App.tsx
```diff
@@ -105,7 +105,12 @@ const demoComponents: DemoComponent[] = [
     render: () => <PatientDocuments />,
     title: "Patient Documents",
   },
-  { render: () => <PatientAllergies />, title: "Patient Allergies" },
+  {
+    render: () => (
+      <ZusAggregatedProfile resources={["allergies"]} title="Allergies" />
+    ),
+    title: "Patient Allergies",
+  },
```
eric-gunther, line 113: Non-blocking feedback, but this whole App.tsx is starting to get a bit messy again. Perhaps we could just use a single instance of `ZusAggregatedProfile` where folks can play around with the `resources` prop.

src/components/content/allergies/helpers/history.tsx
```diff
@@ -0,0 +1,65 @@
...
+  );
+
+  const searchParams: SearchParams = {
+    _include: ["AllergyIntolerance:patient"],
+  };
+
+  // If we have any known codings, then do an OR search.
+  // Otherwise fall back to searching for this single condition.
+  // That way, conditions that don't have any good codes to match on
+  // will only show themselves in the history.
```
eric-gunther, line 32: Copy/paste comment still referencing conditions.

src/components/content/allergies/helpers/history.tsx
```diff
@@ -0,0 +1,65 @@
...
+  if (tokens.length > 0) {
+    searchParams.code = tokens.join(",");
+  } else {
+    // eslint-disable-next-line no-underscore-dangle
+    searchParams._id = allergy.id;
+  }
```
eric-gunther, line 38: You should investigate the performance of these queries by code in production, I suspect they are going to be slow (which I think I noticed for conditions already in prod).

src/components/content/allergies/helpers/history.tsx
```diff
@@ -0,0 +1,65 @@
...
+    date: allergy.recordedDate,
+    versionId: allergy.versionId,
+    title: allergy.managingOrganization,
+    details: detailData,
+  };
+}
+
+const valuesToDedupeOn = (allergy: AllergyModel) => [
+  allergy.knownCodings.map((coding) => [coding.system, coding.code]),
+];
```
eric-gunther, line 65: These are the values to de-dupe on in the history, right? So I would expect this to include the date field and perhaps others.

eric-gunther (changes requested):

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -1,12 +1,19 @@
+/* eslint-disable import/order */
```
pmadria3, line 1: eslint messed up in editor so had to add this - please can someone remove this

src/fhir/models/allergies.ts
```diff
@@ -1,6 +1,11 @@
+/* eslint-disable import/order */
```
pmadria3, line 1: same as prev comment

src/App.tsx
```diff
@@ -63,52 +50,12 @@ const demoComponents: DemoComponent[] = [
...
-    title: "Patient Conditions",
-  },
-  {
-    render: () => <PatientDocuments />,
-    title: "Patient Documents",
-  },
-  { render: () => <PatientAllergies />, title: "Patient Allergies" },
-  { render: () => <PatientImmunizations />, title: "Patient Immunizations" },
-  { render: () => <PatientSearch />, title: "Patient Search" },
-  { render: () => <PatientTimelineV2 />, title: "Patient Timeline" },
```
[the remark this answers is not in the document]
pmadria3: Eric had a comment on it above - we can just do one instance and then pass the resource and use that

luthfib (approved):

PR #622, opened 2023-03-31
CTW-865/generic history meds

WesleyKapow:
Refactor medication component to use generic details drawer.

TODO (follow ups?):
1. For some reason the `useLastPrescriber` always fetches history even if react-query has it cached. You'll notice in storybook that opening the same drawer again, the prescriber line always has a loading spinner while the history does not as it uses the cache. Not sure what is going on here :'(
2. The way timeline 2.0 takes a dispense or request model, fetches a statement, and opens the medication drawer (using the statement) is a bit wonky with its use of the `MedicationDrawerComponent` helper in `timeline-2.0/helpers/modal-hooks.tsx`. The tricky part here is the use of `useQueryMedicationStatement` .. instead we should have just a function that can do that (and use react-query client directly) and not a hook. That way we can call it outside of a hook. See `getBinaryDocumentReq` in `fhir/binaries.ts` for an example of using `queryClient.fetchQuery`.

Screenshot 2023-03-31 at 2 20 02 PM

History consists of different types of resources (statements, dispense, requests, administrations), each with their own set of details (as before):
Screenshot 2023-03-31 at 2 20 18 PM


mrosata (approved):
I think this looks good.. I've run it locally as well. Don't see anything that stands out as an issue

pmadria3 (approved):
LGTM

PR #623, opened 2023-03-31
Quick fix for telemetry env and CareTeam table rows

mrosata:
CareTeamPractitionerModel now uses "key" for table row identification.


WesleyKapow (approved):
LGTM

PR #624, opened 2023-03-31
CTW-878/Patient History button in zap

mrosata:
Show request records button in top right of ZAP.
Fixes issue where tables show infinite loading spinner if patient is not found. 

Screen Shot 2023-03-31 at 3 51 04 PM

Screen Shot 2023-03-31 at 4 00 42 PM


src/components/content/resource/resource-table.tsx
```diff
@@ -30,10 +32,19 @@ export const ResourceTable = <
...
   const shouldShowTableHead =
     typeof showTableHead === "boolean" ? showTableHead : !breakpoints.sm;
+  const emptyMessageWithRequestRecords = patient.data ? (
+    emptyMessage
+  ) : (
+    <div className="ctw-space-y-4">
+      Patient not found. Please fill out a short form to{" "}
+      <RequestRecordsButton className="ctw-lowercase" />.
+    </div>
+  );
```
mrosata, line 47: Will likely be removing this. I thought this was part of the AC but it is not.

src/utils/telemetry.ts
```diff
@@ -81,13 +81,13 @@ export class Telemetry {
...
-    if (["dev", "development"].includes(environment.toLowerCase())) {
+    this.environment = environment.toLowerCase();
+    if (["dev", "development"].includes(this.environment)) {
       this.environment = "dev";
     }
-    if (["prod", "production"].includes(environment.toLowerCase())) {
+    if (["prod", "production"].includes(this.environment)) {
       this.environment = "prod";
     }
-    this.environment = environment;
```
mrosata, line 90: This was causing the environment to be overwritten by the passed in value everytime 🤦🏼

PR #640, opened 2023-04-07
Add patient history request status to the ZusAggregatedProfile

luthfib:
- Show patient history status in the mega zap component
- If an error occurs, notify the user that the request was not completed

image

### DIFF
--- a/src/components/content/patient-history/patient-history-last-retrieved.tsx
+++ b/src/components/content/patient-history/patient-history-last-retrieved.tsx
@@ -1,4 +1,5 @@
 import { usePatientHistory } from "./use-patient-history";
+import { formatISODateStringToDate } from "@/fhir/formatters";
 
 export const PatientHistoryLastRetrieved = () => {
   const patientHistory = usePatientHistory();
@@ -7,7 +8,8 @@ export const PatientHistoryLastRetrieved = () => {
     <>
       {patientHistory.lastRetrievedAt && (
         <div className="ctw-text-sm ctw-italic ctw-text-black">
-          Last Retrieved {patientHistory.lastRetrievedAt}
+          Last Retrieved{" "}
+          {formatISODateStringToDate(patientHistory.lastRetrievedAt)}
         </div>
       )}
     </>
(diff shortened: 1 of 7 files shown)


src/components/content/patient-history/patient-history-last-retrieved.tsx
```diff
@@ -7,7 +8,8 @@ export const PatientHistoryLastRetrieved = () => {
     <>
       {patientHistory.lastRetrievedAt && (
         <div className="ctw-text-sm ctw-italic ctw-text-black">
-          Last Retrieved {patientHistory.lastRetrievedAt}
+          Last Retrieved{" "}
```
pmadria3, line 11: is this through prettier?

pmadria3:
The date looks a little weird in your screenshot

src/components/content/patient-history/patient-history-message-status.tsx
```diff
@@ -23,20 +33,26 @@ export const PatientHistoryStatus = ({
...
-            <div className="ctw-font-normal ctw-text-error-text">
-              Contact your organization’s technical support if this issue
-              persists for more than 24 hours.
-            </div>
-          </div>
-        </div>
-      );
+      return <ErrorState />;
+    case "done_with_errors":
+      return <ErrorState />;
```
pmadria3, line 38: so you can use the case cascade effect: 

so you can do 

```
case blah: 
case blah: 
       return blah 
```

so you can essentially remove the return <ErrorState />

src/components/content/patient-history/patient-history-message-status.tsx
```diff
@@ -23,20 +33,26 @@ export const PatientHistoryStatus = ({
...
     case "done":
+      return servicesStatus ? null : <ErrorState />;
```
pmadria3, line 40: also with done you can just in ErrorState just return null so you don't need the check

pmadria3: and can cascade to default

PR #642, opened 2023-04-09
Ctw 903/open multiple drawers

luthfib:
Fix bug where opening successive drawers outside of zap component causes apps to crash.


pmadria3 (approved):
LGTM

PR #646, opened 2023-04-11
CTW-924 Sort Allergies by Name

bryanjknight:


eric-gunther:
@bryanjknight for most feature improvements like this we'll do a patch bump - so run `npx changeset` and then select "patch" (we try to abide by semver).

eric-gunther (approved):

PR #663, opened 2023-04-18
Kludge attempt

pmadria3:
CTW-855

This PR fixes the post and pre kludge by adding a conditional to make sure based on environment the data is being obtained in the old/new fashion.

### DIFF
--- a/.changeset/old-eels-arrive.md
+++ b/.changeset/old-eels-arrive.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Fix search for pre and post kludge for different environments.
--- a/.changeset/old-eels-arrive.md
+++ b/.changeset/old-eels-arrive.md
@@ -2,4 +2,4 @@
 "@zus-health/ctw-component-library": patch
 ---
 
-Fix search for pre and post kludge for different environments.
+Update search helpers to support Zus refactoring to store 3rd party data in dedicated builders.
--- a/src/api/urls.ts
+++ b/src/api/urls.ts
@@ -27,3 +27,16 @@ export const getMetricsBaseUrl = (env: string) => {
     ? `https://ctw.zusapi.com`
     : `https://ctw.${env}.zusapi.com`;
 };
+
+export const getLensUrl = (env: Env) => {
+  switch (env) {
+    case "dev":
+      return "bc1c791a-581c-4117-8b17-f9c59f39caf8";
+    case "sandbox":
+      return "d8ab3e86-7ff2-482d-bbed-b30df2dd7ec7";
+    case "production":
+      return "5e0f86eb-bb0b-4543-8489-262fcbf661da";
+    default:
+      throw new Error("Unknown environment");
+  }
+};
--- a/src/api/urls.ts
+++ b/src/api/urls.ts
@@ -28,14 +28,16 @@ export const getMetricsBaseUrl = (env: string) => {
     : `https://ctw.${env}.zusapi.com`;
 };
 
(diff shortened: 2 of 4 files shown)


src/api/urls.ts
```diff
@@ -27,3 +27,16 @@ export const getMetricsBaseUrl = (env: string) => {
     ? `https://ctw.zusapi.com`
     : `https://ctw.${env}.zusapi.com`;
 };
+
+export const getLensUrl = (env: Env) => {
+  switch (env) {
```
sarahmangels, line 32: add phitest

src/api/urls.ts
```diff
@@ -27,3 +27,16 @@ export const getMetricsBaseUrl = (env: string) => {
     ? `https://ctw.zusapi.com`
     : `https://ctw.${env}.zusapi.com`;
 };
+
+export const getLensUrl = (env: Env) => {
```
sarahmangels, line 31:
```suggestion
export const getLensBuilderId = (env: Env) => {
```

src/fhir/search-helpers.ts
```diff
@@ -146,11 +147,18 @@ export async function searchLensRecords<T extends ResourceTypeString>(
   });
   const records = await searchAllRecords(resourceType, requestContext, params);
 
-  const { entry, resources } = filterSearchReturnByBuilderId(
+  let { entry, resources } = filterSearchReturnByBuilderIdorLensId(
```
sarahmangels, line 150: nit: you could leave the name of this as `filterSearchReturnByBuilderId` because its still filtering by builderId in both case. Just that in the post-kludge way you're using the default lens builder ID instead of the users builder ID

.changeset/old-eels-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Fix search for pre and post kludge for different environments.
```
eric-gunther, line 5: Non-blocking but I think you should be a bit more explicit about what was done here, I think we over-use the term "kludge".

```suggestion
Update search helpers to support Zus refactoring to store 3rd party data in dedicated builders.
```

src/fhir/search-helpers.ts
eric-gunther: +1 on this, you're always filtering by builder ID, whether that's the user's builder or the lens builder is determined by an input to this function.

src/fhir/search-helpers.ts
```diff
@@ -146,11 +147,18 @@ export async function searchLensRecords<T extends ResourceTypeString>(
...
-    requestContext.builderId
+    getLensUrl(requestContext.env)
   );
 
+  if (resources.length === 0 && entry.length === 0) {
+    ({ entry, resources } = filterSearchReturnByBuilderIdorLensId(
+      records,
+      requestContext.builderId
+    ));
+  }
```
eric-gunther, line 160: I don't think what's going on here is all that intuitive so I think it's worth a comment explaining that you are filtering by the lens builder to see if you get back data from that builder indicating that we are in the post-kludge world. If you get back no data when filtering by lens then either we are in the pre-kludge world and still need to filter by the user's builder ID to avoid potentially getting duplicates records (if the patient exists in multiple builders), or the patient simply doesn't have any data to return.

It's all a bit head-spinning and at some point we're going to be firmly in the post-kludge world and we'll be able to refactor this to ignore the requirements of the pre-kludge times. So you may also want to put in a comment that clearly indicates what can be removed once we've been in post-kludge time long enough.

sarahmangels (approved):

PR #664, opened 2023-04-19
Ctw 974/update post endpoint

luthfib:
Update endpoint to schedule patient history jobs to newer JSON compliant endpoint.


src/api/patient-history.ts
```diff
@@ -19,21 +20,42 @@ export const schedulePatientHistory = async (
...
+        practitioner: {
+          npi: resultData.npi,
+          name: resultData.name,
+          role: resultData.role.toLocaleLowerCase(),
+        },
+      },
+      relationships: {
+        patient: {
+          // If patientID is empty, just pass any non-identifying string in url.
+          data: { type: "fhir/Patient", id: resultData.id ?? "NULL" },
```
pmadria3, line 39: Nit: Could we just leave it as an empty string? ""

pmadria3 (approved):
LGTM

src/api/patient-history.ts
luthfib: I think we pass null because sometimes when you pass empty string it actually omits the patient id filter which can mean we will actually get a bunch of patients back because it will just filter by the builder id

pmadria3: ah good point - okay disregard

src/api/patient-history.ts
```diff
@@ -19,21 +20,42 @@ export const schedulePatientHistory = async (
...
+  };
 
   try {
     const response = await ctwFetch(endpointUrl, {
       method: "POST",
       headers: {
+        "ehr-data-integration-proxy": JSON.stringify({
+          "patient-id": patientID,
+          "system-url": systemURL,
+        }),
```
mrosata, line 52:
```suggestion
        "ehr-data-integration-proxy": JSON.stringify({
          patientID,
          systemURL,
        }),
```

src/api/patient-history.ts
mrosata: actually you can set this to an empty string. Originally we had set this to NULL so the url string parsing had something to match on.. but now that it's not in the URL it can be empty. In the case where it is empty (ie: elation patient history request for a patient that doesn't exist yet) we fill in the value with the ID of the patient we create in the proxy endpoint.

The logic we'll be using on the backend is now essentially going to be:
```ts
const { patientID, systemURL } = JSON.parse(headers['ehr-data-integration-proxy']);
if (systemURL === ELATION_URL) {
  // create patient in zus and backfill from elation
}
const patient = await zusApi.getPatient(patientID, systemURL);
body.data.relationships.patient.data.id = body.data.relationships.patient.data.id || patient.id;

callPatientHistoryJobsApi(body);
```
so having that set to "NULL" will actually become problematic in this new pt hx api

mrosata (approved):

PR #671, opened 2023-04-25
wip: poc of allergies on fqs

bryanjknight:


luthfib:
I don't think this is something we should do in this PR, but if we can figure out how to generate types from these queries, we should. I know there should be a way so that when we get data back we should be able to know exactly what is coming back

luthfib:
Also we will want to start mocking FQS endpoints

PR #673, opened 2023-04-26
Prettier/PrintLine 100

mrosata:
This bumps line length up from 80 to 100. Decided not to make a changeset on this because it shouldn't affect the build output.


luthfib (approved):
Wow you really got a lot of work done

PR #674, opened 2023-04-26
Fix documentRef in table

pmadria3:
CTW-907

This PR fixes to check if more than one coding exist to only populate those in the table.


src/components/content/document/patient-document-filters.tsx
```diff
@@ -3,7 +3,17 @@ import { isEqual, uniqWith } from "@/utils/nodash";
...
   const documentModel = data.map((document) => new DocumentModel(document));
-  const documentData = uniqWith(documentModel, (a, b) =>
+
+  const actualDocument = documentModel.filter(
+    (document) =>
+      document.category &&
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      document.category.filter(
+        (category) => category.coding && category.coding.length > 1
+      )
```
luthfib, line 13:
```suggestion
 document.category.filter(
        (category) => category.coding && category.coding.length > 1
      ).length > 0
```

I believe you want something like this. The reason eslint is warning you about this is because since you already checked if document.category is truthy and in the next line you re doing the same thing. You probably want to check that it has actual elements inside of it

PR #678, opened 2023-04-28
CTW-1006 Updated metrics

eric-gunther:
This PR adds a few metrics:

- `component.{component name}.failure` will indicate instances of a component failing to load. This will help us report out on how many times a user got an error when trying to pull up the ZAP.
- `component.{component name}.loaded` will indicate how many times a user pulled up the ZAP. This may be a noisy metric as this is based on `componentDidMount`, so it's possible the component was loaded but wasn't actually visible to the user. But this is a good start.
- `records.{record type}` (eg. `records.outside_medications`) will capture how many records were loaded for a given component in the ZAP. We'll use this to capture things like "on average users see N outside medications in the ZAP" and will also help to identify the number of instances where the ZAP was empty for a given resource type. Right now this PR only adds that metric for medications. If there's no opposition to this approach I'll add the rest.

These metrics are sent to a Zus endpoint and not directly to DataDog from the client. Our plan moving forward is to remove the RUM telemetry from this repo (which requires opting in to enable telemetry) and just use that for the standalone Zus app.

### DIFF
--- a/src/components/content/allergies/patient-allergies.tsx
+++ b/src/components/content/allergies/patient-allergies.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import { useRef } from "react";
+import { useEffect, useRef } from "react";
 import { allergyFilter, defaultAllergyFilters } from "./helpers/filters";
 import { useAllergiesHistory } from "./helpers/history";
 import { allergySortOptions, defaultAllergySort } from "./helpers/sort";
@@ -13,6 +13,7 @@ import { usePatientAllergies } from "@/fhir/allergies";
 import { AllergyModel } from "@/fhir/models/allergies";
 import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
 import { capitalize } from "@/utils/nodash";
+import { Telemetry } from "@/utils/telemetry";
 
 export type PatientAllergiesProps = {
   className?: string;
@@ -37,6 +38,12 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
 
   // Get our allergies.
   const { isLoading } = patientAllergiesQuery;
+
+  useEffect(() => {
+    if (!isLoading) {
+      Telemetry.reportZAPRecordCount("allergies", data.length);
+    }
+  }, [isLoading, data]);
 
   return (
     <div
--- a/src/components/content/allergies/patient-allergies.tsx
+++ b/src/components/content/allergies/patient-allergies.tsx
@@ -41,9 +41,9 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
 
   useEffect(() => {
     if (!isLoading) {
-      Telemetry.reportZAPRecordCount("allergies", data.length);
+      Telemetry.reportZAPRecordCount("allergies", patientAllergiesQuery.data?.length);
     }
-  }, [isLoading, data]);
+  }, [isLoading, patientAllergiesQuery.data]);
 
   return (
     <div
--- a/src/components/content/allergies/patient-allergies.tsx
+++ b/src/components/content/allergies/patient-allergies.tsx
@@ -40,8 +40,8 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
   const { isLoading } = patientAllergiesQuery;
 
   useEffect(() => {
-    if (!isLoading) {
-      Telemetry.reportZAPRecordCount("allergies", patientAllergiesQuery.data?.length);
+    if (!isLoading && patientAllergiesQuery.data) {
+      Telemetry.reportZAPRecordCount("allergies", patientAllergiesQuery.data.length);
     }
   }, [isLoading, patientAllergiesQuery.data]);
 
--- a/src/components/content/conditions/patient-conditions.tsx
+++ b/src/components/content/conditions/patient-conditions.tsx
@@ -24,8 +24,8 @@ const PatientConditionsComponent = ({ className, readOnly = false }: PatientCond
   const { t } = useBaseTranslations();
 
   useEffect(() => {
-    if (!query.isLoading) {
-      Telemetry.reportZAPRecordCount("builder_conditions", query.data?.length);
+    if (!query.isLoading && query.data) {
+      Telemetry.reportZAPRecordCount("builder_conditions", query.data.length);
     }
   }, [query.isLoading, query.data]);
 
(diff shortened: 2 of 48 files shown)


mrosata (approved):
lgtm so far

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -38,6 +39,12 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
   // Get our allergies.
   const { isLoading } = patientAllergiesQuery;
 
+  useEffect(() => {
+    if (!isLoading) {
+      Telemetry.reportZAPRecordCount("allergies", data.length);
+    }
+  }, [isLoading, data]);
+
```
eric-gunther, line 47: It seems appropriate to put this logic in the actual component since our intention is to capture the count of records displayed in the component. Could've moved the logic further down into `usePatientAllergies` but then it's possible that query gets used elsewhere and we don't necessarily want to capture those towards the count of records actually presented in a ZAP.

I may consider moving the logic out of use-medications.ts and in to the medication component as well.

mrosata: Yeah I think if we're not going to put this one in `usePatientAllergies` then we shouldn't have the medications be in a hook either.

mrosata: Also good to note, if you don't want this to resend a new count metric every-time the user sorts or filters the table, then the useEffect should target `patientAllergiesQuery.data` instead of `data`

src/components/content/conditions/patient-conditions.tsx
```diff
@@ -21,6 +23,12 @@ const PatientConditionsComponent = ({ className, readOnly = false }: PatientCond
   const showAddConditionForm = useAddConditionForm();
   const { t } = useBaseTranslations();
 
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("builder_conditions", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 30: We probably don't want to call `reportZAPRecordCount` unless `query.data` exists
```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("builder_conditions", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

src/components/content/document/patient-documents.tsx
```diff
@@ -16,15 +17,20 @@ export type PatientDocumentProps = {
...
 
-  const document = patientDocumentQuery.data ?? [];
-  const { isLoading } = patientDocumentQuery;
+  const document = query.data ?? [];
+
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("documents", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 33: same here (checking query.data exists before calling `reportZAPRecordCount`
```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("documents", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

src/components/content/immunizations/patient-immunizations.tsx
```diff
@@ -23,13 +24,19 @@ function PatientImmunizationsComponent({ className }: PatientImmunizationsProps)
...
     header: (m) => m.description,
     details: immunizationData,
     getSourceDocument: true,
   });
 
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("immunization", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 38: same here (checking query.data exists before calling reportZAPRecordCount

```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("immunization", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

src/components/content/observations/patient-observations-outside.tsx
```diff
@@ -1,18 +1,26 @@
...
 
 const Component = ({ className }: PatientObservationsOutsideProps) => {
-  const diagnosticReports = usePatientDiagnosticReportsOutside();
+  const query = usePatientDiagnosticReportsOutside();
+
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("outside_observations", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 19: same here (checking query.data exists before calling reportZAPRecordCount

```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("outside_observations", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

src/components/content/observations/patient-observations.tsx
```diff
@@ -1,16 +1,24 @@
...
 const Component = ({ className }: PatientObservationsProps) => {
-  const diagnosticReports = usePatientDiagnosticReports();
+  const query = usePatientDiagnosticReports();
 
-  return <PatientObservationsBase className={className} query={diagnosticReports} />;
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("builder_observations", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 19: same here (checking query.data exists before calling reportZAPRecordCount

```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("builder_observations", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

src/components/content/timeline/patient-timeline.tsx
```diff
@@ -1,29 +1,37 @@
...
+    query.data ?? [],
     [(encounter) => encounter.resource.period?.start ?? ""],
     ["desc"]
   );
 
+  useEffect(() => {
+    if (!query.isLoading) {
+      Telemetry.reportZAPRecordCount("encounters", query.data?.length);
+    }
+  }, [query.isLoading, query.data]);
```
mrosata, line 29: same here (checking query.data exists before calling reportZAPRecordCount

```suggestion
  useEffect(() => {
    if (!query.isLoading && query.data) {
      Telemetry.reportZAPRecordCount("encounters", query.data.length);
    }
  }, [query.isLoading, query.data]);
```

mrosata (approved):

mrosata (approved):

PR #685, opened 2023-05-01
Ctw 974/fix filter bug

luthfib:
Fix bug where in_progress and done_with_errors don't filter correctly in patient history table.


pmadria3 (approved):
LGTM

PR #693, opened 2023-05-02
fix: filter out lenses post query as opposed to in ODS query

bryanjknight:
This fix no longer filters out lens data as part of the FHIR query but rather filters out the lenses post query. This reduce query times from 10s to 100ms

### DIFF
--- a/src/components/content/conditions/helpers/history.tsx
+++ b/src/components/content/conditions/helpers/history.tsx
@@ -1,13 +1,13 @@
+import { Condition } from "fhir/r4";
 import { SearchParams } from "fhir-kit-client";
 import { HistoryEntryProps } from "../../resource/helpers/history-entry";
 import { useHistory } from "../../resource/history";
 import { CodingList } from "@/components/core/coding-list";
 import { NotesList } from "@/components/core/notes-list";
 import { ConditionModel } from "@/fhir/models";
+import { SYSTEM_SUMMARY, SYSTEM_ZUS_LENS } from "@/fhir/system-urls";
 import { capitalize, startCase } from "@/utils/nodash";
 import { QUERY_KEY_CONDITION_HISTORY } from "@/utils/query-keys";
-import { Condition } from "fhir/r4";
-import { SYSTEM_SUMMARY, SYSTEM_ZUS_LENS } from "@/fhir/system-urls";
 
 export function useConditionHistory(condition: ConditionModel) {
   return useHistory({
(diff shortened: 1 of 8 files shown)


src/components/content/conditions/helpers/history.tsx
```diff
@@ -15,9 +17,35 @@ export function useConditionHistory(condition: ConditionModel) {
...
+  if (!metaTags) {
+    return true;
+  }
+
+  const lenTags = metaTags.filter((tag) => 
+    isChronicConditionLens(tag.system, tag.code) ||
+    isSummaryConditionLens(tag.system, tag.code)
+  )
+
+  const hasLensTags = lenTags.length > 0
```
pmadria3:
```suggestion
  const lenTags = metaTags.filter((tag) => 
    isChronicConditionLens(tag.system, tag.code) ||
    isSummaryConditionLens(tag.system, tag.code)
  ).length > 0
```

Nit: you can combine the .length and then just do the not of that and return it

bryanjknight: yeah, I went with your suggestion but also like decoupling what's returned from the business logic. If we wanted to do additional filtering, we'd just have to add another boolean to the return value. Certainly YAGNI though so game for taking the suggestion

pmadria3 (dismissed):
LGTM - not sure if Eric is also reviewing or not

src/fhir/search-helpers.ts
```diff
@@ -199,16 +199,24 @@ export async function searchSummaryRecords<T extends ResourceTypeString>(
...
 
-// Like searchAllRecords, but filters out lens resources.
+// Like searchAllRecords, but filters out lens resources. If postQueryFilter
+// is specified, exclusion will happen client side, not server side
 export async function searchCommonRecords<T extends ResourceTypeString>(
   resourceType: T,
   requestContext: CTWRequestContext,
-  searchParams?: SearchParams
+  searchParams?: SearchParams,
+  postQueryFilter?: (r: ResourceType<T>) => boolean
```
eric-gunther, line 208: I don't think `postQueryFilter` should be an input here, `searchCommonRecords` should simply filter out lens resources as part of its default behavior with no other options.

src/components/content/conditions/helpers/history.tsx
```diff
@@ -15,9 +17,33 @@ export function useConditionHistory(condition: ConditionModel) {
...
+    return true;
+  }
+
+  const hasLensTags =
+    metaTags.filter(
+      (tag) =>
+        isChronicConditionLens(tag.system, tag.code) || isSummaryConditionLens(tag.system, tag.code)
+    ).length > 0;
+
+  return !hasLensTags;
```
eric-gunther, line 44: Per the other comment, I think all this logic doesn't belong in history.tsx, I think it belongs inside `searchCommonRecords`.

eric-gunther (changes requested):

src/fhir/search-helpers.ts
bryanjknight: Gotcha, so just filter out any lens (ActiveMedication, ChronicCondition, Summary, etc)

src/fhir/search-helpers.ts
```diff
@@ -199,16 +199,33 @@ export async function searchSummaryRecords<T extends ResourceTypeString>(
   return records;
 }
 
-// Like searchAllRecords, but filters out lens resources.
+// Like searchAllRecords, but filters out lens resources. If postQueryFilter
```
eric-gunther, line 202:
```suggestion
// Like searchAllRecords, but filters out lens resources.
```

src/fhir/search-helpers.ts
```diff
@@ -199,16 +199,33 @@ export async function searchSummaryRecords<T extends ResourceTypeString>(
...
+// is specified, exclusion will happen client side, not server side
```
eric-gunther, line 203:
```suggestion
```

eric-gunther (approved):

PR #694, opened 2023-05-02
CTW-1006 Metrics for component loading, crashing and stats on record counts

eric-gunther:
- Added metric for component failures
- Added metrics for counts of records returned by Zus
- Removed interaction tracking (we may return to this someday but right now it's a lot of noise that we aren't getting value out of)
- A bunch of renaming to help distinguish builder records vs summary records vs CPR records

### DIFF
--- a/src/components/content/conditions/patient-conditions.tsx
+++ b/src/components/content/conditions/patient-conditions.tsx
@@ -7,7 +7,7 @@ import {
 import { PatientConditionsBase } from "./helpers/patient-conditions-base";
 import { withErrorBoundary } from "@/components/core/error-boundary";
 import { RowActionsProps } from "@/components/core/table/table";
-import { usePatientConditions } from "@/fhir/conditions";
+import { usePatientBuilderConditions } from "@/fhir/conditions";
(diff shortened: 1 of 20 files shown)


src/components/content/conditions/patient-conditions.tsx
```diff
@@ -7,7 +7,7 @@ import {
 import { PatientConditionsBase } from "./helpers/patient-conditions-base";
 import { withErrorBoundary } from "@/components/core/error-boundary";
 import { RowActionsProps } from "@/components/core/table/table";
-import { usePatientConditions } from "@/fhir/conditions";
```
eric-gunther, line 10: There's a fair amount of renaming in here to distinguish getting "builder records" (eg. the conditions written by the user's builder) from getting "summary records" (eg. the aggregated/summarized conditions written by our lens service) or "all records" (eg. all conditions for a patient across all builders as available through the Common Patient Record [CPR]). A lot of our code has references to outside records when we're actually dealing with summary records.

bryanjknight (approved):

PR #695, opened 2023-05-02
[BUG] Loading State for Dismiss/Restore Buttons

Stephen-Tsen:
When the dismiss/restore button is clicked for patient outside conditions and medications, the button shows a loading state and the add button is disabled until the action is complete.

- Create `useAction` hook to help with chaining callbacks
- Create `useToggleArchive` hook to help with providing input feedback to the user when the action is invoked
- Add styling for disabled primary and secondary buttons
- Update the patient outside conditions and medications components to provide feedback to the user when the dismiss/restore buttons are clicked. When it is clicked, both the dismiss/restore and add buttons are disabled until the action is completed. If it is successful, then the query is invalidated and rerun

image


src/i18n/index.ts
```diff
@@ -18,7 +18,10 @@ export const resources = {
       "resource.remove.body":
         "This will remove <2>{{resourceName}}</2> from this patient's {{resource}} list.",
       "resource.remove.heading": "Remove this {{resource}}?",
-      "resource.unamed": "unamed {{resource}}",
+      "resource.unnamed": "unnamed {{resource}}",
```
Stephen-Tsen, line 21: Please let me know if it was intentionally spelled `unamed`

src/fhir/basic.ts
```diff
@@ -51,24 +50,17 @@ export async function recordProfileAction<T extends fhir4.Resource>(
     Telemetry.reportActionFailure(profileAction);
     throw new Error(`Failed to ${profileAction} resource with id of ${model.id}`);
   } else {
-    Telemetry.reportActionFailure(profileAction);
+    Telemetry.reportActionSuccess(profileAction);
```
Stephen-Tsen, line 53: This looked wrong to me...

bryanjknight (changes requested):

src/components/content/hooks/use-toggle-archive.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    onError: () => {
+      setIsLoading(false);
+    },
+  });
+
+  return {
+    toggleArchive: handleUpdateArchive,
+    isLoading,
+  };
```
mrosata, line 54: The `useAction` pattern seems nice. Does it offer a benefit over (in this case) just using the built in `.then` and `.finally`?
```suggestion
  const { getRequestContext } = useCTW();
  const [isLoading, setIsLoading] = useState<boolean>(false);

  const handleUpdateArchive = useCallback(async () => {
    setIsLoading(true);

    await toggleArchive(model, await getRequestContext())
      .then(() => queryClient.invalidateQueries([queryToInvalidate]))
      .finally(() => setIsLoading(false));
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [model]);

  return {
    toggleArchive: handleUpdateArchive,
    isLoading,
  };
```

src/components/content/conditions/patient-conditions-outside.tsx
```diff
@@ -63,28 +64,36 @@ export const PatientConditionsOutside = withErrorBoundary(
 
 const RowActions = ({ record }: RowActionsProps<ConditionModel>) => {
   const showAddConditionForm = useAddConditionForm();
-  const { getRequestContext } = useCTW();
+  const { isLoading, toggleArchive } = useToggleArchive(
+    record,
+    QUERY_KEY_OTHER_PROVIDER_CONDITIONS
+  );
+  const archiveLabel = record.isArchived ? t("resourceTable.restore") : t("resourceTable.dismiss");
```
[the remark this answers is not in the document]
Stephen-Tsen: I have not. I'll update to using the `useBaseTranslations` hook

src/i18n/index.ts
eric-gunther: Elation isn't specifically setting an override for the string we display when a resource doesn't have a display value. They just use the glossary object to set overrides for `condition_one` and `condition_other` (the plural). I don't think this merits a formal deprecation or any real communication, I'd just fix the typo and call it a day.

bryanjknight: @eric-gunther is there any way for us to gauge what is being used across the EHR integrations when we need to rename other things?

src/components/content/conditions/helpers/modal-hooks.tsx
```diff
@@ -73,7 +73,7 @@ export function useConfirmDeleteCondition() {
   return (condition: ConditionModel, onDelete?: (condition: ConditionModel) => void) => {
```
pmadria3, line 73: This is just an other comment - need changeset. You can add one by doing `npx changeset` and then probably a minor in this case.

src/fhir/basic.ts
```diff
@@ -51,24 +50,17 @@ export async function recordProfileAction<T extends fhir4.Resource>(
...
   const existingBasic =
     model.getBasicResourceByAction("archive") || model.getBasicResourceByAction("unarchive");
   const profileAction = model.isArchived ? "unarchive" : "archive";
 
   await recordProfileAction(existingBasic, model, requestContext, profileAction);
-
-  // Refresh our data (this is really just needed to update
-  // otherProviderRecord state).
-  if (queriesToInvalidate) {
-    await queryClient.invalidateQueries(queriesToInvalidate);
```
Stephen-Tsen, line 72: This was moved to `useToggleArchive` because invalidating the query can cause the component (row in the table) to be unmounted before the update to remove the spinner causing a react error.

src/components/content/hooks/use-toggle-archive.ts
```diff
@@ -0,0 +1,52 @@
...
+  queryToInvalidate: string
+): UseToggleArchiveResult {
+  const { getRequestContext } = useCTW();
+
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const handleToggleArchive = useCallback(async () => {
+    setIsLoading(true);
+
+    toggleArchive(model, await getRequestContext()).then(
```
WesleyKapow, line 36: nit: favor using async/await and not `.then`.

src/components/content/medications/patient-medications-outside.tsx
```diff
@@ -52,28 +53,37 @@ const getRowActions =
...
       >
-        {record.isArchived ? "Restore" : "Dismiss"}
+        {isLoading ? (
+          <div className="ctw-flex">
+            <Spinner className="ctw-mx-4 ctw-align-middle" />
+          </div>
+        ) : (
+          archiveLabel
+        )}
       </button>
```
WesleyKapow, line 79: Could refactor this into a re-usable component to DRY things up.

WesleyKapow (approved):
Really great work here! Using this minor review as a PING, what's going on with this PR? 2 weeks old!? Let's get this in!

src/components/content/hooks/use-toggle-archive.ts
Stephen-Tsen: I didn't want to invalidate the query if the toggle failed, hence why there's the success and failure functions that are ever so slighly different.

src/components/content/medications/patient-medications-outside.tsx
Stephen-Tsen: I started down that path but do not remember why I reverted to this...

PR #700, opened 2023-05-04
fix: address data issues with table view

bryanjknight:
This addresses the following issues:
1. When diagnostic reports had no effective date, the date was blank in the table. This now will take the earliest start date as described from the associated observations
2. When diagnostic reports had no `code.text`, the display field was blank. This now leverages enrichment values as well as other coding display values to provide a better display value for the diagnostic report

### DIFF
--- a/src/fhir/models/diagnostic-report.test.ts
+++ b/src/fhir/models/diagnostic-report.test.ts
@@ -1,0 +1,70 @@
+import { Observation } from "fhir/r4";
+import * as dr from "./diagnostic-report";
+
+const syntheticObservation = (): Observation => ({
+  resourceType: "Observation",
+  status: "unknown",
+  code: {},
+});
+
+// This is a workaround to having to build a separate builder pattern for the entire
+// observation
+const set = <T>(field: string, x: T, v: unknown): T => ({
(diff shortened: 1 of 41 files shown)


src/fhir/models/diagnostic-report.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const table: TestType[] = [
+      {
+        expectedValue: undefined,
+        obs: undefined,
+      },
+      {
+        expectedValue: "04/01/2022",
+        obs: [
+          set("effectiveDateTime", syntheticObservation(), "2022-04-02"),
```
bryanjknight, line 49: This feels very fragile given I'm hardcoding the property field, but I also don't want to over-engineer a builder pattern to create synthetic observations quickly

luthfib: Probably fine for this test. We can refactor later if it becomes a pain

luthfib (approved):
LGTM. Only nitpick would be helpers usually go at bottom/other file so model can be quickly be inspected.

luthfib (approved):

PR #701, opened 2023-05-04
CTW-1006 Adjusting metrics

eric-gunther:
- Separating metrics for timing, record counts and "zero moments"
- Using histogram metric type for record counts

### DIFF
--- a/src/fhir/conditions.ts
+++ b/src/fhir/conditions.ts
@@ -79,7 +79,7 @@ export function usePatientBuilderConditions() {
           }
         );
         const results = filterAndSort(setupConditionModels(conditions, bundle));
-        Telemetry.countMetric("req.builder_conditions", results.length);
+        Telemetry.histogramMetric("req.count.builder_conditions", results.length);
         return results;
       } catch (e) {
         throw Telemetry.logError(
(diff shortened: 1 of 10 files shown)


src/fhir/conditions.ts
```diff
@@ -79,15 +79,15 @@ export function usePatientBuilderConditions() {
           }
         );
         const results = filterAndSort(setupConditionModels(conditions, bundle));
-        Telemetry.countMetric("req.builder_conditions", results.length);
+        Telemetry.histogramMetric("req.count.builder_conditions", results.length);
```
eric-gunther, line 82: We don't care how often there are 0 builder conditions available so not tracking `req.count.builder_conditions.none`.

src/fhir/conditions.ts
```diff
@@ -106,15 +106,18 @@ function usePatientSummaryConditions() {
           }
         );
         const results = filterAndSort(setupConditionModels(conditions, bundle));
-        Telemetry.countMetric("req.summary_conditions", conditions.length);
+        if (results.length === 0) {
+          Telemetry.countMetric("req.count.summary_conditions.none");
+        }
+        Telemetry.histogramMetric("req.count.summary_conditions", results.length);
```
eric-gunther, line 112: Summary conditions we actually care about tracking b/c this will indicate how often we actually have novel, outside data, to surface.

mrosata (approved):

PR #704, opened 2023-05-05
CTW-1069 Fixing queries that need to filter by users builder ID

eric-gunther:
We removed the logic that filters by the users builder ID because we believed that the firstparty search parameter would take care of that for us. But that search parameter doesn't filter out all other builders data, it just filters out third party data. So because of CPR our query still returns data from other builders.

### DIFF
--- a/.changeset/giant-waves-destroy.md
+++ b/.changeset/giant-waves-destroy.md
@@ -1,5 +1,5 @@
 ---
-"@zus-health/ctw-component-library": minor
+"@zus-health/ctw-component-library": patch
 ---
 
 Fixes issue with missing builder filter.
--- a/.changeset/giant-waves-destroy.md
+++ b/.changeset/giant-waves-destroy.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Fixes issue with missing builder filter.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -65,15 +65,15 @@ const demoComponents: DemoComponent[] = [
     render: () => (
       <ZusAggregatedProfile
         includePatientDemographicsForm={false}
-        resources={["timeline"]}
+        resources={["medications", "medications-outside"]}
         title="ZAP"
       />
     ),
   },
-  { render: () => <PatientSearch />, title: "Patient Search" },
-  { render: () => <PatientHistoryTable />, title: "Patient History Table" },
-  { render: () => <PatientConditionsOutside />, title: "Patient Conditions" },
-  { render: () => <PatientDocuments />, title: "Patient Documents" },
+  // { render: () => <PatientSearch />, title: "Patient Search" },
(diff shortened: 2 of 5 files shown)


src/fhir/patient-helper.ts
```diff
@@ -19,7 +19,6 @@ export async function getBuilderFhirPatient(
     const response = await searchBuilderRecords("Patient", requestContext, {
       ...searchParams,
       identifier: `${systemURL}|${patientID}`,
-      _count: 1,
```
eric-gunther, line 22: Because `searchBuilderRecords` is not passing the tag parameter with the user's builder then this is sometimes returning a single patient instance that actually belongs to a different builder. Removing this count has us pull all the patient records in the query and then they get filtered out by the call to `filterSearchReturnByBuilderId` in `searchBuilderRecords`.

bryanjknight (dismissed):

src/fhir/search-helpers.ts
```diff
@@ -114,20 +117,34 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
...
+    records,
+    requestContext.contextBuilderId || requestContext.builderId
+  );
+
+  records.resources = resources;
+  records.bundle.entry = entry;
+  records.bundle.total = entry.length;
+  records.total = resources.length;
+
+  return records;
```
mrosata, line 147: Not for today, but we should wrap this logic in a helper that creates Bundles _(I've seen this pattern elsewhere in the code)_
```ts
  return new Bundle(resources, entry);
```

mrosata (dismissed):

.changeset/giant-waves-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
```
mrosata, line 2: This would be patch correct? It's fixing a bug
```suggestion
"@zus-health/ctw-component-library": patch
```

src/App.tsx
```diff
@@ -65,15 +65,15 @@ const demoComponents: DemoComponent[] = [
...
     ),
   },
-  { render: () => <PatientSearch />, title: "Patient Search" },
-  { render: () => <PatientHistoryTable />, title: "Patient History Table" },
-  { render: () => <PatientConditionsOutside />, title: "Patient Conditions" },
-  { render: () => <PatientDocuments />, title: "Patient Documents" },
+  // { render: () => <PatientSearch />, title: "Patient Search" },
+  // { render: () => <PatientHistoryTable />, title: "Patient History Table" },
+  // { render: () => <PatientConditionsOutside />, title: "Patient Conditions" },
+  // { render: () => <PatientDocuments />, title: "Patient Documents" },
```
pmadria3, line 76: Nit: do you just want to remove them instead?

pmadria3 (approved):
LGTM

mrosata (approved):

PR #707, opened 2023-05-05
CTW-785/FHIR Client proxy for writes

mrosata:
CTW-785

Adding a FHIR Client proxy for writes (creation writes). The CTWContext now exposes two clients, `fhirClient` and `fhirWriteBackClient`.

| `fhirClient` | `fhirWriteBackClient` |
|--|--|
| requests directly to Zus ODS API | requests indirectly to Zus ODS API via a proxy in our internal ehr-data-integration service |
| No longer be used to handle `.create` _(though it is still able)_ |  Is used when creating a new resource via `createOrUpdateFhirResource` function |
| Handles all reads | Should not be used for reads; The write-back proxy currently returns 409 on GETs |
| Handles all updates | Should not be used for updates; The write-back proxy can handle updates but we have no foreseeable need for this |
| Success responses are Smile CDR based | Success response are Smile CDR based |
| Error responses are Smile CDR based | Error responses could possibly be created by the ehr-data-integration rather than ODS if a problem happens before the service writes to ODS (ie: an EHR has write-back configuration but writing to that EHR fails and so ehr-data-service should return a JSON:API error) |


mrosata:
> @mrosata What's the status of this? Is this still an approach we want to take?

Yeah, this is the approach we're going to take. We had gone back and forth on another idea that used ZusHooks instead of a different write endpoint, but have since moved back to this strategy. We anticipate this fhir write endpoint will be utilized for multiple ehrs and resources.

src/fhir/client.ts
```diff
@@ -1,21 +1,64 @@
...
+            const method = requestOptions?.method?.toUpperCase() ?? "";
+            if (method === "POST") {
+              return fhirWriteClient.request(url, requestOptions);
+            }
+            return fhirClient.request(url, requestOptions);
+          };
+        }
+
+        if (typeof targetValue === "function") {
+          const isWrite = prop === "save" || prop === "create" || prop === "transaction";
```
Stephen-Tsen, line 51: `batch` should also be included?

nit: Create a list/set of the supported proxy write functions as opposed to multiple conditions

src/fhir/client.ts
```diff
@@ -1,21 +1,64 @@
...
     customHeaders,
   });
+
+  // Use Proxy to trap fhirClient writes that would create new resources.
+  // Learn more about proxies here: https://javascript.info/proxy
+  return new Proxy(fhirClient, {
+    get(target, prop) {
+      if (prop in target) {
+        const targetValue = target[prop as keyof typeof target];
+        if (prop === "request") {
```
Stephen-Tsen, line 33: This would intercept post search requests if I'm understanding this correctly.

Stephen-Tsen:
Does this mean standalone also goes through the proxy?

mrosata:
> Does this mean standalone also goes through the proxy?

That is correct

src/fhir/action-helper.ts
```diff
@@ -12,14 +12,15 @@ export async function createOrEditFhirResource(
...
-  const { fhirClient } = requestContext;
+  const { fhirClient, fhirWriteBackClient } = requestContext;
   const resourceModified = resource;
   let action = "create";
   let response;
 
   try {
     if (resource.id) {
       action = "update";
+      // Use the ODS client to update a resource
```
Stephen-Tsen, line 23: Is the plan to also have the EHR integration service also support updates?

Stephen-Tsen (approved):

src/fhir/action-helper.ts
mrosata: @Stephen-Tsen Currently the ehr-data-integration service supports updates. Based on an out-of-band discussion you and I had Friday, the ctw-component-library should not need to do any updates through the proxy because we don't currently have any use-case to "write-back" after an update _(and possibly never will)_.

PR #708, opened 2023-05-08
CTW-1006 Add optional EHR to metrics

eric-gunther:
Adds an `ehr` prop to `CTWProvider` which we forward along as a tag in any metrics.

Also fixed an issue where console errors appear if metrics are firing off before an environment has been set.

### DIFF
--- a/src/components/core/providers/ctw-provider.tsx
+++ b/src/components/core/providers/ctw-provider.tsx
@@ -42,6 +42,7 @@ type CTWProviderProps = {
   featureFlags?: FeatureFlags;
   theme?: Theme;
   locals?: Locals;
+  ehr?: string;
 } & (AuthTokenSpecified | AuthTokenURLSpecified);
 
 declare global {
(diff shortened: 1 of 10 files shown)


src/components/core/providers/ctw-provider.tsx
```diff
@@ -146,12 +147,12 @@ function CTWProvider({
   }, [token, ctwState]);
 
   useEffect(() => {
-    Telemetry.init(ctwState.env, enableTelemetry);
+    Telemetry.init(ctwState.env, ctwState.ehr, enableTelemetry);
```
luthfib, line 150: Is this second argument to allow us to discern whethern from who the telemetry is coming from?

luthfib (approved):
Left a nitpick, but LGTM

src/utils/telemetry.ts
```diff
@@ -85,9 +87,14 @@ export class Telemetry {
     }
   }
 
-  static init(environment: string, allowDataDogLogging = false) {
+  static init(
+    environment: string,
+    ehr: string | undefined = undefined,
```
mrosata, line 92: In the ehr-data-integration service we are using "unknown" in place of undefined. It could be good to do the same thing here.
```suggestion
    ehr = "unknown",
```

luthfib (approved):

mrosata (approved):

luthfib (approved):

PR #725, opened 2023-05-10
Ctw1065/admin user UI

luthfib:
Currently Can:
- Add User

Still Need: 
- Edit User
- Delete User
- Dynamically get Role ID's in case they change, we are not referrring to old role id's
- Refactor code a bit

image


WesleyKapow:
@luthfib please add comment as to why this was closed (good for visibility)

PR #727, opened 2023-05-12
fix: use FQS for Binary look ups

bryanjknight:
The "kludge removal" initiative is to store our data in manner consistent with our core beliefs without duplicating data. Part of this transition requires us to use FQS (Fast Query Service) to fetch binaries instead of ODS. Over the course of the next month, we will transition ODS queries to FQS


src/fhir/binaries.ts
```diff
@@ -27,12 +27,17 @@ async function getBinaryDocumentReq(
   requestContext: CTWRequestContext,
   binaryId: string
 ): Promise<fhir4.Binary> {
-  return queryClient.fetchQuery([QUERY_KEY_BINARY, binaryId], async () =>
-    requestContext.fhirClient.read({
+  return queryClient.fetchQuery([QUERY_KEY_BINARY, binaryId], async () => {
+    const resp = await requestContext.fqsRestClient(`Binary/${binaryId}`, {
+      method: "GET",
+    });
+    const retval: fhir4.Binary = {
```
luthfib, line 34: `retval` is an odd name for a variable. What are we getting back from this request? Is it a binary response?

luthfib (approved):
`src/components/content/CCDA/rawFieldToBinaryField.tsx` could be reanamed to `src/components/content/CCDA/raw-field-to-binary-field.tsx` to be consistent with our file naming

LGMT, some nitpick comments, feel free to do with it what you feel best.

src/fhir/client.ts
```diff
@@ -18,6 +18,25 @@ export function getFhirClient(env: Env, accessToken: string, builderId?: string)
   });
 }
 
+export function getFqsRestClient(env: Env, accessToken: string, builderId?: string) {
+  const baseUrl =
+    env === "production" ? `https://api.zusapi.com/fqs` : `https://api.${env}.zusapi.com/fqs`;
```
pmadria3, line 23: Building on Luthfi's comment, we probably want to store this globally so people (aka me 😁 ) can utilize this for fqs

src/components/content/CCDA/rawFieldToBinaryField.tsx
WesleyKapow, line 1: Files that provide components should be named after the component, so this should be `binary-field.tsx` (note the hyphen usage in file names).

Ah re hyphen naming convention -- most of the CCDA files were copied from FHIRplace and did not get this convention fix.

src/fhir/client.ts
```diff
@@ -18,6 +18,25 @@ export function getFhirClient(env: Env, accessToken: string, builderId?: string)
   });
 }
 
+export function getFqsRestClient(env: Env, accessToken: string, builderId?: string) {
```
WesleyKapow, line 21: As this isn't FHIR, we have a `services/fqs/client.ts` that this can get moved to.

WesleyKapow: nit: Also "Client" isn't very accurate or helpful as this is really a wrapper around `fetch` for fqs rest. Maybe `fqsRestFetch` or just `fqsFetch` or `fetchFromFqs`.

WesleyKapow: Wrapper looks great though!

src/fhir/binaries.ts
```diff
@@ -27,12 +27,17 @@ async function getBinaryDocumentReq(
   requestContext: CTWRequestContext,
   binaryId: string
 ): Promise<fhir4.Binary> {
-  return queryClient.fetchQuery([QUERY_KEY_BINARY, binaryId], async () =>
-    requestContext.fhirClient.read({
+  return queryClient.fetchQuery([QUERY_KEY_BINARY, binaryId], async () => {
+    const resp = await requestContext.fqsRestClient(`Binary/${binaryId}`, {
```
WesleyKapow, line 31: nit: favor not creating abbreviations, so this would be `response`.

Pandora's box opens...

I know this varies by language/team/repo/etc. We've mostly avoided these and as I've mentored a friend I came to realize that abbreviations have a way of creating new things for my brain to keep track of. It adds a very slight but noticeable amount of friction. What is `resp`, oh right that's `response`... all these little frictions add to cognitive load as we read and try to understand the code </2cents>

src/fhir/binaries.ts
WesleyKapow: Looks like we don't need this variable, can just return the object ya?

src/components/content/CCDA/modal-ccda.tsx
```diff
@@ -1,5 +1,5 @@
 import { XIcon } from "@heroicons/react/outline";
-import { Base64BinaryField } from "./base64toBinaryField";
```
WesleyKapow, line 2: Is this file no longer used? Kill it! Ah looks like this is really a file/component rename. Hopefully killing the original file will get git to detect it as such.

src/fhir/binaries.ts
bryanjknight: yeah, TS was giving me issues with the variable but worked when I created a variable. I'll try again with just returning the value

`retval` is classic C++ for return value :). Been using it since 2001

src/fhir/client.ts
bryanjknight: Going to move it to a separate non-fhir package (probably`fqs`) so feel free to move it around as you see fit!

src/fhir/binaries.ts
WesleyKapow: I'm more familiar with `retval`'s shorter cousin, `ret`

src/components/content/CCDA/modal-ccda.tsx
WesleyKapow: @bryanjknight this file was not removed :P (`base64binaryfield.tsx`)

PR #737, opened 2023-05-17
add way to automate updating packages

luthfib:
Add way to update packages via npm. This should be valid based on documentation but not 100% this will work as I didn't see a good way to test this. This shouldn't require a changeset.

We will: 
- Update weekly
- Only update minor and patches not major versions
- Have max of 3 pull requests from dependabot to limit amount of noise

This will NOT affect security updates


bryanjknight (approved):
LTGM, curious if aikido will replace this in the future

luthfib:
> LTGM, curious if aikido will replace this in the future

Perhaps. I thought it was more for vulnerability scanning. But not super familiar with all aikido can do

PR #748, opened 2023-05-18
Fix dismissing bug

WesleyKapow:
This PR fixes numerous issues with dismiss/restore functionality.

1. Don't filter out basic resources when fetching summary or lens resources. Post-kludge we want to filter these requests down to where the builder owner = the lens builder EXCEPT for resources there were included via `include` or `revinclude`. For those we want the ones owned by the builder.
2. Dismiss/restore loading spinner now stays until the new data has been fetched. This fixes a bug where you click "dismiss", and we show a loading spinner, and then when the spinner goes away, you briefly see "dismiss" again, before the new data arrives and we show "restore".
3. Support the edge-case where we have multiple basic resources for archive/unarchive. We'll now grab the most recent one.

### DIFF
--- a/src/components/content/hooks/use-toggle-archive.ts
+++ b/src/components/content/hooks/use-toggle-archive.ts
@@ -32,16 +32,11 @@ export function useToggleArchive<T extends fhir4.Resource>(
 
   const handleToggleArchive = useCallback(async () => {
     setIsLoading(true);
-
-    toggleArchive(model, await getRequestContext()).then(
-      async () => {
-        setIsLoading(false);
-        await queryClient.invalidateQueries([queryToInvalidate]);
-      },
-      () => {
-        setIsLoading(false);
-      }
-    );
+    await toggleArchive(model, await getRequestContext())
+    await queryClient.invalidateQueries([queryToInvalidate]);
+    // Timeout here fixes bug where we would briefly flash
+    // the old dismiss/restore text.
+    setTimeout(() => setIsLoading(false), 0);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [model]);
 
--- a/src/components/content/hooks/use-toggle-archive.ts
+++ b/src/components/content/hooks/use-toggle-archive.ts
@@ -32,7 +32,7 @@ export function useToggleArchive<T extends fhir4.Resource>(
 
   const handleToggleArchive = useCallback(async () => {
     setIsLoading(true);
-    await toggleArchive(model, await getRequestContext())
+    await toggleArchive(model, await getRequestContext());
     await queryClient.invalidateQueries([queryToInvalidate]);
     // Timeout here fixes bug where we would briefly flash
     // the old dismiss/restore text.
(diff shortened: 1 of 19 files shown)


mrosata (dismissed):

src/fhir/models/fhir-model.ts
```diff
@@ -60,7 +60,7 @@ export abstract class FHIRModel<T extends fhir4.Resource> {
   // In practice, there should only be a single Basic resource, but
   // we need to handle the case where there may be several.
   getLatestBasicResourceByActions(profileActions: string[]): Basic | undefined {
-    const ordered = orderBy(this.revIncludes, "meta.lastUpdated", "desc");
+    const ordered = orderBy(this.revIncludes, "meta.lastUpdated", "desc") as Resource[];
```
mrosata, line 63: If we changed the property 
```ts
  readonly revIncludes?: Resource[];
```
to
```ts
  readonly revIncludes: Resource[] = [];
```
or modified the constructor signature from
```ts
  constructor(resource: T, includedResources?: ResourceMap, revIncludes?: Resource[]) {
```
to  something like
```ts
  constructor(resource: T, includedResources?: ResourceMap, revIncludes: Resource[] = []) {
```
Would we not need to hold typescripts hand here with `as Resource[]`? Is the complaint typescript gives here just that it can't infer the return type from this call to `orderBy` because it doesn't know if `this.revIncludes` is an array of resources or undefined?

src/fhir/search-helpers.ts
```diff
@@ -209,10 +206,7 @@ export async function searchSummaryRecords<T extends ResourceTypeString>(
   This will help avoid getting duplicate results or know there is no data for the builder. 
   Once we have been in the post-kludge world long enough we can remove this functionality. */
   if (resources.length === 0 || entry.length === 0) {
-    ({ entry, resources } = filterSearchReturnByBuilderId(
-      records,
-      builderId,
-    ));
+    ({ entry, resources } = filterSearchReturnByBuilderId(records, builderId));
```
mrosata, line 209: TIL, I didn't know reassignments could be done this way

mrosata (approved):

src/components/content/hooks/use-toggle-archive.ts
```diff
@@ -32,16 +32,11 @@ export function useToggleArchive<T extends fhir4.Resource>(
...
-      },
-      () => {
-        setIsLoading(false);
-      }
-    );
+    await toggleArchive(model, await getRequestContext());
+    await queryClient.invalidateQueries([queryToInvalidate]);
+    // Timeout here fixes bug where we would briefly flash
+    // the old dismiss/restore text.
+    setTimeout(() => setIsLoading(false), 0);
```
Stephen-Tsen, line 39: I believe this will get a react warning for `Can't perform a React state update on an unmounted component.`.

WesleyKapow: There may be a race condition for getting the warning. I can reliably get it for the first dismiss when I bump the timeout to `200` but I have not seen it when it is at `0`.

src/fhir/models/fhir-model.ts
WesleyKapow: Sadly no :'( -- the problem has something to do with `orderBy` and a complex set of things it says the return type can be.

PR #749, opened 2023-05-18
CTW-691: Rewire Conditions table to use FQS

patrickvalle:
https://zeushealth.atlassian.net/browse/CTW-691


src/App.tsx
```diff
@@ -45,11 +45,11 @@ const theme = {
...
   en: {
     glossary: {
-      condition_one: "problem",
-      condition_other: "problems",
+      condition_one: "conditions",
+      condition_other: "conditions",
     },
     main: {
-      "zap.tabs.conditionsOutside": "problems",
+      "zap.tabs.conditionsOutside": "Outside Conditions",
```
Stephen-Tsen, line 52: `conditions`?

src/fhir/conditions.ts
```diff
@@ -71,16 +91,37 @@ export function usePatientBuilderConditions() {
...
-          "Condition",
-          requestContext,
-          {
-            patientUPID: patient.UPID,
-          }
-        );
-        const results = filterAndSort(setupConditionModels(conditions, bundle));
-        Telemetry.histogramMetric("req.count.builder_conditions", results.length);
-        return results;
+        const graphClient = createGraphClient(requestContext);
```
Stephen-Tsen, line 94: Can we add this to `getRequestContext` like we have for `fhirClient`?

src/services/fqs/conditions.ts
```diff
@@ -125,20 +157,26 @@ export function usePatientConditionsOutside() {
   const [conditions, setConditions] = useState<ConditionModel[]>([]);
   const patientConditionsQuery = usePatientBuilderConditions();
   const otherConditionsQuery = usePatientSummaryConditions();
+  const basicQuery = useBasic();
```
patrickvalle, line 160: @WesleyKapow I know we talked about having the `Basic` logic get married to the `Condition` logic at the component level vs in this `fqs` namespace, but this just felt like the easiest/better place to put the logic given the current structure of things. Curious on your thoughts here, if this is totally revolting or not.

src/fhir/models/condition.ts
```diff
@@ -178,7 +178,7 @@ export class ConditionModel extends FHIRModel<fhir4.Condition> {
     return formatStringToDate(this.resource.onsetString);
   }
 
-  get patient(): PatientModel | undefined {
+  get patientOrganizationName(): string | undefined {
```
patrickvalle, line 181: Reduced the scope of this getter to return only the field needed by the app- this makes tracking what graphql properties we're actually using in the view way easier

src/services/fqs/conditions.ts
```diff
@@ -0,0 +1,331 @@
...
+            return results;
+          } catch (e) {
+            throw Telemetry.logError(
+              e as Error,
+              `Failed fetching conditions for patient: ${patient.UPID}`
+            );
+          }
+        }, "req.timing.builder_conditions")
+  );
+}
```
luthfib: Couple of things here...

1. I think this is a big large codewise to use ternary and it makes it hard to read
2. We are repeating a lot of logic between FQS and ODS. I think what would make more sense it to wrap the fetching for each in a function and handle the error part and histogram part right after the fetch. Soo might look kinda like...

```
withTimerMetric(async (requestContext, patient) => {
  try {
      const results = enableFQS ? fetchFQSConditions : fetchODSConditions
      Telemetry.histogramMetric("req.count.builder_conditions", results.length);
      return results;
  } catch (e) {
            throw Telemetry.logError(
              e as Error,
              `Failed fetching conditions for patient: ${patient.UPID}`
            );
          }
        }, "req.timing.builder_conditions")
}
```

luthfib (changes requested):
Left some comments on potential changes

src/services/conditions.ts
```diff
@@ -0,0 +1,331 @@
...
+            Telemetry.histogramMetric("req.count.summary_conditions", results.length);
+            return results;
+          } catch (e) {
+            throw Telemetry.logError(
+              e as Error,
+              `Failed fetching conditions outside for patient: ${patient.UPID}`
+            );
+          }
+        }, "req.timing.summary_conditions")
+  );
```
mrosata, line 203: I wouldn't do this like
```ts
enableFQS ?
  withTimerMetric(/**/, "req.timing.summary_conditions.fqs") :
  withTimerMetric(/**/, "req.timing.summary_conditions") :
```
I'd do it all in one `withTimerMetric`. You could still use different metric names.. though I probably wouldn't use different metric names either. Could tag the metric with `fqs` tho to be able to filter it out

mrosata: Could write it like
```ts
withTimerMetric(async (requestContext, patient) => {
  if (enableFQS) {
    /* work work */
  } else {
    /* work work */
  }
}, "req.timing.summary_conditions", enableFQS ? ["fqs"] : [])
```

src/fhir/basic.ts
```diff
@@ -54,6 +57,36 @@ export async function recordProfileAction<T extends fhir4.Resource>(
   }
 }
 
+export function useBasic(enableFQS: boolean) {
+  return useQueryWithPatient(
+    QUERY_KEY_BASIC,
+    [],
+    enableFQS
+      ? // FQS implementation
```
WesleyKapow, line 65: nit: Comments like this = a sign you should pull things out to helper functions (I like to put non-exported helper functions below all of the exported functions to make it easier to see in the file the API it exposes [the exports]).

src/fhir/basic.ts
```diff
@@ -54,6 +57,36 @@ export async function recordProfileAction<T extends fhir4.Resource>(
...
+            }
+            Telemetry.histogramMetric("req.count.basic", resources.length);
+            return resources;
+          } catch (e) {
+            throw new Error(
+              `Failed fetching basic information for builder ${requestContext.builderId}`
+            );
+          }
+        }, "req.timing.basic")
+      : // ODS implementation (noop)
```
WesleyKapow, line 82: What's not obvious here is why do we do a no-op here in this case.

src/fhir/models/condition.ts
```diff
@@ -19,8 +20,7 @@ import {
   ClinicalStatus,
   CONDITION_CODE_PREFERENCE_ORDER,
   VerificationStatus,
-} from "@/fhir/conditions";
-import { findReference } from "@/fhir/resource-helper";
+} from "@/services/conditions";
```
WesleyKapow, line 23: Things like `ClinicalStatus` and `VerificationStatus` seem very much like FHIR things and not some service thing.

src/services/conditions.ts
```diff
@@ -0,0 +1,331 @@
...
+
+export type ClinicalStatus =
+  | "active"
+  | "recurrence"
+  | "relapse"
+  | "inactive"
+  | "remission"
+  | "resolved";
+
+export const CONDITION_CODE_PREFERENCE_ORDER: CodePreference[] = [
```
WesleyKapow, line 52: Check usage of this, maybe this belongs in the model?

src/services/fqs/conditions.ts
```diff
@@ -0,0 +1,331 @@
...
+];
+
+export interface ConditionConnection {
+  pageInfo: GraphqlPageInfo;
+  edges: GraphqlConnectionNode<Condition>[];
+}
+
+export interface ConditionGraphqlResponse {
+  ConditionConnection: ConditionConnection;
+}
```
WesleyKapow: nit: Maybe belongs in `services/fqs/queries/condition.ts` or some other file under `services/fqs`.

The thinking here is, this is specific to FQS.

src/services/fqs/conditions.ts
```diff
@@ -0,0 +1,331 @@
...
+  const newCondition: fhir4.Condition = {
+    resourceType: "Condition",
+    subject: {
+      type: "Patient",
+      reference: `Patient/${patientId}`,
+    },
+    ...getClincalAndVerificationStatus("Active"),
+  };
+  return getAddConditionWithDefaults(newCondition);
+}
```
WesleyKapow: nit: Looks like a FHIR thing, so probably `fhir/conditionsts`

src/fhir/basic.ts
```diff
@@ -54,6 +57,36 @@ export async function recordProfileAction<T extends fhir4.Resource>(
...
+        async () =>
+          new Promise<Basic[]>((resolve) => {
+            resolve([]);
+          })
+  );
+}
```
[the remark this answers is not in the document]
patrickvalle: When we're hitting ODS, the `Basic` resources get `_revincluded` in that request and filtered out already in `usePatientSummaryConditions`. We only need to independently fetch `Basic` like this when we're hitting FQS.

src/services/conditions.ts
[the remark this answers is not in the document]
patrickvalle: IMO a "service" has business logic pertaining to a specific part of the domain and is responsible for orchestrating API calls- so the concept of a "conditions service" exists regardless of the backend(s) it's using. That way when your backend needs evolve, you're not constantly refactoring where your code lives. 

To me something like a "helper" feels super generalized and is, at best, a place to park functions that get called from multiple parts of the app (components, services, etc)... and at worst, anemic code you can't figure out a better home for.

src/fhir/basic.ts
```diff
@@ -54,6 +57,20 @@ export async function recordProfileAction<T extends fhir4.Resource>(
   }
 }
 
+export function useBasic(enableFQS: boolean) {
```
patrickvalle, line 60: @mrosata @luthfib @WesleyKapow I moved the fetching into a helper function and added a comment attempting to explain the noop. LMK what you think

src/services/conditions.ts
```diff
@@ -0,0 +1,341 @@
...
+    subject: {
+      type: "Patient",
+      reference: `Patient/${patientId}`,
+    },
+    ...getClincalAndVerificationStatus("Active"),
+  };
+  return getAddConditionWithDefaults(newCondition);
+}
+
+export function usePatientBuilderConditions(enableFQS: boolean) {
```
patrickvalle, line 84: Moved the fetch functionality into separate functions for `usePatientBuilderConditions` and `usePatientSummaryConditions`

src/components/content/conditions/patient-conditions-outside.tsx
```diff
@@ -69,7 +69,8 @@ const RowActions = ({ record }: RowActionsProps<ConditionModel>) => {
   const showAddConditionForm = useAddConditionForm();
   const { isLoading, toggleArchive } = useToggleArchive(
     record,
-    QUERY_KEY_OTHER_PROVIDER_CONDITIONS
+    QUERY_KEY_OTHER_PROVIDER_CONDITIONS,
+    QUERY_KEY_BASIC
```
WesleyKapow, line 73: FYI when we are just using FQS, we won't need to invalidate `QUERY_KEY_OTHER_PROVIDER_CONDITIONS` anymore.

src/components/content/hooks/use-toggle-archive.ts
```diff
@@ -33,7 +33,10 @@ export function useToggleArchive<T extends fhir4.Resource>(
   const handleToggleArchive = useCallback(async () => {
     setIsLoading(true);
     await toggleArchive(model, await getRequestContext());
-    await queryClient.invalidateQueries([queryToInvalidate]);
+    await queriesToInvalidate.forEach((queryToInvalidate) =>
+      queryClient.invalidateQueries([queryToInvalidate])
+    );
```
WesleyKapow, line 38: 🚫 **BLOCKER**

`forEach` doesn't return a promise. Took some digging but I found an eslint rule we should have, please add this to `eslintrc.cjs` and verify this now shows you an error `"@typescript-eslint/await-thenable": "error",`

You should instead `await Promise.all()` and you can use `queriesToInvalidate.map` to get an array of promises from `invalidateQueries`.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
```diff
@@ -76,7 +76,7 @@ export const zusAggregatedProfileTabs: ZusAggregatedProfileTabs = {
     display: () => (
       <div className="ctw-space-x-2">
         <span className="ctw-capitalize">{i18next.t("zap.tabs.conditionsOutside")}</span>
-        <PatientConditionsOutsideBadge />
+        <PatientConditionsOutsideBadge {...{ enableFQS: props.enableFQS }} />
```
WesleyKapow, line 79:
```suggestion
        <PatientConditionsOutsideBadge enableFQS={props.enableFQS} />
```

WesleyKapow (changes requested):

src/components/content/hooks/use-toggle-archive.ts
WesleyKapow: Oh or maybe simply `await queryClient.invalidateQueries(queriesToInvalidate)`

patrickvalle: @WesleyKapow the eslint rule and `Promise.all` are pushed

src/components/content/conditions/patient-conditions-outside.tsx
patrickvalle: This is a good catch, but is there a non-heinous way to wire through `enableFQS` to the `RowActions`? If not, can I advocate for us to keep the code as-is since supporting both FQS and ODS should be a short-lived state?

WesleyKapow: I wasn't proposing a change, just noting that when we do switch to just FQS, this can get changed then.

WesleyKapow (approved):
LGTM, great work on this btw :D

PR #755, opened 2023-05-22
Fix storybook

WesleyKapow:
Refactor mocks to use constants for builder and patient ids. Refactor stories to use same constants when setting up `CTWProvider` and `PatientProvider`. This should get all of our components and mocks aligned so that we get back real data. This use to be less of a concern when we did builder filtering in ODS. Now that we do it client side, we need our IDs to really match up!

ALSO dropped timeline V1 and dropped `V2` from v2 name.


pmadria3 (approved):
LGTM

patrickvalle (approved):

pmadria3 (approved):

PR #758, opened 2023-05-23
Extend exports & add dynamic form select options

Stephen-Tsen:
Export more helpers to help facilitate building tables and queries in Standalone. Also update forms to allow dynamic dropdowns.


luthfib (dismissed):

WesleyKapow:
Update title and PR description as this does more than just add some exports.

src/components/core/providers/authentication/authentication-provider.tsx
```diff
@@ -7,7 +7,7 @@ import { ctwFetch, queryClient } from "@/utils/request";
 // We use an expiry padding to provide a buffer to prevent race conditions.
 // A race condition could happen in that we check if the token is expired,
 // it isn't, but by time we do our request(s) it has expired.
-const EXPIRY_PADDING_MS = 60000;
+const EXPIRY_PADDING_MS = 60_000;
```
bryanjknight, line 10: I forgot about this trick!

bryanjknight (approved):

src/components/core/providers/authentication/authentication-provider.tsx
```diff
@@ -61,10 +61,13 @@ async function checkOrRefreshAuth(
     const response = await queryClient.fetchQuery([QUERY_KEY_AUTH_TOKEN, url], async () =>
       ctwFetch(url as string, { headers })
     );
-    const newToken = await response.json();
+
+    // For some reason we can end up calling json() twice on the same response.
+    // We avoid this double read error by cloning the response.
+    const newToken = await response.clone().json();
```
mrosata, line 67: The reason this happens is the response is cached by tanstack so subsequent calls to `checkOrRefreshAuth` may or may not get that cached fetch response object.

Since this is the only place using this query cache key, it might make sense for us to just return the object to tanstack to cache directly in the `fetchQuery` function instead of caching the fetch response.
```ts
    const response = await queryClient.fetchQuery([QUERY_KEY_AUTH_TOKEN, url], async () => {
      const response = ctwFetch(url as string, { headers });
      const newToken = await response.json();
      return {
        accessToken: newToken.access_token,
        expiresAt: claimsExp(newToken.access_token) * 1000,
        expiresAt: claimsExp(newToken.access_token) * 1_000,
      };
    });
```

Doesn't need to happen in this PR, but that is the reason it sometimes is called twice on same response

mrosata (approved):

pmadria3 (approved):

src/components/core/providers/authentication/authentication-provider.tsx
WesleyKapow: Perfection! It didn't occur to me that everywhere else we store a processed response. We should be careful not to repeat this mistake, we NEVER want to store the raw response in tanstack!

WesleyKapow (approved):
LGTM

PR #762, opened 2023-05-24
Fix date in patient history status

luthfib:
Fix issue with patient history api status box where date shows up in the past.


src/fhir/formatters.ts
```diff
@@ -104,7 +104,7 @@ export function formatAge(age: fhir4.Age): string {
 // Formats a string ISO date to MM/DD/YYY.
 export function formatISODateStringToDate(dateStr?: string | number): string | undefined {
   if (!dateStr) return undefined;
-  const date = typeof dateStr === "number" ? new Date(dateStr).toISOString() : dateStr;
+  const date = typeof dateStr === "number" ? new Date(dateStr * 1000).toISOString() : dateStr;
```
mrosata, line 107: Do we know this will always be a timestamp that is not properly using ms rather than shorter unix timestamp? This is ugly, but should we do something here like:
```suggestion
  const ensureJsTimestamp = (n: number) => `${n}`.length <= 11 ? n * 1000 : n;
  const date = typeof dateStr === "number" ? new Date(ensureJsTimestamp(dateStr)).toISOString() : dateStr;
```

mrosata (dismissed):

mrosata (approved):

src/api/patient-history.ts
```diff
@@ -62,7 +62,7 @@ export const schedulePatientHistory = async (
     return await response.json();
   } catch (e) {
     const err = e as Error;
-    const errorMessage = `Error scheduling patient history job with id of ${patientID}: ${err.message}`;
+    const errorMessage = `Error scheduling patient history job with patientID of ${patientID}: ${err.message}`;
```
Stephen-Tsen, line 65: Why not `patient ID` We should be mindful of how we present errors to users and the language should reflect the audience.

PR #767, opened 2023-05-25
CTW-692: part 1 - med statement

sarahmangels:


src/services/fqs/queries/fragments.ts
```diff
@@ -94,3 +94,24 @@ export const fragmentPatient = gql`
     }
   }
 `;
+
+export const fragmentReference = gql`
```
sarahmangels, line 98: see TODOs in here --- not sure how deep we need to go on References. They just keep going 😱

src/fhir/models/medication-statement.ts
```diff
@@ -27,6 +27,18 @@ export class MedicationStatementModel extends FHIRModel<fhir4.MedicationStatemen
 
   readonly builderPatientRxNormStatus?: Record<string, string>;
 
+  get aggregatedFrom(): Reference[] {
```
sarahmangels, line 30: just alphabetizing - no changes

src/fhir/medication.ts
```diff
@@ -69,7 +69,7 @@ export function getIdentifyingRxNormCoding(
   const rxNorm = codeableConcept?.coding?.find(
     (code) =>
       // must be an RxNorm code
-      code.system === SYSTEM_RXNORM && code.extension === undefined
+      code.system === SYSTEM_RXNORM && !code.extension?.some((e) => e.url === SYSTEM_ENRICHMENT)
```
sarahmangels, line 72: switched this because the FQS data has code.extension as `null` so the existing check didnt work

patrickvalle (approved):
lgtm

PR #769, opened 2023-05-26
Fix CCDA viewer when there's no ReceivedOrganization 

WesleyKapow:
Tested this by updating a binary used in mocks. It would crash before this change, and after, it works!

This is a port of https://github.com/zeus-health/fhir-place/pull/203

Information recipient, one with org, others without:
Screenshot 2023-05-30 at 12 37 10 PM

### DIFF
--- a/.changeset/early-waves-argue.md
+++ b/.changeset/early-waves-argue.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Added `fhirWriteBackClient` to proxy creation of new FHIR resources through the ehr-data-integration server.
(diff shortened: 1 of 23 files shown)


mattgayzus (approved):
This is pretty much a perfect port of [the change](https://github.com/zeus-health/fhir-place/pull/203) made in FHIRplace.

One thing to be aware of is that if there are multiple information recipients and some have organizations, but not others, the CCDA will render in a very ugly fashion (there's an extra column sticking out to the side).  When I made this change in FHIRplace, I wasn't super-concerned as FHIRplace users are mostly internal or engineering types at our builders.   In this case, you might worry, since these users are more likely to be providers with a higher standard for UI.

pmadria3 (approved):

WesleyKapow:
@mattgayzus FYI, I added a screen to PR description, I think showing the case you're talking about, but it looks good to me /shrug

PR #774, opened 2023-05-30
Ctw 1062/labs-fqs

pmadria3:
This PR wires the DiagnosticReports table with FQS.

Screenshot 2023-05-31 at 4 04 16 PM


src/fhir/timeline-event.ts
```diff
@@ -16,7 +16,7 @@ type TimelineEventModelParams = {
 export function useTimelineEvents() {
   const [timelineEvents, setTimelineEvents] = useState<TimelineEventModel[]>();
   const patientEncountersQuery = usePatientEncounters();
-  const diagnosticReportQuery = usePatientAllDiagnosticReports();
+  const diagnosticReportQuery = usePatientAllDiagnosticReports(true);
```
patrickvalle, line 19: This shouldn't always be FQS, right?

bryanjknight: It actually should be always be FQS because ODS has a bug that's not returning all observations

pmadria3: I added the toggle-able nature but set it to true

src/App.tsx
bryanjknight, line 1: Let's use the VITE_DEMO_COMPONENTS variable and stop changing the `App.tsx` file

src/components/content/timeline/patient-timeline.tsx
```diff
@@ -15,10 +15,11 @@ import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
 
 export type PatientTimelineProps = {
   className?: cx.Argument;
+  enableFQS?: boolean;
 };
 
-export function PatientTimeline({ className }: PatientTimelineProps) {
-  const timelineEventsQuery = useTimelineEvents();
+export function PatientTimeline({ className, enableFQS = false }: PatientTimelineProps) {
```
bryanjknight, line 21: is this going to be a problem that we're defaulting to ODS as opposed to making this required?

src/fhir/fqs.ts
```diff
@@ -0,0 +1,14 @@
+import { GraphQLClient } from "graphql-request";
+import { CTWRequestContext } from "@/components/core/providers/ctw-context";
+
+export const createGraphClient = (requestContext: CTWRequestContext) => {
```
bryanjknight, line 4: Shouldn't this be `services/fqs/client.ts`?

bryanjknight (changes requested):
A few small blockers but hopefully nothing too crazy

src/components/content/timeline/patient-timeline.tsx
pmadria3: It is going to be for a little bit because the problem is FQS is running into errors on their side and thus we are not getting the correct results back atm (tested with a couple people in prod but sandbox works fine) so we need to default to false for the time being.

bryanjknight: Right, my concern is that we are creating a default value as opposed to failing if it's not configured (which I think guarantees that we tie all the components together)

bryanjknight (dismissed):

bryanjknight (dismissed):

bryanjknight (approved):

PR #779, opened 2023-06-01
Allergies on FQS

pmadria3:
This PR is wiring Allergies onto FQS. 
Screenshot 2023-06-06 at 3 51 43 PM


src/fhir/allergies.ts
```diff
@@ -1,48 +1,67 @@
...
+              upid: patient.UPID,
+              cursor: "",
+              first: 1000,
+              sort: {
+                lastUpdated: "DESC",
+              },
+            })) as AllergyGraphqlResponse;
+
+            const nodes = data.AllergyIntoleranceConnection.edges.map((x) => x.node);
+            // TODO: No longer needed once https://zeushealth.atlassian.net/browse/DRT-249 is resolved.
```
sarahmangels, line 31: which line is this TODO for?

src/fhir/allergies.ts
```diff
@@ -1,48 +1,67 @@
...
+
+            const results = applyAllergyFiltersFQS(nodes);
+            if (results.length === 0) {
+              Telemetry.countMetric("req.count.allergies.none");
```
sarahmangels, line 35: do you want to add the `["fqs"]` tag thingy like this
https://github.com/zeus-health/ctw-component-library/blob/5717f890f9583e67c024b93794afe0bdb4835aaa/src/services/conditions.ts#L44

src/fhir/allergies.ts
```diff
@@ -1,48 +1,67 @@
...
+            }
+            Telemetry.histogramMetric("req.count.allergies", results.length);
```
sarahmangels, line 37: same here

src/fhir/allergies.ts
```diff
@@ -1,48 +1,67 @@
...
+            return results;
+          } catch (e) {
+            throw new Error(`Failed fetching allergies information for patient ${patient.UPID}`);
           }
-        );
+        }, "req.timing.allergies")
```
sarahmangels, line 42: and/or here?

src/fhir/models/allergies.ts
```diff
@@ -59,6 +59,11 @@ export class AllergyModel extends FHIRModel<fhir4.AllergyIntolerance> {
     return organizationDisplay?.managingOrganization?.display || organizationName;
   }
 
+  get managingOrganizationFQS(): string | undefined {
```
sarahmangels, line 62: is this used anywhere?

pmadria3: Yep the table (specified in the columns for the table) we need to specify since the object is mutated from FHIR and breaks the drawer since managingOrg isn't retrieved the same in FQS vs ODS

sarahmangels (approved):

src/fhir/models/allergies.ts
WesleyKapow, line 62: Instead of introducing a new method which pushes the complexity to consumers, we can encapsulate this complexity. Modify `managingOrganization` to "do the right thing". It can see if we have the resource provided from FQS. In fact, you might be able to modify the `findReference` helper to look on the passed `reference` object for these FQS `resource` keys.

src/components/content/allergies/story-helpers/mocks/allergy-intoleranceFQS.ts
```diff
@@ -0,0 +1,638 @@
+export const allergyIntoleranceFQS = {
```
WesleyKapow, line 1: nit: We don't use capitals in our file names.. this should be `allergy-intolerance-fqs.ts`

src/components/content/allergies/helpers/allergies-filter.tsx
```diff
@@ -18,4 +18,16 @@ export const applyAllergyFilters = (
   return allergyData;
 };
 
+export const applyAllergyFiltersFQS = (data: fhir4.AllergyIntolerance[]) => {
```
WesleyKapow, line 21: Why is this needed? Seems almost identical to the above. If it's due to not having a `includedResources` map, you can mark that as optional on the above and then use it.

.changeset/heavy-ways-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Update Allergies to utilize FQS.
```
WesleyKapow, line 5: Should indicate this is guarded by a prop. As in, this note as written reads like allergies now uses FQS which is not true.

> Allergies can now read from FQS with the new `enableFQS` prop.

PR #781, opened 2023-06-02
Fix bug updating resources fetched from FQS

WesleyKapow:
FQS returns some non FHIR compliant data (nulls and reference resources). We now filter those out before saving to ODS.


patrickvalle (approved):
lgtm

PR #791, opened 2023-06-05
CTW-1185/fqs read your writes

WesleyKapow:
CTW-1185

Setup long polling from FQS in our generic `createOrEditFhirResource` helper. We'll need to be very careful when introducing other helpers that mutate data in ODS. Hopefully we can keep using this existing helper which now await's FQS to have the updated data.

### DIFF
--- a/src/fhir/action-helper.ts
+++ b/src/fhir/action-helper.ts
@@ -2,7 +2,9 @@ import { Resource } from "fhir/r4";
 import { fixupFHIR } from "./client";
 import { isFhirError } from "./errors";
 import { createProvenance } from "./provenance";
+import { isFHIRResource } from "./types";
 import { CTWRequestContext } from "@/components/core/providers/ctw-context";
+import { longPollFQS } from "@/services/fqs/long-poll-fqs";
 import { Telemetry } from "@/utils/telemetry";
 
 export async function createOrEditFhirResource(
@@ -39,6 +41,20 @@ export async function createOrEditFhirResource(
       }
     }
     Telemetry.reportActionSuccess(`${resource.resourceType}.${action}`);
+
+    // Read-Your-Writes!
+    // Wait for FQS to have the latest version of the resource.
+    // This way callers can safely refetch/invalidateQueries
+    // and be sure to get fresh data from FQS.
+    if (isFHIRResource(response) && response.id && response.meta?.lastUpdated) {
+      await longPollFQS(
+        requestContext,
+        resource.resourceType,
+        response.id,
+        response.meta.lastUpdated
+      );
+    }
+
     return response;
   } catch (err) {
     Telemetry.reportActionFailure(`${resource.resourceType}.${action}`);
--- a/src/fhir/action-helper.ts
+++ b/src/fhir/action-helper.ts
@@ -1,11 +1,8 @@
-import { FhirResource, Resource } from "fhir/r4";
-import { v4 as uuidv4 } from "uuid";
+import { Resource } from "fhir/r4";
 import { fixupFHIR } from "./client";
 import { isFhirError } from "./errors";
-import { ASSEMBLER_CODING, CREATE_CODING, createProvenance } from "./provenance";
+import { createProvenance } from "./provenance";
 import { CTWRequestContext } from "@/components/core/providers/ctw-context";
-import { getUsersPractitionerReference } from "@/fhir/practitioner";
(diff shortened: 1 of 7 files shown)


src/fhir/action-helper.ts
```diff
@@ -35,114 +34,31 @@ export async function createOrEditFhirResource(
...
+
     return response;
   } catch (err) {
     Telemetry.reportActionFailure(`${resource.resourceType}.${action}`);
     Telemetry.logError(err as Error);
     return err;
   }
 }
-
-export async function createFhirResourceWithProvenance(
```
WesleyKapow, line 51: Instead of replacing this with the existing `createOrEditFhirResource`, I could have made this work for both update and edit.

I don't feel too strongly but the change I made was the simplest to avoid having to mess with updating transactions.

src/services/fqs/long-poll-fqs.ts
```diff
@@ -0,0 +1,52 @@
...
+    // Refetch the resource from FQS.
+    try {
+      const data = await graphClient.request(query);
+      currentTimestamp = Date.parse(String(get(data, "Condition.meta.lastUpdated")));
+    } catch (err) {
+      // Ignore errors and retry.
+      // This can happen if we just created a new resource that isn't in FQS yet.
+    }
+  } while (currentTimestamp < targetTimestamp && performance.now() - startTime < TIMEOUT_MS);
+}
```
luthfib, line 46: Might be worth doing a test around this. If this function goes wrong we could theoretically end up making lots of requests.

WesleyKapow: Ah goes wrong as in, gets modified and there's a problem? I originally had included `retryCount` in the `while` but decided it was superfluous.

I'll add a test to lock in the behavior.

luthfib: Yeah, just because this would touch every resource. So someone changes it and we are firing of potentially tons of requests

src/services/fqs/long-poll-fqs.ts
```diff
@@ -0,0 +1,52 @@
...
+import { get } from "@/utils/nodash";
+
+// We'll keep retrying until we've gone over our timeout.
+const TIMEOUT_MS = 20_000;
+
+// Due to mutations going to ODS, there is a delay
+// before FQS has the latest fresh resource.
+// This function will poll FQS until the resource.meta.lastUpdated timestamp
+// is greater or equal to the provided lastUpdated timestamp.
+export async function longPollFQS(
```
bryanjknight, line 13: It seems that after 20 seconds we stop polling. What is the user experience at that point?

src/fhir/types.ts
```diff
@@ -13,3 +13,12 @@ export type ResourceMap = { [key: string]: fhir4.Resource | undefined };
 export type ResourceArrayMap = Map<string, fhir4.Resource[]>;
 
 export type Tag = { system: string; code: string };
+
+export function isFHIRResource(resource: unknown): resource is fhir4.DomainResource {
+  return (
+    typeof resource === "object" &&
+    resource !== null &&
+    "resourceType" in resource &&
+    resource.resourceType !== "Binary"
```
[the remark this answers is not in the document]
WesleyKapow: Open to suggestions here, otherwise I'll add a comment.

The issue is `Binary extends Resource` while the others extend `DomainResource`. I should change this to be more accurate and say `isDomainResource` (which is what has `extensions` which is what I'm using).

src/services/fqs/long-poll-fqs.ts
WesleyKapow: After the timeout this function will just return and the consuming code will just continue. For users, this means the refetch from FQS that is suppose to get the latest/fresh data, it will likely get the stale data. So editing a condition = condition is edited in ODS but the UI still shows it as before :(.

src/fhir/types.ts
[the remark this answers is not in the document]
bryanjknight: gotcha, i would just leave a comment

src/services/fqs/long-poll-fqs.ts
bryanjknight: can we monitor this so that we can track if 20 seconds is sufficient?

luthfib (approved):

luthfib (approved):

PR #793, opened 2023-06-06
CTW-1186/onResourceSave optional callback

mrosata:
This feature allows an optional callback function `onResourceSave` that providers can pass in from on the provider context. Whenever a resource gets created or edited in the function `createOrEditFHIRResource` the callback is fired. The backstory is that some providers wish to be notified in the frontend after a save or edit is performed so they can perform work as well.

The callback lives on the requestContext, this is mainly because the request context is being passed into the `createOrEditFHIRResource` function already and it is already a bit unwieldy expecting all those callers to get this context and pass it in, so having to update all those callers to additionally get this callback would be a bit cringe.

One thought, we could rename this `onCreateOrEditFHIRResource` as that function of a similar name is the only caller of this callback at the moment, however, `onResourceSave` feels better, matches the `onPatientSave` callback we already make available on the patient provider.

https://www.loom.com/share/447e1b60daaf446dbb080daba7003485


src/components/core/providers/ctw-provider.tsx
```diff
@@ -17,6 +18,7 @@ type CTWProviderProps = {
   enableTelemetry?: boolean;
   ehr?: string;
   featureFlags?: FeatureFlags;
+  onResourceSave?: OnResourceSaveCallback;
```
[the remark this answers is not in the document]
mrosata: @WesleyKapow This PR addresses a need for an EHR embedding this library to perform work whenever a resource has been created or edited in ODS. This callback allows them to optionally hit their backend or refresh data in their own UI after the resource is saved.

The use-case we 100% want to support here is for a partner that is currently using the write-back data-loop to create resources via their API and would like to refresh relevant parts of their UI after a user has added a resource to the builder record in case their own data has been changed.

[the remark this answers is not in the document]
WesleyKapow: And that's different from the ehr-hook because it's a front-end callback so they can update the UI as appose to JUST the data in their backend?

.changeset/few-lemons-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": minor
+---
+
+Add onResourceSave callback to CTWProvider context
```
WesleyKapow, line 5:
```suggestion
Add onResourceSave callback to CTWProvider context which will get called anytime a resource is modified (edited or created).
```

src/fhir/action-helper.ts
```diff
@@ -57,8 +60,11 @@ export async function createOrEditFhirResource(
 
     return response;
   } catch (err) {
+    action = "failure";
```
WesleyKapow, line 63: This prevents consumers of `onResourceSave` from knowing what the actual action was (create or update).

Trying to think through what a good API here would be. They should get the resource, action, and error? status? response?

src/components/core/providers/patient-context.tsx
```diff
@@ -7,6 +7,7 @@ export type PatientState = {
   systemURL: string;
   tags?: Tag[];
   onPatientSave?: (data: PatientFormData) => Promise<void>;
+  onResourceSave?: (data: fhir4.Resource, action: "create" | "update") => void;
```
WesleyKapow, line 10: Thoughts on if we'll want this to be async so that we can `await onResourceSave(...)`? That way EHRs could update their state before we consider the operation complete. E.g. we'll wait for them to update their state BEFORE we close our form drawer.

src/components/core/providers/use-ctw.ts
```diff
@@ -28,6 +28,11 @@ export function useCTW() {
       builderId: context.builderId ?? claimsBuilderId(authToken) ?? "",
       contextBuilderId: context.builderId,
       fetchFromFqs: getFetchFromFqs(context.env, authToken, context.builderId),
+      onResourceSave: (resource, action) => {
+        if (context.onResourceSave) {
+          context.onResourceSave(resource, action);
+        }
+      },
```
WesleyKapow, line 35: nit: This is a little gross, alternatively you could make `onResourceSave` optional, and this would simply be:
```suggestion
      onResourceSave: context.onResourceSave,
```

Then callers would do `context.onResourceSave?.(resource, action)` to call it.

FYI, I don't have a strong opinion here.

WesleyKapow (changes requested):
LGTM, not approving yet as I think the `action: failure` is problematic and we should figure out what API we want here.

Also, I think `onResourceSave` is the right name, so good as is there!

src/fhir/action-helper.ts
mrosata: good point I'll update

src/components/core/providers/patient-context.tsx
mrosata: It should be fine for us to `await` a callback with type `() => void` regardless if it is async or not. If the type is void then the person writing the callback may write a sync or async function and return whatever they want from it iirc. Specifying `Promise<void>` forces them to write an async callback regardless if they need one or not.

src/components/core/providers/use-ctw.ts
mrosata: Yeah I want it to be optional for the consumers of the library but internally callers don't need to know whether the consumer actually implemented it or not, it's something the context provides always.

src/fhir/action-helper.ts
mrosata: Updated it to just be `"create"|"update"` and a third argument of error is passed instead where it's undefined if the operation was successful
```tsx
  onResourceSave={(resource, operation, error) => {
    if (!error) {
      celebrateGoodTimesComeOn();
    }
  }}
```

src/components/core/providers/patient-context.tsx
WesleyKapow: I just checked and you cannot `await` something that is not "thenable". How about this, go with what you have, and down the road if we want to `await`, we can change this type to be `void | Promise<void>` and then we can detect if we have a promise and await it otherwise do nothing.

WesleyKapow (approved):

PR #804, opened 2023-06-07
CTW-692 storybook

sarahmangels:
adding story for patient medications on FQS

### DIFF
--- a/src/components/content/medications/story-helpers/mocks/requests.ts
+++ b/src/components/content/medications/story-helpers/mocks/requests.ts
@@ -1,5 +1,5 @@
 import { MedicationStatement } from "fhir/r4";
-import { rest } from "msw";
+import { graphql, rest } from "msw";
 import { ComponentType, createElement } from "react";
 import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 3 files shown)


src/components/content/medications/story-helpers/mocks/requests.ts
```diff
@@ -114,7 +115,12 @@ function mockRequests() {
     }
   );
 
+  const mockFQS = graphql.query("MedicationStatements", async (req, res, ctx) =>
```
sarahmangels, line 118: @WesleyKapow this is working but now im getting `providerMedicationsFQS` for both inside and outside medications since they are both queries called "MedicationStatements" . I want to return a different value based on if the graphql query has a filter on the lens tag -- any idea how to do that?

WesleyKapow: `console.log(req)` or look at documentation. You should be able to use `req.variables` or anywhere else those filters exist on `req`.

WesleyKapow (approved):

pmadria3:
Closing and taking over updating meds stories in my Storybook ticket

PR #806, opened 2023-06-08
fix: show only renderable binaries (at this time, only CW and CQ docs)

bryanjknight:
Currently the library can only render CCDA documents; hence, we filter out documents we cannot render to reduce confusion

### DIFF
--- a/src/components/content/document/helpers/filters.ts
+++ b/src/components/content/document/helpers/filters.ts
@@ -1,12 +1,12 @@
 import { DocumentModel } from "@/fhir/models/document";
 import { isEqual, uniqWith } from "@/utils/nodash";
 
+export const THIRD_PARTY_SOURCE_SYSTEM = "https://zusapi.com/thirdparty/source";
+export const ZUS_CREATION_DATE_URL = "https://zusapi.com/created-at";
 const RAINBOW_CUTOVER_DATE = new Date("2023-03-08T05:00:00.000Z");
 
 const zusCreationDate = (doc: fhir4.DocumentReference) => {
-  const createExtension = doc.meta?.extension?.filter(
-    (ext) => ext.url === "https://zusapi.com/created-at"
-  );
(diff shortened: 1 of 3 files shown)


src/components/content/document/helpers/filters.ts
```diff
@@ -27,9 +27,17 @@ const isViewablePostRainbow = (docRef: fhir4.DocumentReference) => {
   );
 };
 
+const isRenderableBinary = (doc: fhir4.DocumentReference): boolean => {
+  const thirdPartyTag = doc.meta?.tag?.find((tag) => tag.system === THIRD_PARTY_SOURCE_SYSTEM);
+  const isSupportedThirdParty = ["commonwell", "carequality"].includes(thirdPartyTag?.code || "");
```
WesleyKapow, line 32: nit: for things that may be `undefined` or `null`, favor using `??` instead of `||`. It's a good habit to avoid `0 ?? default_value` bugs.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/components/content/document/helpers/filters.ts
```diff
@@ -27,9 +27,17 @@ const isViewablePostRainbow = (docRef: fhir4.DocumentReference) => {
...
+  return isSupportedThirdParty;
+};
+
 export const applyDocumentFilters = (data: fhir4.DocumentReference[]) => {
   const documentModels = data
-    .filter((doc) => isViewablePreRainbow(doc) || isViewablePostRainbow(doc))
+    .filter(
```
WesleyKapow, line 38: Alternatively you could do
```ts
.filter(isRenderableBinary)
.filter((doc) => isViewablePreRainbow(doc) || isViewablePostRainbow(doc))
```

The bool logic you got isn't wild yet so this isn't a problem, but for more complex logic we can break things down to multiple `.filter` statements. Helpful to isolate the reasons for being filtered too.

WesleyKapow (approved):
LGTM, few nit/comments

PR #812, opened 2023-06-09
WIP: Initial hookup of Unleash directly to components

patrickvalle:
https://zeushealth.atlassian.net/browse/CTW-1132

This appears to be functional locally for the conditions component; the only issue I'm seeing is that there appears to be a race condition where the first couple calls to fetch the feature flag return `false` when it should be `true`, and sometimes the call to fetch data will go to ODS instead of FQS

### DIFF
--- a/src/components/core/providers/ctw-provider.tsx
+++ b/src/components/core/providers/ctw-provider.tsx
@@ -1,12 +1,11 @@
 import { QueryClientProvider } from "@tanstack/react-query";
-import { FlagProvider } from "@unleash/proxy-client-react";
-import jwtDecode from "jwt-decode";
 import { PropsWithChildren, useEffect, useMemo } from "react";
 import {
   AuthenticationProvider,
   AuthenticationProviderProps,
 } from "./authentication/authentication-provider";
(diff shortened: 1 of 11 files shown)


src/components/core/providers/ctw-provider.tsx
```diff
@@ -54,6 +56,14 @@ export function CTWProvider({
     };
   }, []);
 
+  const unleashConfig = {
+    url: "https://unleash-proxy-prod.zusapi.com/proxy",
+    clientKey: "MDE0NDU5NTQtNEIyNC00RUVGLUI4NDUtRTE3QjYyMUQ3NTAzCg==",
+    refreshInterval: 15, // (in seconds)
+    appName: "ctw-component-library",
+    context: authToken ? getUnleashContext(authToken) : {},
```
WesleyKapow, line 64: This will always be `{}` when we use `authTokenURL`

WesleyKapow: I think this will need to be dynamic and depend on `AuthenticationProvider`.

patrickvalle:
Closing in favor of https://github.com/zeus-health/ctw-component-library/pull/819

PR #819, opened 2023-06-12
CTW-1132: Feature-flaggable components

patrickvalle:
https://zeushealth.atlassian.net/browse/CTW-1132

This PR introduces:
1. `<FeatureFlagProvider>` provider that wraps Unleash's `FlagProvider` and ensures the connection to the Unleash proxy as well as being responsible for setting up the Unleash context
2. `useFeatureVariant` hook that ensures Unleash is done loading before fetching an Unleash variant
3. `useFQSFeatureToggle` hook that returns whether or not FQS is enabled for the specified resource type

Additionally, a subset of the Conditions components have been updated to now manage reading their own feature flags.

### DIFF
--- a/src/components/content/conditions/patient-conditions-outside-badge.tsx
+++ b/src/components/content/conditions/patient-conditions-outside-badge.tsx
@@ -3,7 +3,20 @@ import { useFQSFeatureToggle } from "@/hooks/use-fqs-feature-toggle";
 import { usePatientConditionsOutside } from "@/services/conditions";
 
 export const PatientConditionsOutsideBadge = () => {
-  const enableFQS = useFQSFeatureToggle("conditions");
+  const fqs = useFQSFeatureToggle("conditions");
+  if (!fqs.ready) {
+    return <></>;
+  }
+  return <CorePatientConditionsOutsideBadge enableFQS={fqs.enabled} />;
+};
+
+type CorePatientConditionsOutsideBadgeProps = {
+  enableFQS: boolean;
+};
+
+const CorePatientConditionsOutsideBadge = ({
+  enableFQS,
+}: CorePatientConditionsOutsideBadgeProps) => {
   const otherConditionsQuery = usePatientConditionsOutside(enableFQS);
   const activeUnarchivedConditions = otherConditionsQuery.data.filter(
     (condition) => condition.active && !condition.isArchived
--- a/src/components/content/conditions/patient-conditions-outside-badge.tsx
+++ b/src/components/content/conditions/patient-conditions-outside-badge.tsx
@@ -1,13 +1,9 @@
 import { Badge } from "@/components/core/badge";
+import { useFQSFeatureToggle } from "@/hooks/use-fqs-feature-toggle";
 import { usePatientConditionsOutside } from "@/services/conditions";
 
-export type PatientConditionsOutsideBadgeProps = {
-  enableFQS?: boolean;
-};
-
-export const PatientConditionsOutsideBadge = ({
-  enableFQS = false,
-}: PatientConditionsOutsideBadgeProps) => {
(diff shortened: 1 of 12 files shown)


src/components/content/conditions/patient-conditions-outside-badge.tsx
```diff
@@ -1,13 +1,22 @@
 import { Badge } from "@/components/core/badge";
+import { useFQSFeatureToggle } from "@/hooks/use-fqs-feature-toggle";
 import { usePatientConditionsOutside } from "@/services/conditions";
 
-export type PatientConditionsOutsideBadgeProps = {
-  enableFQS?: boolean;
+export const PatientConditionsOutsideBadge = () => {
```
patrickvalle, line 5: Components are "wrapped" to first check for FQS or ODS before rendering the actual core component

src/components/content/conditions/patient-conditions-outside-badge.tsx
```diff
@@ -1,13 +1,22 @@
...
+  const fqs = useFQSFeatureToggle("conditions");
+  if (!fqs.ready) {
+    return <></>;
+  }
+  return <CorePatientConditionsOutsideBadge enableFQS={fqs.enabled} />;
+};
+
+type CorePatientConditionsOutsideBadgeProps = {
+  enableFQS: boolean;
 };
```
mrosata, line 15: First thought is that this doesn't seem the right pattern here. Could the hooks that fetch data consume this hook instead?

This is probably very crude, but the idea being that until the feature flag resolves we return empty state from the query and then once the feature flag is loaded the cache is invalidated (because query key IS_FQS_READY changes) and it triggers one of the actual query functions to fetch results
```ts
export function usePatientAllergies(enableFQS: boolean) {
  const fqs = useFQSFeatureToggle("allergies");
  const IS_FQS_READY = fqs.ready;
  return useQueryWithPatient(
    QUERY_KEY_PATIENT_ALLERGIES,
    [IS_FQS_READY],
    async () => {
    if (!fqs.ready) {
      return { data: [] }
    }
    return fqs.enabled ?
    ? withTimerMetric(
        async (requestContext, patient) => getAllergyIntoleranceFromFQS(requestContext, patient),
        "req.timing.allergies",
        ["FQS"]
      )()
    : withTimerMetric(
        async (requestContext, patient) => getAllergyIntoleranceFromODS(requestContext, patient),
        "req.timing.allergies"
      )()
    }
  );
}
```

If that's not possible and this is the pattern that is going to be used on every data component. It might be worth abstracting it to a higher order function or class.

WesleyKapow: I like this approach, one gotcha is if we return an empty array, how will we know we are still loading? Ideally we'd show a loading spinner until feature flags are ready and we fetched from the correct place.

WesleyKapow: Maybe instead we use the `enabled` flag on the queries?

WesleyKapow: Oh I remember, the issue would be we'd be setting up the ODS query with `enabled: false`.. we don't want to setup ODS or FQS until we know our feature flags.

patrickvalle: > one gotcha is if we return an empty array, how will we know we are still loading
I thought about this too, but I think we effectively have the same issue w/ the "wrapped component" approach in this PR... the difference is we don't paint the component at all until the flag is fetched vs painting it's skeleton and not showing any data yet

patrickvalle:
Closing in favor of https://github.com/zeus-health/ctw-component-library/pull/820

PR #831, opened 2023-06-15
Ctw 1200/rewire use history

pmadria3:
This PR sets up the drawer to retrieve the versions history for resources (currently conditions) and also rewires the useHistory hook to utilize FQS for resources.


src/components/content/resource/history.ts
```diff
@@ -29,53 +37,110 @@ export function useHistory<T extends ResourceTypeString, M extends FHIRModel<Res
...
-        return entries;
-      } catch (e) {
-        throw Telemetry.logError(
-          e as Error,
-          `Failed fetching condition history for patient: ${patient.UPID}}`
-        );
-      }
-    }, `req.${model.resourceType.toLowerCase()}_history`),
+    enableFQS
+      ? withTimerMetric(
```
bryanjknight, line 47: Let's break these up into two functions; it's a bit hard to read with the inline if-else statement

bryanjknight (changes requested):

src/components/content/resource/history.ts
```diff
@@ -150,3 +147,149 @@ function wasEnteredInError(resource: fhir4.FhirResource) {
...
 }
+
+function getResourceFQSQuery(resourceType: ResourceTypeString) {
+  switch (resourceType) {
+    case "Condition":
+      return conditionsQuery;
+    case "AllergyIntolerance":
+      return allergyQuery;
+    default:
+      return "";
```
WesleyKapow, line 158: Let's make sure we don't silently error if we accidentally use this when it isn't setup.
```suggestion
      throw new Error(`Resource type to FQS query not implemented yet for ${resourceType}`);
```

src/services/fqs/queries/versions.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  `;
+}
+
+function getResourceFragment(resourceType: ResourceTypeString) {
+  switch (resourceType) {
+    case "Condition":
+      return fragmentConditionHistory;
+    default:
+      return "";
```
WesleyKapow, line 25: Another `throw` here.

src/services/fqs/queries/fragments/condition.ts
```diff
@@ -0,0 +1,224 @@
...
+      type {
+        coding {
+          ...Coding
+        }
+      }
+    }
+  }
+`;
+
+export const fragmentConditionHistory = gql`
```
WesleyKapow, line 117: What's the difference between this and the one above? Is there a way to DRY this up?

src/components/content/resource/history.ts
```diff
@@ -150,3 +147,149 @@ function wasEnteredInError(resource: fhir4.FhirResource) {
...
+  }
+}
+
+export async function getVersionHistoryFQS<T extends ResourceTypeString>(
```
WesleyKapow, line 162:
```suggestion
async function getVersionHistoryFQS<T extends ResourceTypeString>(
```

WesleyKapow (changes requested):
Looks great, few nitpicks and then it'll be good to go!

src/components/content/resource/history.ts
```diff
@@ -293,3 +306,25 @@ async function fetchResourcesODS<
     );
   }
 }
+
+export function filterLensandSummary<T extends ResourceTypeString>(
```
WesleyKapow, line 310: nit: `filterLens**A**ndSummary` (capital A)

src/components/content/resource/history.ts
```diff
@@ -39,7 +43,7 @@ export function useHistory<T extends ResourceTypeString, M extends FHIRModel<Res
   getSearchParams,
   getHistoryEntry,
   getFiltersFQS,
-  enableFQS = true,
+  enableFQS = false,
```
WesleyKapow, line 46: nit: Do we need a default here? Do all consumers of this now explicitly pass this?

PR #832, opened 2023-06-15
POC for writing documents directly to EHR API

eric-gunther:
Exposes an `onAddToRecord` prop for `PatientDocuments` that accepts a callback function that takes a `DocumentModel` and `fhir4.Binary`. If that prop is set then we show an "Add" button on the documents component that executes the callback function. Before executing the callback it retrieves the binary for the document. Ultimately this can be used by consumers to trigger a post to an endpoint in their API that imports documents.

I think an interesting longer term option here may be a more generic prop called `additionalRowActions` that takes an array of callback functions and names (for the buttons) so that consumers can trigger whatever they want for an arbitrary number of actions. But for now I plan to leave this as a single onAddToRecord prop.


src/App.tsx
```diff
@@ -18,6 +18,9 @@ import {
   PatientTimeline,
   ZusAggregatedProfile,
 } from ".";
+import { title } from "process";
+import { render } from "react-dom";
```
mrosata, line 22: I believe your IDE added these but shouldn't be needed
```suggestion
```

src/components/content/document/patient-documents.tsx
```diff
@@ -1,26 +1,36 @@
...
+import { usePatientDocuments } from "@/fhir/document";
 import { DocumentModel } from "@/fhir/models/document";
 import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
+import { useBaseTranslations } from "@/i18n";
 
 export type PatientDocumentProps = {
   className?: string;
   enableFQS?: boolean;
+  // could call this additionalRowActions and make it an array
+  onAddToRecord?: (document: DocumentModel, binary: fhir4.Binary) => void;
```
mrosata, line 21: maybe with something like
```suggestion
  rowAction: RowAction;
```
or 
```suggestion
  rowActions: RowAction[];
```
where `RowAction` is
```tsx
type RowAction = {
  text: string;
  onAction: (document: DocumentModel, binary: fhir4.Binary) => void;
  className?: cx.Argument;
};
```
and that way consumer of the library who is implementing this action is required to add
 - `text` - Text that appears on the button
 - `onAction` - Callback to handle the click, may be async, if this errors, we should show this error message in the UX.
 - `className` - optionally add a classname to the button

src/components/content/document/patient-documents.tsx
```diff
@@ -51,11 +61,46 @@ function PatientDocumentsComponent({ className, enableFQS = false }: PatientDocu
...
+        data-testid="add-to-record"
+        onClick={async () => {
+          const binary = await getBinaryDocument(await getRequestContext(), binaryID);
+          onAddToRecord(record, binary);
+        }}
+      >
+        {t("resourceTable.add")}
+      </button>
+    </div>
+  ) : null;
```
mrosata, line 101: If we take the route above we should add a useState variable to this component like
```
const [actionError, setActionError] = useState<string>();
```
and down here we could catch any error and set it to `actionError`
```suggestion
  return rowAction && binaryID ? (
    <div className="ctw-flex ctw-space-x-2">
      <button
        type="button"
        className={cx(rowAction.className, "ctw-btn-primary ctw-ml-1 ctw-capitalize")}
        data-zus-telemetry-click="Document custom row action"
        data-testid="custom-row-action"
        onClick={async () => {
          try {
            const binary = await getBinaryDocument(await getRequestContext(), binaryID);
            rowAction.onAction(record, binary);
          } catch (err) {
            setActionError(err.message ?? "unknown error")
          }
        }}
      >
        {rowAction.text}
      </button>
    </div>
  ) : null;
```

src/components/content/document/patient-documents.tsx
```diff
@@ -53,11 +58,46 @@ function PatientDocumentsComponent({ className }: PatientDocumentProps) {
...
+type ExtraRowActionProps = {
+  onAddToRecord?: (document: DocumentModel, binary: fhir4.Binary) => void;
+};
+
+const getRowActions =
+  ({ onAddToRecord }: ExtraRowActionProps) =>
+  (props: RowActionsProps<DocumentModel>) =>
+    <RowActions {...props} onAddToRecord={onAddToRecord} />;
+
+type RowActionsProps2 = RowActionsProps<DocumentModel> & ExtraRowActionProps;
```
bryanjknight, line 76: Slight aversion to using `Thing2` for naming -- perhaps `ExtendedRowActionProps`?

bryanjknight (approved):
one nit but otherwise lgtm

PR #833, opened 2023-06-15
CDEV-52: Fixing race condition w/ feature flags

patrickvalle:
https://zeushealth.atlassian.net/browse/CDEV-52

The Unleash client now is explicitly stopped before updating the Unleash context and start after the context is updated.

### DIFF
--- a/.changeset/silly-taxis-join.md
+++ b/.changeset/silly-taxis-join.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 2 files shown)


bryanjknight (approved):

patrickvalle:
> LGTM, I thought you had issues with `startClient={false}` where unleash hooks would fail when it wasn't started yet.

We did in one of the earlier iterations, but now that no longer appears to be true. I can't fully recall what the state of the provider/hook was when we saw that issue

PR #836, opened 2023-06-15
CDEV-52: Fixing flag ready slowness

patrickvalle:
https://zeushealth.atlassian.net/browse/CDEV-52

Removing the "on client update" listener that was originally put in so that `useFeatureVariant` would (presumably) wait until the Unleash client had been updated w/ the user context before returning flags. However, this should no longer be needed after the explicit client start/stop fix introduced [here](https://github.com/zeus-health/ctw-component-library/pull/833/files#diff-a98d6d28f3b21099b8ab57cd7e1478dd7e17b477159b55a3af61509509524572R50-R55).

The "on client update" hook is causing the standalone app to delay marking feature flags as ready, which then delays when Conditions data gets fetched.

### DIFF
--- a/.changeset/gentle-pugs-pay.md
+++ b/.changeset/gentle-pugs-pay.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Fixed issue where Unleash client was sometimes taking too long to load flags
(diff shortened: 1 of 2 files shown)


WesleyKapow (approved):
LGTM but so did the other code 😬 🤷‍♂️

PR #839, opened 2023-06-16
CDEV-210: Update Allergies components to read FQS feature flags directly from Unleash

patrickvalle:
https://zeushealth.atlassian.net/browse/CDEV-210

Allergies component no longer reads whether or not FQS is enabled a property; now fetches toggle directly from Unleash


WesleyKapow (approved):
LGTM

PR #843, opened 2023-06-20
fix: don't hard fail if creation date is missing, instead don't show the document

bryanjknight:
It is not guaranteed for a zus creation date to exist on a document reference; thus, we will default to not showing the document if the creation date is not present

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -74,7 +74,11 @@ const components: DemoComponent[] = [
     render: () => <PatientConditionsOutside />,
     title: "Patient Conditions Outside",
   },
-  { name: "documents", render: () => <PatientDocuments />, title: "Patient Documents" },
+  {
+    name: "documents",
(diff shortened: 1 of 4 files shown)


sarahmangels (dismissed):

src/App.tsx
```diff
@@ -74,7 +74,11 @@ const components: DemoComponent[] = [
     render: () => <PatientConditionsOutside />,
     title: "Patient Conditions Outside",
   },
-  { name: "documents", render: () => <PatientDocuments />, title: "Patient Documents" },
+  {
+    name: "documents",
+    render: () => <PatientDocuments enableFQS={true} />,
+    title: "Patient Documents",
+  },
```
luthfib, line 81: Is FQS always enabled because this only works on FQS?

pmadria3 (approved):

PR #852, opened 2023-06-21
CTW-1240 Distinguish builder and outside data in allergy component

eric-gunther:
This PR primarily focuses on the following improvements to the allergy component:

- Better de-duping of builder records vs outside records. Builder records are prioritized over outside records and will be displayed with a check icon.
- Outside records now have a "Dismiss" button that works the same as it does for medications and conditions.
- Outside records also have a "Mark as Viewed" button that makes an outside record go from bold to normal font. Clicking on an outside record to view the details also has this effect. This is the first step towards having this functionality in other components.

There's also some general refactoring around re-naming "archive" to "dismiss" in a bunch of places to better match the current product terminology.

https://github.com/zeus-health/ctw-component-library/assets/97455910/6734fa58-38f4-479e-b826-9462d19bce26


src/components/content/allergies/helpers/history.tsx
```diff
@@ -19,6 +19,11 @@ export function useAllergiesHistory(enableFQS: boolean, allergy: AllergyModel) {
 }
 
 function getSearchParams(allergy: AllergyModel) {
+  // TODO - this will only find other allergies with the same code, but it does
+  // not handle allergies that have a different code but the same display, which is
+  // what we de-dupe on.
+  // Perhaps a better solution would be to just get all the patients allergies
+  // and pull out the relevant ones.
   const tokens = allergy.knownCodings.map((coding) => `${coding.system}|${coding.code}`);
```
eric-gunther, line 27: We currently get the history for a given allergy by searching for other allergies with a matching code/system. However many allergies are just going to have a codeable concept containing only a text value (like "bee pollen") and this will not find other allergies in the FHIR store with just that text value.

Since we're already getting all of the patients allergies in the first request to build the table I suggest we just use that same set of data to determine the history for a single allergy.

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -30,6 +30,17 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
...
+
+  useEffect(() => {
+    async function load() {
+      const requestContext = await getRequestContext();
+      setUserBuilderId(requestContext.builderId);
+    }
+
+    void load();
+  }, [getRequestContext]);
+
```
eric-gunther, line 43: This seems like a very awkward way to get the user's builder ID so that I can pass it along to the columns function, but I couldn't find a better precedent elsewhere in the codebase. Am I doing something wrong?

src/components/content/allergies/helpers/allergies-filter.tsx
```diff
@@ -1,21 +1,30 @@
...
+    const existingAllergy = allergyDataMap.get(lowercaseDisplay);
 
-  return allergyData;
-};
+    if (!existingAllergy) {
+      allergyDataMap.set(lowercaseDisplay, allergy);
+    } else if (allergy.ownedByBuilder(builderId) && !existingAllergy.ownedByBuilder(builderId)) {
+      allergyDataMap.set(lowercaseDisplay, allergy);
+    }
+  });
```
mrosata, line 25: To make the `Map<string, AllergyModel>` _(instead of needing `string|undefined` as the map key)_ We could use lodash `lowerCase(this.display)` which would ensure that key is always a string, as lodash will return empty string when passed undefined.

---

Mini rant: The reason TypeScript would have prodded us in the `Map<string | undefined, AllergyModel>` direction instead of plain `Map<string, AllergyModel>` is because our getter `allergyModel.lowercaseDisplay ` _(and in turn `allergyModel.display`)_ returns `string|undefined`. Those accessor helpers on our FHIR models that do their best to return some meaningful value _(if possible)_ by cascading through possibly undefined FHIR resource attributes, should probably also return a static type like `string`.

There are a few places this might not play well, but with properties like `display` or `lowercaseDisplay` it feels always getting a string would simplify some typing inside components.

It would even simplify some code. It looks like @bryanjknight had to implement `lowercaseDisplay` like this
```ts
  get lowercaseDisplay(): string | undefined {
    return this.display ? this.display.toLowerCase() : this.display;
  }
```
because I'm assuming he didn't want to have to write
```ts
this.display ? this.display.toLowerCase() : this.display;
```
in multiple places. However, if `allergyModel.display` always returned a string, not only could `lowercaseDisplay` have been written like this
```ts
  get lowercaseDisplay(): string | undefined {
    return this.display.toLowerCase();
  }
```

but it likely wouldn't have needed to have been created in the first place as in the view layer one could simply and safely do things like
```tsx
  <div>{allergy.display.toLowerCase}</div>
```

src/components/content/allergies/patient-allergies.tsx
mrosata: I think you could do
```tsx
const { builderId } = useContext(CTWStateContext)
```
But since it's on that context already, maybe it's better to add the builderId to the return value of [`useCTW()`](https://github.com/zeus-health/ctw-component-library/blob/773d86c48047dbb073a2917623b640e523e623e0/src/components/core/providers/use-ctw.ts#L40-L43) so it would instead be
```tsx
const { builderId } = useCTW();
```
or we could make an `useBuilderId`
```tsx
function useBuilderId() {
  const ctx = useContext(CTWStateContext);
  return ctx.builderId;
}
```

mrosata (approved):

src/fhir/models/fhir-model.ts
```diff
@@ -88,6 +94,19 @@ export abstract class FHIRModel<T extends fhir4.Resource> {
...
+  // Returns true if this resource is owned by the specified builder.
+  ownedByBuilder(builderId: string): boolean {
+    try {
+      if (this.resource.meta?.tag?.some((tag) => tag.system === SYSTEM_ZUS_THIRD_PARTY)) {
+        return false;
+      }
+      const ownerTag = this.resource.meta?.tag?.find((tag) => tag.system === SYSTEM_ZUS_OWNER);
+      return ownerTag?.code?.split("/")[1] === builderId;
+    } catch (error) {
+      throw new Error("Expected builder owner tag missing or malformed.");
```
bryanjknight, line 106: Throwing an error recently burned me where it resulted in the entire page not rendering any data (hence a zero moment". I would suggest having some additional safe guards, particularly around the owner tag splitting given some of our older data may not have it in the same format

src/components/core/providers/use-ctw.ts
```diff
@@ -39,6 +39,7 @@ export function useCTW() {
 
   return {
     getRequestContext,
+    builderId: context.builderId || "",
```
eric-gunther, line 42: Unsure how I feel about this coalesce but it makes using this `builderId` downstream much easier...

src/components/content/allergies/helpers/allergies-filter.tsx
```diff
@@ -1,21 +1,30 @@
...
+    const { lowercaseDisplay } = allergy;
+    const existingAllergy = allergyDataMap.get(lowercaseDisplay);
 
-  return allergyData;
-};
+    if (!existingAllergy) {
+      allergyDataMap.set(lowercaseDisplay, allergy);
+    } else if (allergy.ownedByBuilder(builderId) && !existingAllergy.ownedByBuilder(builderId)) {
+      allergyDataMap.set(lowercaseDisplay, allergy);
+    }
```
mrosata, line 24: Regardless of which condition matches, they both do the same `allergyDataMap.set(lowercaseDisplay, allergy)`. Perhaps it's broken out into multiple if statements for readability, if so we could use variables for the conditions to summarize
```suggestion
    const allergyOwnedByBuilder = allergy.ownedByBuilder(builderId);
    const existingAllergyNotOwned = !existingAllergy.ownedByBuilder(builderId);
    if (!existingAllergy || (allergyOwnedByBuilder && existingAllergyNotOwned)) {
      allergyDataMap.set(lowercaseDisplay, allergy);
    }
```

src/components/core/table/table.scss
```diff
@@ -69,7 +69,9 @@
 
   .ctw-table-row-actions {
     background: linear-gradient(90deg, #f9fafb00 0%, #f9fafb 4.17%);
-    @apply ctw-invisible ctw-absolute ctw-right-0 ctw-top-0 ctw-z-20 ctw-flex ctw-h-full ctw-items-center ctw-space-x-2 ctw-px-4;
+    height: calc(100% - 3px) !important;
```
WesleyKapow, line 72: 3px? Was 1 not enough? This will cause the vertical centering to be slightly off.

eric-gunther: I had changed to 3px in the course of some troubleshooting because the bottom border was then disappearing, but now it seems to be working with just 1px.

src/components/content/allergies/helpers/allergies-filter.tsx
```diff
@@ -1,21 +1,31 @@
...
+      (allergy.ownedByBuilder(builderId) && !existingAllergy.ownedByBuilder(builderId))
+    ) {
+      allergyDataMap.set(lowercaseDisplay, allergy);
+    }
+  });
+
+  const uniqueAllergies = Array.from(allergyDataMap.values());
 
-const valuesToDedupeOn = (allergy: AllergyModel) => [allergy.lowercaseDisplay];
+  return uniqueAllergies;
```
WesleyKapow, line 30: Simplification:

```ts
// Bump builder owned allergies to the front, so uniqBy favors them!
const builderOwnedFirst = orderBy(sortedByDate, (a) => a.ownedByBuilder(builderId), 'desc')
return uniqBy(builderOwnedFirst, 'lowercaseDisplay')
```

`orderBy(sortedByDate...)` should preserve the date ordering too.. so it will end up being, all builder owned allergies first, sorted by date, followed by all non-builder owned allergies sorted by date.

src/components/content/allergies/helpers/column.tsx
```diff
@@ -1,23 +1,29 @@
...
   const allergyColumns: TableColumn<AllergyModel>[] = [
     {
       title: "Name",
       render: (allergy) => (
         <div>
-          <div className="ctw-font-medium">{capitalize(allergy.display)} </div>
+          <div className="ctw-flow-root">
+            {capitalize(allergy.display)}
+            <span className="ctw-float-right">
+              {allergy.ownedByBuilder(builderId) ? <CheckIcon /> : <></>}
```
WesleyKapow, line 16: As mentioned elsewhere, this would instead be `<FontAwesomeIcon icon={faCheck}/>`

src/components/content/allergies/helpers/column.tsx
```diff
@@ -1,23 +1,29 @@
...
+          </div>
         </div>
       ),
     },
     {
       title: "Last Updated",
       render: (allergy) => (
         <div>
-          <div className="ctw-font-medium">{allergy.recordedDate} </div>
+          <div>{allergy.recordedDate} </div>
```
WesleyKapow, line 26: Trailing space inside div needed here?

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -38,13 +43,20 @@ function PatientAllergiesComponent({ className }: PatientAllergiesProps) {
...
+  const rowActions = useMemo(() => getRowActions(builderId), [builderId]);
+
+  const { toggleRead } = useToggleRead(QUERY_KEY_PATIENT_ALLERGIES, QUERY_KEY_BASIC);
+
+  const handleRowClick = (record: AllergyModel) => {
+    if (!record.isRead) {
+      toggleRead(record);
+    }
+    openDetails(record);
+  };
```
WesleyKapow, line 55: Ah neat! I can also see us rolling this into `openDetails` in the future, if this becomes the standard across our resources. Could be an option passed to `useResourceDetailsDrawer`.

src/components/content/allergies/patient-allergies.tsx
```diff
@@ -84,3 +96,64 @@ const allergyData = (allergy: AllergyModel) => [
...
+      <div className="ctw-flex ctw-space-x-2">
+        <button
+          type="button"
+          className="ctw-btn-default"
+          disabled={isToggleDismissLoading || isToggleReadLoading}
+          onClick={() => {
+            toggleDismiss(record);
+            if (!record.isRead) {
+              toggleRead(record);
+            }
```
WesleyKapow, line 130: This is interesting to me. So dismiss is treated like read? Would we ever want to distinguish these? I could see the UI treating dismissed as read (not bold), but unclear if we want to act like it was read in this case.

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -0,0 +1,48 @@
+import { useCallback, useState } from "react";
+import { useCTW } from "@/components/core/providers/use-ctw";
+import { toggleRead } from "@/fhir/basic";
+import { FHIRModel } from "@/fhir/models/fhir-model";
+import { queryClient } from "@/utils/request";
+
+interface UseToggleReadResult {
+  /**
+   * Function to call to toggle the archive status of the FHIR model
```
WesleyKapow, line 9:
```suggestion
   * Function to call to toggle the read status of the FHIR model
```

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  toggleRead: (model: FHIRModel<fhir4.Resource>) => void;
+
+  /**
+   * True when `toggleArchive` is called
```
WesleyKapow, line 14:
```suggestion
   * True when `toggleRead` is called
```

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  isLoading: boolean;
+}
+
+/**
+ * This hook is toggles the read status for the specified FHIR model.
```
WesleyKapow, line 20:
```suggestion
 * This hook toggles the read status for the specified FHIR model.
```

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -0,0 +1,48 @@
...
+    setIsLoading(true);
+    await toggleRead(model, await getRequestContext());
+    await Promise.all(
+      queriesToInvalidate.map((queryToInvalidate) =>
+        queryClient.invalidateQueries([queryToInvalidate])
+      )
+    );
+
+    // Timeout here fixes bug where we would briefly flash
+    // the old read/restore text.
```
WesleyKapow, line 39:
```suggestion
    // the old mark as read/new text.
```

WesleyKapow (changes requested):

src/fhir/models/fhir-model.ts
```diff
@@ -88,6 +102,20 @@ export abstract class FHIRModel<T extends fhir4.Resource> {
     return JSON.stringify(this.resource).includes(SYSTEM_ENRICHMENT);
   }
 
+  // Returns true if this resource is owned by the specified builder.
+  ownedByBuilder(builderId: string): boolean {
```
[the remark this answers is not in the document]
eric-gunther: `ownedByBuilder` is intended to indicate if the data is "first party" data that your organization wrote to Zus. `isSummaryResource` indicates if the resource was generated by our lens process - it's not really 3rd party or 1st party data, it's just special Zus data that merges the two. I think these are meaningful distinctions that drive different behavior in the app.

src/components/content/allergies/patient-allergies.tsx
eric-gunther: I think this is subject to change, but I believe "read" means "I interacted with this record" - ie. you clicked it to view the details, you dismissed it, or you explicitly clicked "Mark as Viewed" (note that "read" and "viewed" have been used interchangeably and I think we're still debating what the final word may be).

src/components/content/hooks/use-toggle-archive.ts
```diff
@@ -17,22 +17,18 @@ interface UseToggleArchiveResult {
...
-  model: FHIRModel<T>,
-  ...queriesToInvalidate: string[]
-): UseToggleArchiveResult {
+export function useToggleDismiss(...queriesToInvalidate: string[]): UseToggleDismissResult {
   const { getRequestContext } = useCTW();
 
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
-  const handleToggleArchive = useCallback(async () => {
+  const handleToggleDismiss = useCallback(async (model: FHIRModel<fhir4.Resource>) => {
```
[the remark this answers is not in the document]
eric-gunther: Did that so I could use it in the row click handler:

```const { toggleRead } = useToggleRead(QUERY_KEY_PATIENT_ALLERGIES, QUERY_KEY_BASIC);

  const handleRowClick = (record: AllergyModel) => {
    if (!record.isRead) {
      toggleRead(record);
    }
    openDetails(record);
  };```

File: src/components/content/allergies/patient-allergies.tsx (hunk 11 above)
Comment (eric-gunther) [reply src/components/content/allergies/patient-allergies.tsx:55]: Yeah agreed that there's room for refactoring once it's universally rolled out across components.

File: src/fhir/allergies.ts [hunk 19]
```diff
@@ -1,23 +1,59 @@
...
+      patientAllergies.forEach((a, i) => {
+        const filteredBasics = basics.filter(
+          (b) => b.subject?.reference === `AllergyIntolerance/${a.id}`
+        );
+        patientAllergies[i].revIncludes = filteredBasics;
+      });
+    }
+
+    setAllergies([...patientAllergies]); // spread syntax here needed to make sure the array is a new reference in order to trigger a re-render
+  }, [basicsQuery.data, patientAllergiesQuery.data]);
```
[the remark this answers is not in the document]
eric-gunther: Yeah it's kinda ugly. Given where I'm at with this PR (and the fact that you didn't mark this as blocker) I'd like to punt on this refactor and address it as I need to add similar functionality for other resources.

eric-gunther:
Fair amount of changes made via PR feedback but functionality still looking good, here's the latest in a video - 

https://github.com/zeus-health/ctw-component-library/assets/97455910/8d5de355-de1a-4498-9f41-7b32ca99e2c6

WesleyKapow (dismissed):
Looks good, you should also remove the `CheckIcon` you added.

WesleyKapow (approved):

PR #860, opened 2023-06-22
CDEV-90/CDEV-212/timeline on FQS

WesleyKapow:
CDEV-90
CDEV-212

This adds support to fetch timeline data (encounters, diagnostic reports and observations) from FQS. This builds upon previous work to only fetch from FQS if the correct feature-flag variant is enabled.

Additionally, created a new `useFeatureFlaggedQueryWithPatient` wrapper around the existing `useQueryWithPatient` that helps clean up our FQS or ODS branching logic.


pmadria3 (approved):
LGTM

PR #865, opened 2023-06-23
Cdev222/filtering allergies

pmadria3:
CDEV-222

This PR sets up allergies filtering the drawer client side and also utilizes filtering via `system|code` if there is otherwise filtering via `text`


mrosata (approved):
lgtm

PR #870, opened 2023-06-26
CDEV-56: update medication lens query to use the summary|Common tag

sarahmangels:
switch from using the lens|ActiveMedication tag to the summary|Common tag

### DIFF
--- a/src/fhir/search-helpers.ts
+++ b/src/fhir/search-helpers.ts
@@ -159,16 +159,12 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
   return records;
 }
 
-// Like searchAllRecords, but filters down to lens records.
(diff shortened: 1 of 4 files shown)


src/fhir/search-helpers.ts
```diff
@@ -159,16 +159,12 @@ export async function searchBuilderRecords<T extends ResourceTypeString>(
   return records;
 }
 
-// Like searchAllRecords, but filters down to lens records.
-export type LensTag = "ActiveMedications" | "ChronicConditions";
-
 export async function searchLensRecords<T extends ResourceTypeString>(
```
sarahmangels, line 162: it looks like this function isnt being used by condition lens at all so I rm'd the lensTag param

pmadria3 (approved):

PR #880, opened 2023-06-29
Rewire Careteam to FQS

pmadria3:
This PR rewires the CareTeam resource to FQS. 

Screenshot 2023-06-28 at 9 48 01 PM

### DIFF
--- a/src/fhir/care-team.ts
+++ b/src/fhir/care-team.ts
@@ -1,34 +1,61 @@
 import { getIncludedResources } from "./bundle";
 import { searchCommonRecords } from "./search-helpers";
+import { CTWRequestContext, PatientModel, useFeatureFlaggedQueryWithPatient } from "..";
 import { applyCareTeamFilters } from "@/components/content/care-team/helpers/filters";
-import { useQueryWithPatient } from "@/components/core/providers/patient-provider";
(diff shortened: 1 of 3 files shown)


eric-gunther (approved):

src/fhir/care-team.ts
```diff
@@ -1,34 +1,61 @@
...
-          }
-          Telemetry.histogramMetric("req.count.care_teams", results.length);
-          return results;
-        } catch (e) {
-          throw new Error(`Failed fetching care team information for patient: ${e}`);
-        }
-      },
-      "req.timing.care_teams",
-      []
-    )
```
mrosata, line 32: Are we removing the timing metric on purpose? Is it because it's tracked some other way?

pmadria3: I added it on line 15

src/services/fqs/queries/care-teams.ts
```diff
@@ -0,0 +1,89 @@
...
+            }
+            versionId
+          }
+          status
+          category {
+            text
+            coding {
+              ...Coding
+            }
+          }
```
WesleyKapow, line 44: FYI, I added a `fragmentCodeableConcept`, so this could be:

```
category {
 ...codeableConcept
}
```

PR #882, opened 2023-06-29
CTW-1240 Unread allergies badge

eric-gunther:
New notification icon in the ZAP allergy tab to indicate that there are unread allergies. Also replacing conditions/meds badges with the new notification icon and moving everything to the left of the tab title (both suggestions per product/UX guidance).

https://github.com/zeus-health/ctw-component-library/assets/97455910/a944be8c-907d-46ec-aecf-7c7bff7dab44


src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
```diff
@@ -50,7 +51,12 @@ export const zusAggregatedProfileTabs: ZusAggregatedProfileTabs = {
   allergies: (props: PatientAllergiesProps = {}) => ({
     key: "allergies",
     getPanelClassName: () => "ctw-pt-5",
-    display: () => "allergies",
+    display: () => (
+      <div className="ctw-space-x-2">
+        <UnreadAllergiesNotification />
+        <span className="ctw-capitalize">allergies</span>
+      </div>
```
WesleyKapow, line 58: Ping UX, we could make the badges absolutely positioned to the left, that way if when it toggles in/out, we won't get a content shift with the tabs.

WesleyKapow (approved):
LGTM, one comment/question about UX

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
eric-gunther: Gonna merge this now but will check w/ UX on making this change for the future. There's a chance they may go back to the right anyway.

WesleyKapow: Sounds good! Note, that bumping it to the right will have the same problem. Any tabs to the right of the badge will get shifted unless the badge takes zero space.

WesleyKapow: Zero space OR is always there and thus always take space (`opacity: 0` or somethin')

PR #891, opened 2023-06-30
CTW-1240 Preventing non-builder users from marking records as read in production

eric-gunther:
Checking the environment and user type when marking a record as read, if we're in production and it's a non-builder user then don't allow it. We could add this protection to dismissal but this feels a bit hack and it's really to protect against Zus users marking records as read because they simply clicked on them.

### DIFF
--- a/src/components/content/hooks/use-toggle-read.ts
+++ b/src/components/content/hooks/use-toggle-read.ts
@@ -27,6 +27,12 @@ export function useToggleRead(...queriesToInvalidate: string[]): UseToggleReadRe
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
   const handleToggleRead = useCallback(async (model: FHIRModel<fhir4.Resource>) => {
+    const requestContext = await getRequestContext();
+    // In production we want to avoid non-builder users from inadvertantly marking records as read
+    // In lower environments this is allowed for demos/testing
+    if (requestContext.env === "production" && !(requestContext.userType === "builder")) {
+      return;
+    }
     setIsLoading(true);
     await toggleRead(model, await getRequestContext());
     await Promise.all(
--- a/src/components/content/hooks/use-toggle-read.ts
+++ b/src/components/content/hooks/use-toggle-read.ts
@@ -30,7 +30,7 @@ export function useToggleRead(...queriesToInvalidate: string[]): UseToggleReadRe
     const requestContext = await getRequestContext();
     // In production we want to avoid non-builder users from inadvertantly marking records as read
     // In lower environments this is allowed for demos/testing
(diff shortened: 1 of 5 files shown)


src/components/content/hooks/use-toggle-read.ts
```diff
@@ -27,6 +27,12 @@ export function useToggleRead(...queriesToInvalidate: string[]): UseToggleReadRe
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
   const handleToggleRead = useCallback(async (model: FHIRModel<fhir4.Resource>) => {
+    const requestContext = await getRequestContext();
+    // In production we want to avoid non-builder users from inadvertantly marking records as read
+    // In lower environments this is allowed for demos/testing
+    if (requestContext.env === "production" && !(requestContext.userType === "builder")) {
+      return;
+    }
```
luthfib, line 35: Don't feel too strongly against this, but I do wonder if guarding a behavior against prod is something we want to introduce. I'd be more inclined to do it against userType or something similar so it is consistent against envs.

luthfib (dismissed):
Left a comment. Feel free to to decide best on what you think is best.

luthfib:
Also dont forget a changeset

src/components/content/hooks/use-toggle-read.ts
eric-gunther: My concern with disallowing this functionality for Zus users in sandbox/dev is that you then can't demo/test the feature in those environments unless you have a builder-scoped account which is a hassle to provision. So, while this does introduce some inconsistency across the envs, I think it's a pragmatic approach.

eric-gunther: Will check with the team on this during stand-up though.

luthfib (dismissed):

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -27,6 +27,12 @@ export function useToggleRead(...queriesToInvalidate: string[]): UseToggleReadRe
   const [isLoading, setIsLoading] = useState<boolean>(false);
 
   const handleToggleRead = useCallback(async (model: FHIRModel<fhir4.Resource>) => {
+    const requestContext = await getRequestContext();
+    // In production we want to avoid non-builder users from inadvertantly marking records as read
+    // In lower environments this is allowed for demos/testing
+    if (requestContext.env === "production" && !(requestContext.userType === "builder")) {
```
mrosata, line 33: Probably better to use `x !== y` than `!(x === y)`
```suggestion
    if (requestContext.env === "production" && requestContext.userType !== "builder") {
```

bryanjknight (approved):

PR #892, opened 2023-06-30
INFR-626 Create a new function in `urls` that constructs a service-name based URL

thinkbrown:
Also use it for routing requests to FQS via the new URLs.


pmadria3 (approved):
LGTM and tested locally as well

PR #895, opened 2023-06-30
ignore FQS graphql errors when there is also data returned

sarahmangels:
Fixes both https://zeushealth.atlassian.net/browse/CDEV-246 and https://zeushealth.atlassian.net/browse/CDEV-247

FQS can return both `errors` and `data` for a single request. 
This PR allows us to still use `data`, if it exists, even if there are `errors`.

 i got all the requests except these which are a bit different:
* https://github.com/zeus-health/ctw-component-library/blob/29b24fe333ed1c375f33e6dcdbb5f03f2c096f9f/src/components/content/resource/history.ts#L196
* https://github.com/zeus-health/ctw-component-library/blob/29b24fe333ed1c375f33e6dcdbb5f03f2c096f9f/src/components/content/resource/history.ts#L223
* https://github.com/zeus-health/ctw-component-library/blob/29b24fe333ed1c375f33e6dcdbb5f03f2c096f9f/src/services/fqs/long-poll-fqs.ts#L40
* https://github.com/zeus-health/ctw-component-library/blob/29b24fe333ed1c375f33e6dcdbb5f03f2c096f9f/src/services/fqs/queries/provenances.ts#L18

### DIFF
--- a/src/services/fqs/client.ts
+++ b/src/services/fqs/client.ts
@@ -35,7 +35,9 @@ export const fqsRequest = async <T>(client: GraphQLClient, query: string, variab
   const fhirData = graphQLToFHIR(data);
   if (errors) {
     if (data) {
-      Telemetry.logger.error(`Errors in FQS request:${errors}`);
+      errors.forEach((e) => {
+        Telemetry.logger.error("Error in FQS request", undefined, e);
+      });
       return { data: fhirData };
     }
     throw errors;
--- a/src/services/fqs/client.ts
+++ b/src/services/fqs/client.ts
@@ -1,4 +1,5 @@
-import { GraphQLClient } from "graphql-request";
+import { GraphQLClient, Variables } from "graphql-request";
+import { graphQLToFHIR } from "./graphql-to-fhir";
 import { getZusApiBaseUrl } from "@/api/urls";
 import { CTWRequestContext } from "@/components/core/providers/ctw-context";
 import { Env } from "@/components/core/providers/types";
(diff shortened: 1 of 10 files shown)


pmadria3 (approved):

pmadria3 (changes requested):
Wait so those requests do not work with the new changes or?

pmadria3 (approved):

sarahmangels:
> Wait so those requests do not work with the new changes or?

No they work the same as they always did I just didnt update them to be able to handle if the FQS query returned both errors and data. In that case those queries I didn't touch would just error out like they do right now.

We can update those ones in a separate PR

src/services/fqs/client.ts
```diff
@@ -21,12 +23,28 @@ export interface GenericConnection<T extends ResourceTypeString> {
...
 };
 
+export const fqsRequest = async <T>(client: GraphQLClient, query: string, variables: object) => {
+  const { data, errors } = await client.rawRequest<T>(query, variables as Variables);
+  const fhirData = graphQLToFHIR(data);
+  if (errors) {
+    if (data) {
+      errors.forEach((e) => {
+        Telemetry.logger.error("Error in FQS request", undefined, e);
+      });
```
WesleyKapow, line 40: Can we log just once for all of the errors?

WesleyKapow (approved):
LGTM

src/services/fqs/client.ts
sarahmangels: hmm I dont see a way to do that with Telemetry.logger.error but maybe im missing something

PR #901, opened 2023-07-03
fixing styling of notification icons

eric-gunther:
Right now the notification icon doesn't always display to the left, and vertically aligned, with the tab title.

We do still have an issue with the tab title content shifting when the notification icon goes away (ie. you've read the last resource). We could address this by having the icon always there and adjusting the opacity, but that looks pretty awkward with the icon on the left. Perhaps if/when we move the icon to the right we can take this approach.

image

### DIFF
--- a/.changeset/smart-ties-lie.md
+++ b/.changeset/smart-ties-lie.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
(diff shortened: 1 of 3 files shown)


Stephen-Tsen (approved):

luthfib (approved):
LGTM

PR #905, opened 2023-07-05
Fix: outside meds history caching

sarahmangels:
The outside medications history requests were not using a specific enough caching key and so the cache entry was being shared across all meds for a given patient. This PR adds the medication ID to the cache key so that each medication history has a separate cache entry.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -112,14 +112,7 @@ const components: DemoComponent[] = [
           hideRequestRecords: true,
         }}
         includePatientDemographicsForm={false}
-        resources={[
-          "allergies",
-          "conditions",
-          "conditions-outside",
-          "documents",
-          "immunizations",
-          "medications-outside",
-        ]}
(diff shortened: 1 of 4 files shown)


pmadria3 (dismissed):

src/App.tsx
```diff
@@ -112,7 +112,14 @@ const components: DemoComponent[] = [
...
         includePatientDemographicsForm={false}
-        resources={["allergies", "conditions", "conditions-outside", "documents", "immunizations"]}
+        resources={[
+          "allergies",
+          "conditions",
+          "conditions-outside",
+          "documents",
+          "immunizations",
+          "medications-outside",
+        ]}
```
pmadria3, line 122: You can remove this and then in your `env` file you can add `VITE_DEMO_APP_COMPONENTS="zap"` and it will add all the resources

pmadria3 (commented):
Small nit

bryanjknight (approved):

pmadria3 (approved):

PR #915, opened 2023-07-06
CDEV-58/auth fails error

WesleyKapow:
We now detect when we fail to fetch due to auth errors. This will happen, first via the patient fetch query, so we detect if that error'd with a 401 and show our new not authorized message.

Additionally:
* Remove tanstack queries default `retry: 3`. A failed request will fail immediately.
* Fixed `isLoading` logic, to properly take in to account if patient request is loading.

Screenshot 2023-07-06 at 11 05 58 AM
Screenshot 2023-07-06 at 11 06 19 AM


pmadria3 (approved):
LGTM

src/components/core/auth-error.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <NoAuthIcon className="ctw-h-16" />
+    </div>
+    <div className="ctw-space-y-3 ctw-text-center ctw-text-base">
+      <div className="ctw-text-xl ctw-font-medium">Not authorized</div>
+      <div>
+        Your account does not have access to this feature.
+        <br />
+        For further assistance, contact{" "}
+        <a href="mailto:support@zushealth.com" className="ctw-text-content-black">
+          support@zushealth.com
```
luthfib, line 15: Can we have this be something we pass in? I don' think we should have all this email in all EHR partners

luthfib (dismissed):
Left one comment

src/components/core/auth-error.tsx
WesleyKapow: Discussed in slack.. will leave as is for now and potentially make this something EHRs can overwrite.

PR #921, opened 2023-07-06
CTW-1283 Moving view FHIR to quick actions for allergies & immz

eric-gunther:
Modified allergies & immunizations so that the View FHIR button is in the row actions and not an always present column. Will make similar modifications to other tables as we bring them into alignment with the read/unread functionality.

I had considered doing this as generic functionality in `ResourceTable` but was not confident in how to do it without a somewhat major refactor. The `rowActions` that gets passed down is a `JSX.Element` that contains the contents of the row actions that get overlayed on the row content. I can't see a clean way of injecting the "View FHIR" button into that element. One option could've been to modify rowActions to take in an array of actions, and construct the row actions from that array, but then that takes away some of the flexibility that rowActions affords in its current state. I'm also not confident that it's a given that every `ResourceTable` should necessarily show the View FHIR button.

https://github.com/zeus-health/ctw-component-library/assets/97455910/1c21feed-ec98-4b2b-9fb4-126b0f9ea32a


WesleyKapow (changes requested):
ViewFHIR is also used for care-teams, documents, and timeline.

I strongly suggest trying to get this working within `ResourceTable` so that it is setup in a single place for all resources.

As mentioned in slack, you should be able to do something like `rowActions={<div className="ctw-flex ctw-space-x-2"><ViewFHIRThingy/>{rowActions}</div>}`

The `ctw-flex ctw-space-x-2` part is in there to align with how other row action buttons are setup and spaced.

src/components/core/view-fhir.tsx
```diff
@@ -20,13 +20,13 @@ export function useFHIRDrawer() {
...
   const openDrawer = useFHIRDrawer();
 
   return (
     <button
       type="button"
-      className="ctw-btn-primary ctw-text-sm"
+      className="ctw-btn-default"
       onClick={(event) => {
         // Prevents any rowClick handler from firing.
         event.stopPropagation();
```
WesleyKapow, line 32: This is not needed anymore as the parent of where row actions get inserted (in `table-rows.tsx` does this at the parent `td` level).

src/components/content/resource/resource-table.tsx
```diff
@@ -16,7 +17,7 @@ export type ResourceTableProps<T extends MinRecordItem> = {
   emptyMessage?: string | ReactElement;
   isLoading: boolean;
   onRowClick?: TableProps<T>["handleRowClick"];
-  rowActions?: TableProps<T>["RowActions"];
+  RowActions?: TableProps<T>["RowActions"];
```
WesleyKapow, line 20: Props should be camelCase (lower case first). When I setup `ResourceTable`, I made it `rowActions` but was refactoring a lot and didn't want to do more and switch the other cases of `RowActions` to `rowActions`.

src/components/content/resource/resource-table.tsx
```diff
@@ -49,6 +50,15 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
...
 
+  const rowActionsWithViewFHIR =
+    featureFlags?.enableViewFhirButton || RowActions
+      ? ({ record }: RowActionsProps<M>) => (
+          <div className="ctw-flex ctw-space-x-2">
+            {featureFlags?.enableViewFhirButton && <ViewFHIR resource={record.resource} />}
+            {RowActions && <RowActions record={record} />}
+          </div>
+        )
+      : undefined;
```
WesleyKapow, line 61: Beautiful!

WesleyKapow (approved):
Looks great! Only a couple of notes/nitpicks.

src/components/content/resource/resource-table.tsx
eric-gunther: Yeah I normally use lowercase for props but when I made this lowercase I was getting an error trying to reference rowActions as a component in JSX. I had noticed that the `Table` component has `RowActions` with a capital R and I figured it was because of this reason.
image

WesleyKapow: Oh right! That's why it was originally capitalized.. doh! OK carry on!

WesleyKapow (approved):

bryanjknight (approved):

PR #925, opened 2023-07-07
Removing padding from top of allergies/documents/immunizations/timelines

eric-gunther:
This `ctw-pt-5` class is adding padding to the top of some of the tabs in the ZAP and it does not appear necessary anymore. See video below for the discrepancy between conditions (where it looks good) and allergies (where it currently looks bad).

https://github.com/zeus-health/ctw-component-library/assets/97455910/0b75b3d7-ad1d-4672-9b8e-e66d6f6796a0

### DIFF
--- a/.changeset/odd-crabs-grin.md
+++ b/.changeset/odd-crabs-grin.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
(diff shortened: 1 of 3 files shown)


WesleyKapow (changes requested):
Looks like you've removed all uses of `getPanelClassName`, go ahead and remove it from the type as well!

WesleyKapow (approved):

PR #926, opened 2023-07-07
CTW-1286 Generic read/dismiss

eric-gunther:
Note this depends on (https://github.com/zeus-health/ctw-component-library/pull/921) going in first. 

- Remove read/dismiss logic from higher level components (like `PatientImmunizations`) and moving it into `ResourceTable`. Now you can enable read/dismiss simply by setting `enableDismissAndReadActions`.
- Adding read/dismiss for documents.


src/components/content/resource/resource-table.tsx
```diff
@@ -28,12 +35,13 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
   emptyMessage,
   isLoading,
   onRowClick,
-  rowActions,
+  RowActions,
   showTableHead,
-  boldUnreadRows,
+  enableDismissAndReadActions,
+  queryKeyToInvalidateOnReadDismiss,
```
eric-gunther, line 41: This feels a little awkward - if you enableDismissAndRead actions then you most likely need to have a specific query invalidated when dismiss/read happen. So you don't _need_ to also pass in a `queryKeyToInvalidateOnReadDismiss` but you probably should. Is there a more intelligent way to combine these into a single prop? Like an object called `dismissAndReadConfig` that contains an enable property and a query key to invalidate property? Or is this approach reasonable?

src/fhir/document.ts
```diff
@@ -1,22 +1,58 @@
...
+
+  const isLoading = patientDocumentsQuery.isLoading || basicsQuery.isLoading;
+  const isError = patientDocumentsQuery.isError || basicsQuery.isError;
+  const isFetching = patientDocumentsQuery.isFetching || basicsQuery.isFetching || !fqs.ready;
+  return {
+    isLoading,
+    isError,
+    isFetching,
+    data: documents,
+  };
```
eric-gunther, line 55: It definitely smells to keep re-writing basically this same thing to retrieve the Basic resources along with the actual clinical resources. However I think a lot of this is going to be up for refactoring soon when we officially pull the plug on querying ODS, so I'm of the mind that we can punt a cleaner refactoring of this data access layer until that time. FWIW documents is the last resource that I'm implementing this for (later we'll need it for labs...) so it's not like this pattern will grow out of hand too quickly.

src/components/content/document/unread-documents-notification.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [userBuilderId, setUserBuilderId] = useState("");
+
+  useEffect(() => {
+    async function load() {
+      const requestContext = await getRequestContext();
+      setUserBuilderId(requestContext.builderId);
+    }
+
+    void load();
+  }, [getRequestContext]);
```
WesleyKapow, line 19: Suggestion: pull this out into a `useBuilderId` hook. I'd setup a new `use-builder-id.ts` file in `core/providers`.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
```diff
@@ -93,7 +94,8 @@ export const zusAggregatedProfileTabs: ZusAggregatedProfileTabs = {
     key: "documents",
     getPanelClassName: () => "ctw-pt-5",
     display: () => (
-      <div className="ctw-space-x-1">
+      <div className="ctw-flex ctw-items-center ctw-space-x-2">
```
WesleyKapow, line 97: Just calling it out that you're increasing the space between the "documents" and the beta label. It use to be `space-x-1` and now it'll be `space-x-2`.

src/fhir/document.ts
WesleyKapow: Absolutely agreed. This is pretty gross. And agreed on punting for now. Probably good call to just wait until we kill ODS. That should reduce the noise in these areas and make it a little easier to clean this up.

src/components/content/document/unread-documents-notification.tsx
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from "react";
+import { useCTW } from "@/components/core/providers/use-ctw";
+import { UnreadNotification } from "@/components/core/unread-notification";
+import { usePatientDocuments } from "@/fhir/document";
+
+export const UnreadDocumentsNotification = () => {
```
WesleyKapow, line 6: I could see us benefiting from changing `UnreadNotification` to take an array of models as a prop and it do the "unreadOutside" logic. That way all of our resource notifications would simply be:

```ts
export const UnreadDocumentsNotification = () => {
  const query = usePatientDocuments();
  return <UnreadNotification data={query.data}/>
}
```

^ just a thought, not a suggestion for this PR.

src/components/content/resource/resource-table.tsx
```diff
@@ -49,6 +57,40 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
     void load();
   }, [getRequestContext]);
 
+  const queryKeysToInvalidate = [QUERY_KEY_BASIC];
+  if (queryKeyToInvalidateOnReadDismiss) {
+    queryKeysToInvalidate.push(queryKeyToInvalidateOnReadDismiss);
+  }
```
WesleyKapow, line 63: nit: Lodash's `compact` can simplify this a bit.

```ts
const queryKeysToInvalidate = compact([QUERY_KEY_BASIC, queryKeyToInvalidateOnReadDismiss])
```

src/components/content/resource/resource-table.tsx
```diff
@@ -49,6 +57,40 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
...
+  const { toggleRead } = useToggleRead(...queryKeysToInvalidate);
+
+  const onRowClickWithRead =
+    onRowClick || enableDismissAndReadActions
```
WesleyKapow, line 67: If there's no `onRowClick`, we probably don't want to mark it as read right? There wouldn't be any details getting opened in this case.

So this would be `onRowClick ?` without the `|| ...`

src/components/content/resource/resource-table.tsx
```diff
@@ -49,6 +57,40 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
...
+            toggleRead(record);
+          }
+          if (onRowClick) {
+            onRowClick(record);
+          }
+        }
+      : undefined;
+
+  const DismissAndReadActions =
+    enableDismissAndReadActions && data.length > 0
```
WesleyKapow, line 79: Why does `data.length > 0` matter?

src/components/content/resource/resource-table.tsx
```diff
@@ -76,19 +118,82 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
     >
       <Table
         getRowClassName={(record) => ({
-          "ctw-tr-dismissed": record.isDismissed,
+          "ctw-tr-dismissed": enableDismissAndReadActions && record.isDismissed,
```
WesleyKapow, line 121: Why add the `enableDismissAndReadActions` guard here? Shouldn't we display it as dismissed if it is, no matter the `enableDismissAndReadActions` bool?

src/components/content/resource/resource-table.tsx
```diff
@@ -76,19 +118,82 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
...
           "ctw-tr-unread":
-            boldUnreadRows && !record.ownedByBuilder(userBuilderId) && !record.isRead,
+            enableDismissAndReadActions && !record.ownedByBuilder(userBuilderId) && !record.isRead,
```
WesleyKapow, line 123: This one makes more sense. Only render as unread of you can mark it as read.

src/components/content/resource/resource-table.tsx
```diff
@@ -76,19 +118,82 @@ export const ResourceTable = <T extends fhir4.Resource, M extends FHIRModel<T>>(
...
+
+const getDismissAndReadActions =
+  (userBuilderId: string, queryKeyToInvalidate?: string) =>
+  ({ record }: RowActionsProps<FHIRModel<fhir4.Resource>>) => {
+    const { t } = useBaseTranslations();
+
+    const queryKeysToInvalidate = [QUERY_KEY_BASIC];
+    if (queryKeyToInvalidate) {
+      queryKeysToInvalidate.push(queryKeyToInvalidate);
+    }
```
WesleyKapow, line 146: This ends up being the same as `queryKeysToInvalidate` from above.. should you pass that in instead of `queryKeyToInvalidate`?

src/components/content/resource/resource-table.tsx
WesleyKapow: You're not currently using this right? When would we? I think we only need to invalidate the basics query. Any hooks that depend on basics would automatically get updated when `useBasics` gets updated. So I think this can be removed unless I'm missing something.

WesleyKapow (changes requested):
Handle of nits and questions, don't think there was anything major.

Love the new place within `ResourceTable` for us to put generic actions!

src/components/content/resource/resource-table.tsx
eric-gunther: Ha that's a great point, derp

WesleyKapow (dismissed):
LGTM

luthfib (approved):

PR #942, opened 2023-07-10
CTW-1281 New PatientConditionsAll component

eric-gunther:
Added new `PatientConditionsAll` component that, moving forward, will be our standard component for conditions. `PatientConditions`,`PatientConditionsOutside` and `PatientConditionsProfile` should be considered deprecated and will be phased out in coordination with consumers.

The new component has read/dismiss functionality, full CRUD operations for builder owned conditions, and Add functionality for outside conditions. The data displayed for conditions has largely remained the same except for some simplification to the columns that was approved by product/UX during design sessions.

https://github.com/zeus-health/ctw-component-library/assets/97455910/0dc5668b-4f52-4da7-bab6-0d69822558c3


src/components/content/conditions/helpers/columns.tsx
```diff
@@ -34,8 +37,59 @@ export const patientConditionsColumns: TableColumn<ConditionModel>[] = [
...
+      <div>
+        <div className="ctw-flow-root group-hover:ctw-underline">
+          {capitalize(condition.display)}
+          <span className="ctw-float-right">
+            {condition.ownedByBuilder(builderId) ? (
+              <FontAwesomeIcon className="ctw-text-content-light" icon={faCircleCheck} />
+            ) : (
+              <></>
+            )}
+          </span>
```
mrosata, line 77: Can use `null` instead of a document fragment. Can probably omit rendering that empty span as well unless it's needed for some layout reason
```suggestion
          {condition.ownedByBuilder(builderId) ? (
            <span className="ctw-float-right">
              <FontAwesomeIcon className="ctw-text-content-light" icon={faCircleCheck} />
            </span>) : null}
```

src/components/content/conditions/patient-conditions-all.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const openDetails = useConditionDetailsDrawer({
+    canRemove: !readOnly,
+    canEdit: !readOnly,
+  });
+
+  const rowActions = useMemo(() => getRowActions(userBuilderId), [userBuilderId]);
+
+  const action = !readOnly && (
+    <button type="button" className="ctw-btn-primary" onClick={() => showAddConditionForm()}>
```
mrosata, line 57:
```suggestion
    <button type="button" className="ctw-btn-primary" onClick={showAddConditionForm}>
```

src/components/content/conditions/patient-conditions-all.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { withErrorBoundary } from "@/components/core/error-boundary";
+import { useCTW } from "@/components/core/providers/use-ctw";
+import { RowActionsProps } from "@/components/core/table/table";
+import { ConditionModel } from "@/fhir/models";
+import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
+import { useBaseTranslations } from "@/i18n";
+import { usePatientConditionsAll } from "@/services/conditions";
+
+export type PatientConditionsAllProps = {
+  className?: string;
```
mrosata, line 23: Since this gets passed into `cx` it can use type `cx.Argument` which includes string but would also allow us to pass in anything valid to use with `cx`
```suggestion
  className?: cx.Argument;
```

src/components/content/conditions/patient-conditions-all.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  useEffect(() => {
+    async function load() {
+      const requestContext = await getRequestContext();
+      setUserBuilderId(requestContext.builderId);
+    }
+
+    void load();
+  }, [getRequestContext]);
+
```
mrosata, line 48: We should make this a hook. Believe I saw this pattern used recently somewhere else. Be good to shove it into a reuseable hook.

src/components/content/conditions/patient-conditions-all.tsx
eric-gunther: That change gives me this error `Type '(referenceCondition?: ConditionModel) => void' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.`. Think it may need to stay the way it is.

src/components/content/conditions/helpers/columns.tsx
```diff
@@ -17,25 +16,13 @@ export const patientConditionsColumns: TableColumn<ConditionModel>[] = [
...
-            Last Updated:{" "}
-            {compact([
-              condition.recordedDate,
-              condition.recorder ? `(${condition.recorder})` : "",
-            ]).join(" ")}
-          </div>
-        </div>
+      <div>
+        <div>{condition.recordedDate}</div>
+        <div>{condition.recorder}</div>
```
eric-gunther, line 23: This is an intentional change to the existing components approved by product & UX.

src/components/content/conditions/patient-conditions-all.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { RowActionsProps } from "@/components/core/table/table";
+import { ConditionModel } from "@/fhir/models";
+import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
+import { useBaseTranslations } from "@/i18n";
+import { usePatientConditionsAll } from "@/services/conditions";
+
+export type PatientConditionsAllProps = {
+  className?: string;
+  readOnly?: boolean;
+  onlyAllowAddOutsideConditions?: boolean;
```
eric-gunther, line 25: I'm finding these props to be a bit awkward. Here's what I'm intending. `readOnly` means you can't do any CRUD actions against any resources (other than read/dismiss which feel reasonable to leave uncontrolled by a prop). In the case of EHR integrations we don't actually want to give the ability to manage full CRUD on a resource, we only want to expose the ability to add an outside condition. After that you have to do all CRUD through the EHR. Thus the `onlyAllowOutsideConditions` prop.

src/fhir/action-helper.ts
```diff
@@ -54,7 +54,12 @@ export async function createOrEditFhirResource(
     // Wait for FQS to have the latest version of the resource.
     // This way callers can safely refetch/invalidateQueries
     // and be sure to get fresh data from FQS.
-    if (isFHIRDomainResource(response) && response.id && response.meta?.lastUpdated) {
+    if (
+      response.resourceType !== "Basic" && // explicit carve-out for Basics until we are querying FQS for them
+      isFHIRDomainResource(response) &&
+      response.id &&
+      response.meta?.lastUpdated
+    ) {
```
eric-gunther, line 62: This is an important response to the change in longPollFQS further below. FQS requests for Basic resources is not supported for us yet so we should not rely on the longpoll for them.

src/services/fqs/long-poll-fqs.ts
```diff
@@ -34,16 +34,18 @@ export async function longPollFQS(
...
       const data = await graphClient.request(query);
-      currentTimestamp = Date.parse(String(get(data, "Condition.meta.lastUpdated")));
+      currentTimestamp = Date.parse(String(get(data, `${resourceType}.meta.lastUpdated`)));
     } catch (err) {
       // Ignore errors and retry.
       // This can happen if we just created a new resource that isn't in FQS yet.
     }
-  } while (currentTimestamp < targetTimestamp && Date.now() - startTime < TIMEOUT_MS);
+  } while (
+    (Number.isNaN(currentTimestamp) || currentTimestamp < targetTimestamp) &&
```
eric-gunther, line 46: The long poll was not working on net new resources where the first request returns no data and `currentTimestamp` gets set to `NaN` which is not less than `targetTimestamp`.

src/services/conditions.ts
```diff
@@ -148,33 +194,29 @@ export const deleteCondition = async (
...
+      const isMatch = outsideCondition.knownCodingsMatch(builderCondition);
+      const isEnteredInError = builderCondition.verificationStatus === "entered-in-error";
 
       const isOlder =
         !otherRecordedDate || (patientRecordedDate && otherRecordedDate <= patientRecordedDate);
-      const hasSameStatus = otherCondition.clinicalStatus === patientCondition.clinicalStatus;
+      const hasSameStatus = outsideCondition.clinicalStatus === builderCondition.clinicalStatus;
 
       return isMatch && !isEnteredInError && (isOlder || hasSameStatus);
     });
```
eric-gunther, line 222: This is all renaming and adjusting the comments - trying to be clearer with our usage of the terms "summary", "builder" and "outside".  I think we're often confusing summary with outside in the code - not all summary records represent outside data (hence the purpose of this function).

mrosata (approved):

mrosata (approved):

PR #948, opened 2023-07-11
CDEV-59/diagnostics component

WesleyKapow:
CDEV-59

**NOTE TO REVIEWER: Best to review each commit** (also note: re-ordering within tabs did not in-fact re-order the tabs in ZAP.. this is controlled by the consumer.)

Screenshot 2023-07-11 at 3 01 06 PM


src/components/content/diagnostic-reports/helpers/columns.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { TableColumn } from "@/components/core/table/table-helpers";
+import { DiagnosticReportModel } from "@/fhir/models";
+import { capitalize } from "@/utils/nodash";
+
+export const patientDiagnosticReportsColumns = (
+  builderId: string
+): TableColumn<DiagnosticReportModel>[] => [
+  {
+    title: "Diagnostic",
+    render: (diagnosticReport) => (
```
eric-gunther, line 12: Think this could use a `minWidth` or `widthPercent`. Defaults to wrapping for me even when the details/performer is blank.

src/components/content/diagnostic-reports/helpers/columns.tsx
```diff
@@ -0,0 +1,35 @@
...
+          {capitalize(diagnosticReport.displayName)}
+          <span className="ctw-float-right">
+            {diagnosticReport.ownedByBuilder(builderId) ? (
+              <FontAwesomeIcon className="ctw-text-content-light" icon={faCircleCheck} />
+            ) : (
+              <></>
+            )}
+          </span>
+        </div>
+      </div>
```
eric-gunther, line 24: FYI I made a reusable component for this boilerplate "show a title, optional subtitle, and first party check mark" stuff. It's in my conditions PR which this will probably beat in so I'll update this later, just giving you the heads up.

src/fhir/basic.ts
```diff
@@ -80,6 +82,42 @@ export function useBasic(fqs: FQSFeatureToggle) {
   );
 }
 
+export function useIncludeBasics<R extends fhir4.Resource, T extends FHIRModel<R>>(
```
eric-gunther, line 85: I'll update conditions/meds to use this while I'm in there as part of my PR.

eric-gunther (approved):
Amazing template for what is required in making a new component. Only have non-blocking feedback.

Since this is obviously a breaking change by removing the observations components perhaps this justifies a major changeset? Practically I don't think it makes much of a difference right now, so I guess this is just a reminder to add a changeset.

pmadria3:
Need changeset

src/components/content/diagnostic-reports/helpers/columns.tsx
WesleyKapow: Awesome!

src/fhir/basic.ts
WesleyKapow: NOTE: I looked at conditions.. it does have some added complexity. Could probably be cleaned up a bit but I punted on refactoring it for now.

eric-gunther (approved):

PR #954, opened 2023-07-12
CTW-1279 New PatientMedicationsAll component

eric-gunther:
Added new `PatientMedicationsAll` component which will replace the existing medications components in standalone. The existing components should be considered deprecated now but we will need to work with individual consumers to migrate off them.

Note that medications are lacking all CRUD operations like conditions have - you cannot add a new medication and you cannot edit/remove builder-owned medications. All you can do is add a medication from the outside list to your builder. Getting to parity with conditions is not in scope for this effort and that is approved by product/UX.

https://github.com/zeus-health/ctw-component-library/assets/97455910/5956427e-a53a-4d50-9c4e-f876562fb21c


src/fhir/medications.ts
```diff
@@ -505,9 +505,9 @@ export async function getActiveMedicationsFQS(
     const nodes = data.MedicationStatementConnection.edges.map((x) => x.node);
     const models = setupMedicationStatementModelsWithFQS(nodes);
     if (models.length === 0) {
-      Telemetry.countMetric("req.count.active_medications.none", 1, ["fqs"]);
+      Telemetry.countMetric("req.count.summary_medications.none", 1, ["fqs"]);
     }
-    Telemetry.histogramMetric("req.count.active_medications", models.length, ["fqs"]);
+    Telemetry.histogramMetric("req.count.summary_medications", models.length, ["fqs"]);
```
eric-gunther, line 510: This will require updating our dashboards, which I can own once this is merged, but we gotta remove the word "active" since we're not actually limiting to active medications anymore.

mrosata (approved):

PR #955, opened 2023-07-12
CTW-1294 General unread records notification

eric-gunther:
Adds a new component `UnreadRecordsNotification` that displays the notification if there are any unread records across multiple resource types.

There is some duplication across this component and the existing `UnreadConditionsNotification`, `UnreadMedicationsNotification`, etc. We could get rid of those and just use this one but I don't think the proliferation of these different unread notifications is all that bad or confusing.

https://github.com/zeus-health/ctw-component-library/assets/97455910/06648a7a-35f0-49aa-b042-15d6e72115fd


src/components/core/unread-notification-icon.tsx
```diff
@@ -1,11 +1,18 @@
...
   className = "ctw-fill-notification-icon",
-}: UnreadNotificationIconProps) => (
-  <svg height={10} className={className} viewBox="0 0 10 10" xmlns="http://www.w3.org/2000/svg">
-    <circle cx="5" cy="5" r="5" />
-  </svg>
-);
+  text,
+}: UnreadNotificationIconProps) =>
+  text ? (
+    <Badge text={text} className={className} color="notification" />
```
eric-gunther, line 13: Adding a text option for the unread icon so that consumers can display a badge with some text like "New Data" instead of just the red dot.

mrosata (approved):

PR #962, opened 2023-07-13
CDEV-204 patient history empty states

sarahmangels:
expand table empty states to include request records button and show spinner while request is in progress

https://zeushealth.atlassian.net/browse/CDEV-204

image

image


src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -172,3 +177,51 @@ export function usePatientHistoryDetails() {
...
+      if (patient.data && !isPolling) {
+        setIsPolling(true);
+        const requestContext = await getRequestContext();
+        await pollPatientHistory(requestContext, patient.data.id);
+        setIsFetching(false);
+      }
+    }
+    void poll();
+  }, [getRequestContext, patient, isPolling]);
+  return isFetching;
```
eric-gunther, line 197: Should there be more uses of `setIsPolling` and `setIsFetching` here?

src/components/content/patient-history/use-patient-history.tsx
```diff
@@ -172,3 +177,51 @@ export function usePatientHistoryDetails() {
...
+      const response = (await getBuilderRefreshHistoryMessages({
+        requestContext,
+        patientId,
+      })) as PatientHistoryJobResponse;
+
+      isDone = response.data[0]?.attributes.providers.every(
+        (provider) =>
+          provider.status === "done" ||
+          provider.status === "error" ||
+          provider.status === "done_with_errors"
```
eric-gunther, line 214: We reference these statuses in a number of places now (eg. `patient-history-message-status.tsx`, `patient-history-table.tsx`), think we should have em as constants.

src/components/core/icons/refresh.tsx
```diff
@@ -0,0 +1,21 @@
+export const RefreshIcon = ({ className }: { className?: string }) => (
```
eric-gunther, line 1: Is there not an applicable font awesome icon for this?

eric-gunther (changes requested):

src/components/core/icons/refresh.tsx
sarahmangels: no :( Tai said the closest one in font awesome looked too much like an undo icon so she modified it

src/components/content/patient-history/use-patient-history.tsx
sarahmangels: setIsPolling is being used to block this hook from being executed multiple times at once - so there can only be 1 poll going on at a time, so it makes sense to only call it once. 

setIsFetching is used to mark when the polling is complete so we only set it at the end once the API says the job is complete!

src/components/content/patient-history/use-patient-history.tsx
sarahmangels: hm wait i think it sort of is already?? its comparing provider.status which is type PatientRefreshHistoryMessageStatus (https://github.com/zeus-health/ctw-component-library/blob/5d8b23156ed5e9daf5bcf88867f4ef903e5dcd63/src/services/patient-history/patient-history-types.ts#L1) so it wont let me change the strings here to anything outside of that type

eric-gunther: Ah, well that seems fair enough. Won't block on it.

eric-gunther (dismissed):

eric-gunther (approved):

PR #979, opened 2023-07-17
Setting up standalone ZAP that can be embedded in iframe

eric-gunther:
This PR sets up some routes in our test app so that there is a `/debug` route serving our existing test app and a `/ZAP` route that serves a standalone version of the ZAP. In an upcoming PR we will set up some infra to formally host this somewhere, say https://zap.zusapi.com/ZAP, and consumers will be able to load that site in an iframe. Then, using the `postMessage` API, consumers will need to pass a config object containing props for `CTWProvider`, `PatientProvider` and `ZusAggregatedProfile`. Another upcoming PR will also set up a new `ZusAggregatedProfileIframe` component that loads the iframe and handles passing all of that config for you.

https://github.com/zeus-health/ctw-component-library/assets/97455910/60fad556-34bd-4c2f-9bbe-00c4cf9431a4


src/App.tsx
```diff
@@ -1,188 +1,14 @@
...
+import { BrowserRouter, Routes, Route } from "react-router-dom";
+import Debug from "./Debug";
+import ZAP from "./ZAP";
+
+export default function App() {
+  return (
+    <BrowserRouter>
+      <Routes>
+        <Route path="/debug" element={<Debug />} />
+        <Route path="/ZAP" element={<ZAP />} />
```
eric-gunther, line 10: Probably want to make this something like `ZAP-1` so that we can version the URL and eventually host a `ZAP-2` if we need to make non-backwards-compatible changes.

src/ZAP.tsx
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    window.addEventListener("message", handler);
+
+    window.parent.postMessage(
+      {
+        type: "zus-ready",
+      },
+      "*"
+    );
```
eric-gunther, line 39: This is necessary so that the hosting app knows when it can pass the config via `postMessage`, otherwise it may do so before this app has added the event listener to handle the config message.

src/ZAP.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <CTWProvider {...zapConfig.CTWProviderProps}>
+        <PatientProvider {...zapConfig.PatientProviderProps}>
+          <ZusAggregatedProfile {...zapConfig.ZusAggregatedProfileProps} />
+        </PatientProvider>
+      </CTWProvider>
+    );
+  }
+
+  return <div>Waiting...</div>;
```
eric-gunther, line 55: Probably want some better content when it's in this waiting state.

zapIframeTest.html
```diff
@@ -0,0 +1,51 @@
...
+        
+        window.addEventListener('message', ({ data }) => {
+            if (data.type === 'zus-ready') {
+                document
+                    .getElementById('zap')
+                    .contentWindow.postMessage(
+                        {
+                            type: "ZusAggregatedProfileIframeConfig",
+                            config: {
+                                CTWProviderProps: {
```
eric-gunther, line 15: This won't work for things like `onResourceSave`, will need to consider that further.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -75,7 +75,16 @@ const ZusAggregatedProfileComponent = ({
   medicationsOutsideProps,
   medicationsAllProps,
   timelineProps,
-  resources,
+  resources = [
```
eric-gunther, line 78: Seems reasonable to me that we'd have some default opinion about what resources to show in the ZAP.

PR #991, opened 2023-07-18
CTW-1293 Setting up ZusAggregatedProfileIframe component

eric-gunther:
First iteration on `ZusAggregatedProfileIframe` that takes in the same props as `ZusAggregatedProfile` along with a few additional/optional props to specify the iframe ID/height/width. This loads the hosted ZAP in an iframe with the same config already setup through `CTWProvider` and `PatientProvider`, including themes and internationalization. 

https://github.com/zeus-health/ctw-component-library/assets/97455910/cc9c9388-e531-4859-826b-d5f3916178e5

### DIFF
--- a/.storybook/preview.jsx
+++ b/.storybook/preview.jsx
@@ -71,6 +71,11 @@ export const parameters = {
   msw: {
     handlers: {
       metrics: [rest.post("*/report/metric", (_, res, ctx) => res(ctx.status(200), ctx.json({})))],
+      unleashMetrics: [
+        rest.post("https://unleash-proxy-*.zusapi.com/proxy/client/metrics", async (_, res, ctx) =>
+          res(ctx.status(200), ctx.json({}))
+        ),
+      ],
     },
   },
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,7 @@ import {
   UnreadRecordsNotification,
   ZusAggregatedProfileIframe,
 } from ".";
+import { Theme } from "@/styles/tailwind.theme";
 
 const {
   VITE_AUTH0_AUDIENCE,
@@ -50,7 +51,7 @@ type DemoComponent = {
 const componentsToRender = VITE_DEMO_APP_COMPONENTS.split(",").map(trim);
 
 // Feel free to play with this theme object
-const theme = {
+const theme: Theme = {
   colors: {
     notification: {
       main: "#a855f7",
@@ -117,15 +118,31 @@ const components: DemoComponent[] = [
     name: "zap",
     render: () => {
       return (
-        <ZusAggregatedProfileIframe
-          height="600"
+        <ZusAggregatedProfile
+          title="ZAP"
           conditionsOutsideProps={{
             hideRequestRecords: true,
           }}
           conditionsAllProps={{ onlyAllowAddOutsideConditions: true }}
           includePatientDemographicsForm={false}
-          title="ZAP"
         />
+      );
+    },
+  },
+  {
+    name: "iframe-zap",
+    render: () => {
+      return (
+        <div style={{ height: "600px" }}>
+          <ZusAggregatedProfileIframe
+            title="IFrame ZAP"
+            conditionsOutsideProps={{
+              hideRequestRecords: true,
+            }}
+            conditionsAllProps={{ onlyAllowAddOutsideConditions: true }}
+            includePatientDemographicsForm={false}
+          />
+        </div>
       );
     },
   },
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,6 +22,7 @@ import {
   ZAPResourceName,
   ZusAggregatedProfile,
   UnreadRecordsNotification,
+  ZusAggregatedProfileIframe,
 } from ".";
 
 const {
@@ -118,7 +119,7 @@ const components: DemoComponent[] = [
       return (
         <div>
           <UnreadRecordsNotification text="Patient has unread records" />
-          <ZusAggregatedProfile
+          <ZusAggregatedProfileIframe
             conditionsOutsideProps={{
               hideRequestRecords: true,
             }}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -64,8 +64,8 @@ const theme: Theme = {
 const locals = {
   en: {
     glossary: {
-      condition_one: "problem",
-      condition_other: "problems",
+      condition_one: "condition",
+      condition_other: "conditions",
     },
     main: {
       "zap.tabs.conditionsOutside": "conditions",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,8 +63,8 @@ const theme = {
 const locals = {
   en: {
     glossary: {
-      condition_one: "condition",
-      condition_other: "conditions",
+      condition_one: "problem",
+      condition_other: "problems",
     },
     main: {
       "zap.tabs.conditionsOutside": "conditions",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -117,17 +117,15 @@ const components: DemoComponent[] = [
     name: "zap",
     render: () => {
       return (
-        <div>
-          <UnreadRecordsNotification text="Patient has unread records" />
-          <ZusAggregatedProfileIframe
-            conditionsOutsideProps={{
-              hideRequestRecords: true,
-            }}
-            conditionsAllProps={{ onlyAllowAddOutsideConditions: true }}
-            includePatientDemographicsForm={false}
-            title="ZAP"
-          />
-        </div>
+        <ZusAggregatedProfileIframe
+          height="600"
+          conditionsOutsideProps={{
+            hideRequestRecords: true,
+          }}
+          conditionsAllProps={{ onlyAllowAddOutsideConditions: true }}
+          includePatientDemographicsForm={false}
+          title="ZAP"
+        />
       );
     },
   },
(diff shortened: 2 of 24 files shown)


src/App.tsx
```diff
@@ -116,17 +117,15 @@ const components: DemoComponent[] = [
...
-        </div>
+        <ZusAggregatedProfileIframe
+          height="600"
+          conditionsOutsideProps={{
+            hideRequestRecords: true,
+          }}
+          conditionsAllProps={{ onlyAllowAddOutsideConditions: true }}
+          includePatientDemographicsForm={false}
+          title="ZAP"
+        />
```
eric-gunther, line 128: I can remove all the changes in App.tsx before merging, was just using this for testing.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,128 @@
...
+        env: requestContext.env,
+        builderId: requestContext.builderId,
+        ehr: requestContext.ehr,
+        featureFlags,
+        enableTelemetry: requestContext.enableTelemetry,
+        headers: requestContext.headers,
+        locals: requestContext.locals,
+        theme: requestContext.theme,
+        // TODO
+        // onResourceSave
```
eric-gunther, line 69: These callbacks will be addressed in a follow-up PR because they are going to introduce a fair amount of complexity and I'd like to make some more incremental progress here. Can put a better comment to express that.

The challenge at hand is that we can't simply pass the callback function to the ZAP running in the iframe via `postMessage`. Instead we need to set up another message event listener that executes the callback. This will require some work on the hosted ZAP to ensure that we expose the callbacks via a message.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,128 @@
...
+      } as CTWProviderProps;
+
+      const patientProviderProps = {
+        patientID: patient.data?.UPID,
+        systemURL: SYSTEM_ZUS_UNIVERSAL_ID,
```
eric-gunther, line 74: Upon further thought, I should probably change this to use the patient ID and system that were originally passed to patient provider, but I wonder if using the UPID has any side effects I'm not thinking of?

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,128 @@
...
+  ...zapProps
+}: ZusAggregatedProfileIframeProps) => {
+  const [hostedZapReady, setHostedZapReady] = useState(false);
+  const [sentZapConfig, setSentZapConfig] = useState(false);
+  const { getRequestContext, featureFlags } = useCTW();
+  const patient = usePatient();
+
+  const [zapURL, setZAPURL] = useState<string | undefined>(undefined);
+
+  useEffect(() => {
```
eric-gunther, line 31: TODO - put some more comments explaining the purpose of each of the `useEffect` calls: 1) set up listener to know when iframe ZAP is ready, 2) set up ZAP URL, 3) pass config to iframe ZAP.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }, []);
+
+  useEffect(() => {
+    async function load() {
+      const requestContext = await getRequestContext();
+      if (requestContext.env === "production") {
+        setZAPURL("https://zap.zusapi.com");
+      } else {
+        setZAPURL(`https://zap.${requestContext.env}.zusapi.com`);
+      }
```
mrosata, line 46: These might be better as `https://zap.zushealth.com` and `https://zap.<env>.zushealth.com` _(assuming they have been setup with that URL, if not we may want to consider updating them to use `zushealth.com` before it's a more difficult migration process to do so)_.

Also, less important, should we keep these urls in the [urls.ts](https://github.com/zeus-health/ctw-component-library/blob/main/src/api/urls.ts) file?

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,132 @@
...
+    async function load() {
+      const requestContext = await getRequestContext();
+      // setZAPURL(getZusServiceUrl(requestContext.env, "zap"));
+      setZAPURL("http://localhost:5173");
+
+      window.addEventListener("message", ({ data }) => {
+        if (data.type === ZusAggregatedProfileOnResourceSaveMessageType) {
+          requestContext.onResourceSave(data.resource, data.action, data.error);
+        }
+      });
```
eric-gunther, line 50: This is how we'll handle the callbacks now, listening for a message from the iframe and using it to execute the callback.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,132 @@
...
+    });
+  }, []);
+
+  useEffect(() => {
+    async function load() {
+      const requestContext = await getRequestContext();
+      // setZAPURL(getZusServiceUrl(requestContext.env, "zap"));
+      setZAPURL("http://localhost:5173");
+
+      window.addEventListener("message", ({ data }) => {
```
WesleyKapow, line 46: Whenever you see `addEventListener`, you want to see either a `{once: true}` or a corresponding `removeEventListener`.. otherwise we'll end up accumulating event listeners. In this case we'd have a real problem if this effect ran twice as we'd end up with two listeners that would call `onResourceSave`. The use effect gets invoked whenever the dependencies change, so if `getRequestContext` got changed.

To solve this, `useEffect`'s callback can return a cleanup function. See [example here](https://www.pluralsight.com/guides/how-to-cleanup-event-listeners-react#module-manualeventlistenercleanup-anchor).

--- 

From docs, https://react.dev/reference/react/useEffect
> React will first run the cleanup function (if you provided it) with the old values, and then run your setup function with the new values. After your component is removed from the DOM, React will run your cleanup function.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+      const patientProviderProps = {
+        patientID: patient.data?.UPID,
+        systemURL: SYSTEM_ZUS_UNIVERSAL_ID,
+        // TODO
+        // onPatientSave
+        // onResourceSave
+      } as PatientProviderProps;
+
+      iframe.contentWindow?.postMessage(
```
WesleyKapow, line 84: Presumably, iframe correctly handles/updates when receiving a second or third config message?

src/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe.tsx
```diff
@@ -0,0 +1,132 @@
...
+            ZusAggregatedProfileProps: zapProps,
+          },
+        },
+        zapURL as string
+      );
+
+      setSentZapConfig(true);
+    }
+
+    if (hostedZapReady && !sentZapConfig && patient.data?.UPID) {
```
WesleyKapow, line 99: `!sentZapConfig` means we will not update the config if any of the dependencies change. So for example, if auth token expired and go refreshed, I think the iframe would still be on the stale/expired auth token.

.storybook/preview.jsx
```diff
@@ -71,6 +71,11 @@ export const parameters = {
   msw: {
     handlers: {
       metrics: [rest.post("*/report/metric", (_, res, ctx) => res(ctx.status(200), ctx.json({})))],
+      unleashMetrics: [
+        rest.post("https://unleash-proxy-*.zusapi.com/proxy/client/metrics", async (_, res, ctx) =>
+          res(ctx.status(200), ctx.json({}))
+        ),
+      ],
```
mrosata, line 78: Added this because it's a request that isn't covered in storybook, this will just make that a little less noisy

src/components/content/observations/helpers/columns.tsx
```diff
@@ -96,9 +96,7 @@ export const ObservationsColumns = (): TableColumn<ObservationModel>[] => {
                 className="ctw-btn-clear"
                 type="button"
                 onClick={() => {
-                  Telemetry.trackInteraction(
-                    isTrendsShown ? "CollapsedTrendNotShown" : "ExpandedTrendShown"
-                  );
+                  Telemetry.trackInteraction(isTrendsShown ? "collapse_trends" : "expand_trends");
```
mrosata, line 99: Updated this because `ctw.action.collapsed_trend_not_shown` and `ctw.action.expanded_trend_shown` make less sense then `ctw.action.collapse_trends` and `ctw.action.expand_trends` which is the action the user is actually performing

src/utils/telemetry.ts
```diff
@@ -105,35 +102,39 @@ export class Telemetry {
...
-      }
-    });
-    body?.addEventListener("focusin", (event) => {
-      const { target, isTrusted } = event;
-      if (isTrusted && target instanceof HTMLElement) {
-        this.processHTMLEvent(target, "zusTelemetryFocus");
-      }
-    });
+    // We need to ensure that this will run in browser context, not just server.
+    if (!listenersHaveBeenAdded && body) {
```
mrosata, line 106: Added this conditional wrapping because with the move to Telemetry provider I believe we may be initializing telemetry on the server-side and setting these event listeners up before `document.body` exists. @pmadria3 had noticed that interactions which didn't call `Telemetry.trackInteraction` directly (meaning they depended on these browser event handlers) were not being sent in ctw standalone. This is likely why. Since we had to touch the telemetry file anyways to I figured I'd add this fix too.

pmadria3 (approved):

PR #994, opened 2023-07-18
Improve responsiveness of ZAP tabs

WesleyKapow:
CDEV-278
CDEV-273

Features:
* Takes into account any topRightContent like "Request Records"
* Reserves space for the "More" menu unless we're calculating the space for the last tab as we'd then hide the more menu.
* Only show more menu when breakpoints.sm and when doing so, show the tab name instead of "More".

Additionally:
* Setup ListBox to optionally take a `ref` (used to determine size of more button)
* Added optional `selectedIndex` to ListBox so that we can control the selected item from the outside.
* Tweaked unread notification bullet to be absolutely positioned within ZAP tabs and more menu. This prevents content shifting when notification comes in or out.

Jul-18-2023 15-34-10

### DIFF
--- a/.changeset/strange-bobcats-agree.md
+++ b/.changeset/strange-bobcats-agree.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Improve responsiveness of ZAP tabs to collapse overflowed tabs into a more menu.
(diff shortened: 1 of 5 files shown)


pmadria3 (approved):

patrickvalle (dismissed):
lgtm, impressive

eric-gunther (approved):

PR #1011, opened 2023-07-20
CTW-1114/Track active session and generic drawer open

mrosata:
Send a metric for active session and for opening a drawer.
When an interaction is made (currently just opening a drawer) we track a session in localStorage. A new session doesn't begin until an hour after no other interaction is made by the same user. We send a `ctw.active_session` metric count to DataDog.

This doesn't cover the whole of CTW-1114, it doesn't do any tracking in DynamoDB so we can't yet track user retention, only user activity totals. I want to get this merged a bit early to allow others to work on some interaction metrics and while I spike on the best way to complete the rest of CTW-1114 ticket in a way that supports better long term outcomes.

### DIFF
--- a/src/utils/telemetry.ts
+++ b/src/utils/telemetry.ts
@@ -1,5 +1,6 @@
 import { datadogLogs } from "@datadog/browser-logs";
 import jwtDecode from "jwt-decode";
+import { Session } from "./session";
 import packageJson from "../../package.json";
 import { getMetricsBaseUrl } from "@/api/urls";
 import { FhirError, fhirErrorResponse } from "@/fhir/errors";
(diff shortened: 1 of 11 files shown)


src/utils/telemetry.ts
```diff
@@ -182,18 +186,11 @@ export class Telemetry {
...
-      });
-    }
-    // We send a generic action metric to CTW regardless
-    const leafNamespace = namespace.split(">").map(trim).pop();
-    this.countMetric(`interaction.${action}`, 1, compact([leafNamespace]));
+  static trackInteraction(action: string) {
+    // We send an action metric to CTW
+    this.countMetric(`action.${action}`, 1);
+    // We report an active session to CTW
+    this.reportActiveSession().catch((error) => Telemetry.logError(error as Error));
```
mrosata, line 193: Simplified this `trackInteraction` function a lot. We won't log from here and we won't worry about event type (ie: click) or namespace (ie: ComponentA > B > Item). Instead let's just concern ourselves with the action taken, send a metric and that's all.

The `this.reportActiveSession()` bit will report an active session to datadog only if the user hasn't made another interaction in the past hour.

pmadria3 (approved):

PR #1022, opened 2023-07-21
CDEV-70/details drawer actions

WesleyKapow:
Reworked row actions and plumbed them to the details drawer. This way details drawer will always show the same actions as shown on row hover.

Additionally:
* Reworked row actions to optionally take an `onSuccess` prop so that we can close the drawer for certain actions (dismiss/mark as new/remove).
* Refactored the addition of dismiss & read actions into `useAdditionalResourceActions` so that it can be used by both the `ResourceTable` and `ResourceDetailsDrawer`.
* Reworked dismiss/restore and read/new to update the model record optimistically. This fixes a bug where the row action button on hover would briefly show the old state after the spinner was done. It also allows us to mark a record as read and not wait before opening the drawer. The drawer will then correctly render a "mark as new" button and not a "mark as viewed" button.
* Reworked rendering of row actions to return null if there are no row actions. This allows the details drawer footer to be hidden when there are zero row actions.
* Details drawer was previously taking `enableFQS` but not doing anything with it. Removed!

Screenshot 2023-07-21 at 1 13 31 PM

### DIFF
--- a/.changeset/beige-berries-sort.md
+++ b/.changeset/beige-berries-sort.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": patch
+---
+
+Add table row actions to details drawer (dismiss/add/mark as new/etc).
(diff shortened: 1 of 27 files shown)


pmadria3 (approved):

eric-gunther (commented):
@WesleyKapow - I noticed 1 issue during testing, if there are no row actions then we don't display the close button on the drawer.

https://github.com/zeus-health/ctw-component-library/assets/97455910/653f7c60-c2d4-49e2-805c-59432f87c982

WesleyKapow:
@eric-gunther that is by design and part of the ticket.

PR #1034, opened 2023-07-25
CDEV-322: Showing indicator in Diagnostic Report list results

patrickvalle:
https://zeushealth.atlassian.net/browse/CDEV-322

Screenshot 2023-07-25 at 4 45 11 PM

### DIFF
--- a/src/fhir/models/diagnostic-report.ts
+++ b/src/fhir/models/diagnostic-report.ts
@@ -13,7 +13,7 @@ export class DiagnosticReportModel extends FHIRModel<fhir4.DiagnosticReport> {
 
   private observationModels?: ObservationModel[];
 
-  private hasTrends?: boolean;
+  public hasTrends: boolean;
 
   constructor(
     resource: DiagnosticReport,
@@ -22,6 +22,7 @@ export class DiagnosticReportModel extends FHIRModel<fhir4.DiagnosticReport> {
     trends?: ObservationModel[]
   ) {
     super(resource, includedResources, revIncludes);
+    this.hasTrends = false;
     if (resource.id) {
       const resourceId = resource.id;
       this.observationModels = this.resource.result?.map((result) => {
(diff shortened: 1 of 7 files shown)


pmadria3:
need changeset

pmadria3 (dismissed):
LGTM - could you add a screenshot for what it looks like in the table view?

src/fhir/models/diagnostic-report.ts
```diff
@@ -67,7 +100,21 @@ export class DiagnosticReportModel extends FHIRModel<fhir4.DiagnosticReport> {
...
+        display = `${this.results.length} result`;
+        break;
+      default:
+        display = `${this.results.length} results`;
+        break;
+    }
+    if (this.trends) {
+      display += ", result trend available";
+    }
+    return display;
```
mrosata, line 117: Up to you, might read better more compact
```suggestion
    if (!this.results.length) {
      return "";
    }
    const tResult = this.results.length > 1 ? "results" : "result";
    let display = `${this.results.length} ${tResult}`;
    if (this.trends) {
      display += ", result trend available";
    }
    return display;
```

src/fhir/models/diagnostic-report.ts
```diff
@@ -9,6 +11,37 @@ import { find } from "@/utils/nodash";
 export class DiagnosticReportModel extends FHIRModel<fhir4.DiagnosticReport> {
   kind = "DiagnosticReport" as const;
 
+  private observationModels?: ObservationModel[];
+
+  private hasTrends?: boolean;
```
mrosata, line 16: Suggest to make this public and not conditional, make sure to always set this in the constructor, not only when it is false, then remove the `this.trends` accessor method
```suggestion
  private hasTrends: boolean;
```

src/fhir/models/diagnostic-report.ts
```diff
@@ -91,6 +138,14 @@ export class DiagnosticReportModel extends FHIRModel<fhir4.DiagnosticReport> {
...
     return this.resource.result ?? [];
   }
+
+  get observations() {
+    return this.observationModels ?? [];
+  }
+
+  get trends() {
+    return this.hasTrends || false;
+  }
```
mrosata, line 148: Goes with my last comment on the `this.hasTrends`. We should just always set `this.hasTrends` in the constructor (instead of only setting it if it's true) and then code can directly access `model.hasTrends` instead of needing this accessor method to proxy the hasTrends property
```suggestion
```

pmadria3 (approved):

PR #1038, opened 2023-07-27
Fix Show All 

pmadria3:
Added a UseEffect to make sure the show all is only rendered after the records length is 0

### DIFF
--- a/src/components/core/table/table.tsx
+++ b/src/components/core/table/table.tsx
@@ -79,8 +79,9 @@ export const Table = <T extends MinRecordItem>({
   useEffect(() => {
     if (hidePagination && hasData) {
       setCount(records.length);
+    } else {
+      setCount(pageSize);
     }
-    setCount(pageSize);
   }, [records, pageSize, hidePagination, hasData]);
 
   return (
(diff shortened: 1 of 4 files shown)


src/components/core/table/table.tsx
```diff
@@ -75,6 +75,14 @@ export const Table = <T extends MinRecordItem>({
 
   const hasData = !isLoading && records.length > 0;
 
+  // useeffect to set the count to the default page size when the records change
+  useEffect(() => {
+    if (hidePagination && hasData) {
+      setCount(records.length);
+    }
+    setCount(pageSize);
```
patrickvalle, line 83: I think there should be an `else` here, otherwise this is going to execute regardless of the above `if`

patrickvalle (changes requested):

src/components/core/table/table.tsx
pmadria3: good catch!

patrickvalle (approved):
lgtm

patrickvalle (approved):

PR #1040, opened 2023-07-27
Minor observation fixes

WesleyKapow:
Refactored observation request to use existing fragment. Added some formatters and added support for them in `observation.value`.

### DIFF
--- a/src/fhir/models/observation.ts
+++ b/src/fhir/models/observation.ts
@@ -1,8 +1,7 @@
 import { FHIRModel } from "./fhir-model";
 import { SYSTEM_LOINC } from "../system-urls";
 import { codeableConceptLabel } from "@/fhir/codeable-concept";
-import { formatDateISOToLocal } from "@/fhir/formatters";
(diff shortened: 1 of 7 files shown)


src/fhir/models/observation.ts
```diff
@@ -48,13 +47,44 @@ export class ObservationModel extends FHIRModel<fhir4.Observation> {
...
   }
 
-  get value() {
-    return compact([
-      this.resource.valueQuantity?.value ||
-        this.resource.valueString ||
-        codeableConceptLabel(this.resource.valueCodeableConcept),
-      this.resource.valueQuantity?.unit,
-    ]).join(" ");
+  get value(): string {
```
patrickvalle, line 50: What's the deal with these updates? They're not logically equivalent so just wondering what the rationale is

patrickvalle: Answered during standup that we need to support these extra values! Closing comment and approving

patrickvalle (approved):

PR #1063, opened 2023-08-02
CDEV-309 Consistent font size and nowrapping for ZAP More button

melocule:
CDEV-309

AC:
- Tab content honors word-break: keep-all
- The size of all text on tab labels is .875 rem

Solution:
I just applied some relevant tailwind classes that the tabs had but that "More" was missing-- `ctw-whitespace-nowrap` and `ctw-text-sm`.

Font size after fix:
Pasted Graphic 1

No wrapping after fix:
Pasted Graphic 6


src/components/core/tab-group/tab-group.tsx
```diff
@@ -152,7 +152,7 @@ function TabGroupComponent({
               ref={moreMenuRef}
               selectedIndex={selectedTabIndex - tabOverflowCutoff}
               btnClassName={cx(
-                "ctw-more-tab ctw-capitalize ctw-flex-shrink-0 ctw-h-full",
+                "ctw-more-tab ctw-text-sm ctw-whitespace-nowrap ctw-capitalize ctw-flex-shrink-0 ctw-h-full",
```
luthfib, line 155: Curious why did we add the `ctw-text-sm`?

package-lock.json
patrickvalle, line 1: Did you mean to commit these package lock changes?

src/components/core/tab-group/tab-group.tsx
melocule: One of the things noted in the ticket was that the "More" text in Sandbox is a different font size than the text in the rest of the buttons, although this isn't the case in `npm run dev`

melocule: so font size was 100% which can appear inconsistent from the rest of the tabs which had `ctw-text-sm`

package-lock.json
melocule: Uhh I think it modified itself when I ran `npm install`, idk why it did that but i'll try to revert it

luthfib (approved):

pmadria3 (approved):

PR #1066, opened 2023-08-02
CDEV-352 create new encounters component and remove timeline from ZAP

sarahmangels:
image
image

### DIFF
--- a/src/components/content/encounters/helpers/columns.tsx
+++ b/src/components/content/encounters/helpers/columns.tsx
@@ -1,0 +1,51 @@
+import { SimpleMoreList } from "@/components/core/simple-more-list";
+import { TableColumn } from "@/components/core/table/table-helpers";
+import { EncounterModel } from "@/fhir/models/encounter";
+import { compact } from "@/utils/nodash";
+
+export const patientEncounterColumns: TableColumn<EncounterModel>[] = [
+  {
+    title: "Date",
+    widthPercent: 10,
+    minWidth: 120,
+    dataIndex: "periodStart",
+  },
+  {
+    title: "Title",
+    render: (encounter) => (
+      <div>
+        <div>{encounter.typeDisplay}</div>
+      </div>
+    ),
+  },
+  {
+    title: "Organization",
+    render: (encounter) => (
+      <>
+        {compact([encounter.participants, encounter.location]).map((detail) => (
+          <div className="ctw-capitalize" key={detail}>
+            {detail.toLocaleLowerCase()}
+          </div>
+        ))}
+      </>
+    ),
+  },
+  {
+    title: "Details",
+    render: (encounter) => {
+      const diagnoses = compact(encounter.diagnoses);
+      return (
+      <div>
+        {/* TODO: add clinical notes to encounter table CDEV-306 */}
+        {diagnoses.length > 0  && (
+          <SimpleMoreList
+            items={diagnoses}
+            limit={3}
+            total={diagnoses.length}
+            prefix="Diagnosis:"
+          />
+        )}
+      </div>
+    )},
+  },
+];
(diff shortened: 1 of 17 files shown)


luthfib:
changeset?

patrickvalle:
can we get a changeset?

oh luthfi beat me to it again! i was just about to say that too!

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -33,13 +34,14 @@ export type ZAPResourceName =
...
   | "medications-all"
-  | "timeline";
+  | "timeline"
+  | "encounters";
 
 export const defaultZAPResources: ZAPResourceName[] = [
   "conditions-all",
   "medications-all",
   "diagnostic-reports",
-  "timeline",
```
sarahmangels, line 42: removing timeline from megaZAP, adding encounters

src/components/content/encounters/helpers/modal-hooks.tsx
```diff
@@ -0,0 +1,31 @@
+import { useResourceDetailsDrawer } from "../../resource/resource-details-drawer";
+import { CodingList } from "@/components/core/coding-list";
+import { EncounterModel } from "@/fhir/models/encounter";
+import { capitalize } from "@/utils/nodash/fp";
+
+export function usePatientEncounterDetailsDrawer() {
+  return useResourceDetailsDrawer({
+    header: (m) => `${m.periodStart} - ${m.periodEnd}`,
+    subHeader: (m) => m.typeDisplay,
```
melocule, line 9: Subheaders for all of them show up as undefined for me, I'm not sure why considering that both here and in the Title column (where it shows up the way it should) you're using `m.typeDisplay` 😥
image

src/components/content/encounters/helpers/columns.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </div>
+    ),
+  },
+  {
+    title: "Organization",
+    render: (encounter) => (
+      <>
+        {compact([encounter.participants, encounter.location]).map((detail) => (
+          <div className="ctw-capitalize" key={detail}>
+            {detail.toLocaleLowerCase()}
```
melocule, line 27: but what if my last name is McDonald :'C it'll be displayed as Mcdonald right

melocule: Also may be worth noting that the Figma seems to allow stuff like all caps and non-title-caps

src/components/content/encounters/helpers/columns.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <div>
+        <div>{encounter.typeDisplay}</div>
+      </div>
+    ),
+  },
+  {
+    title: "Organization",
+    render: (encounter) => (
+      <>
+        {compact([encounter.participants, encounter.location]).map((detail) => (
```
melocule, line 25: i was personally a bit confused when i saw the "Organization" column being just a bunch of names because I guess they didn't record a location, should we confirm with UX that this should still be the column name?

Figma seems to only have examples where there is a location, so idk if they thought about this case

melocule (dismissed):

src/components/content/encounters/helpers/modal-hooks.tsx
bryanjknight: I think this is because `m.periodEnd` is undefined; could we address this real quick?

bryanjknight (commented):
Mostly lgtm, let's just address the items Melody flagged

src/components/content/encounters/helpers/modal-hooks.tsx
sarahmangels: what do you want it to be if the periodEnd is undefined? This feels out of scope for my work - this is an existing drawer

bryanjknight: Just the start date; I would argue it's in scope as you are passing in the function that defines the header but I also fine with addressing it in a follow up PR

bryanjknight (approved):

src/components/content/encounters/helpers/columns.tsx
sarahmangels: good point - follow up ticket: https://zeushealth.atlassian.net/browse/CDEV-360

PR #1067, opened 2023-08-02
Feature/CDEV 306 notes in encounter table

sarahmangels:
Adding notes to the details column in encounters component

image

### DIFF
--- a/src/components/content/encounters/helpers/columns.tsx
+++ b/src/components/content/encounters/helpers/columns.tsx
@@ -34,7 +34,7 @@ export const patientEncounterColumns: TableColumn<EncounterModel>[] = [
     title: "Details",
     render: (encounter) => {
       const diagnoses = compact(encounter.diagnoses);
-      const notes = compact(encounter.clinicalNotes.map((note) => note.noteTitle));
+      const notes = encounter.clinicalNotes.map((note) => note.noteTitle);
       return (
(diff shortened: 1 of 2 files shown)


melocule:
Wait you need a changeset for this one right

pmadria3 (approved):

src/components/content/encounters/helpers/columns.tsx
```diff
@@ -34,9 +34,17 @@ export const patientEncounterColumns: TableColumn<EncounterModel>[] = [
     title: "Details",
     render: (encounter) => {
       const diagnoses = compact(encounter.diagnoses);
+      const notes = compact(encounter.clinicalNotes.map((note) => note.title));
```
melocule, line 37: If a note has no title should we still list something there instead of compacting it out of the list? Like "untitled"
If there's just an untitled note the user might end up thinking there are none

melocule: other than that looks good

melocule (dismissed):

src/components/content/encounters/helpers/columns.tsx
sarahmangels: good point, will update!

bryanjknight (approved):

PR #1088, opened 2023-08-04
CTW-1315/metrics bake off

mrosata:
Start sending analytics to ctw `/report/analytics`. This will send analytics from all environments, but they don't all get records. CTW determines whether or not to actually send analytic data to a third party, currently it only does so in sandbox, otherwise a 404 is returned ([see PR)](https://github.com/zeus-health/ctw/pull/664).

There's no PHI in the body of the request. On the server CTW can pull any user information from the decoded authentication header.


src/utils/telemetry.ts
```diff
@@ -261,6 +262,38 @@ export class Telemetry {
     });
   }
 
+  /**
+   * Report User analytic events
+   */
+  static async analyticsEvent(eventName: string, eventProperties: Record<string, unknown> = {}) {
+    // if (this.shouldSkipSendingMetrics()) {
+    //   return;
+    // }
```
pmadria3, line 271: can remove

mrosata: good catch

pmadria3 (approved):

PR #1095, opened 2023-08-07
Remove ODS code path for UPID-based queries

pmadria3:
CDEV-330

Note to Reviewer: Please run my branch locally to test and **DO NOT MERGE** immediately after approval.

This PR removes ODS related queries and functions for the resources we have setup.

### DIFF
--- a/src/components/content/allergies/helpers/history.tsx
+++ b/src/components/content/allergies/helpers/history.tsx
@@ -1,5 +1,4 @@
 import { AllergyIntolerance } from "fhir/r4";
-import { SearchParams } from "fhir-kit-client";
 import { HistoryEntryProps } from "../../resource/helpers/history-entry";
 import { useHistory } from "../../resource/history";
 import { AllergyModel } from "@/fhir/models/allergies";
@@ -13,28 +12,9 @@ export function useAllergiesHistory(allergy: AllergyModel) {
     queryKey: QUERY_KEY_ALLERGY_HISTORY,
     includeVersionHistory: false,
     valuesToDedupeOn,
-    getSearchParams,
     getHistoryEntry,
     clientSideFiltersFQS,
   });
-}
-
-function getSearchParams(allergy: AllergyModel) {
-  const tokens = allergy.knownCodings.map((coding) => `${coding.system}|${coding.code}`);
-
-  const searchParams: SearchParams = {
-    _include: ["AllergyIntolerance:patient"],
-    "_include:iterate": "Patient:organization",
-  };
-
-  if (tokens.length > 0) {
-    searchParams.code = tokens.join(",");
-  } else {
-    // eslint-disable-next-line no-underscore-dangle
-    searchParams._id = allergy.id;
-  }
-
-  return searchParams;
 }
 
 function clientSideFiltersFQS(model: AllergyModel, allergies: AllergyIntolerance[]) {
--- a/src/components/content/resource/history.ts
+++ b/src/components/content/resource/history.ts
@@ -64,8 +64,7 @@ export function useHistory<T extends ResourceTypeString, M extends FHIRModel<Res
           clientSideFiltersFQS,
           getFiltersFQS?.(model)
         ),
-      `req.${model.resourceType.toLowerCase()}_history`,
-      ["fqs"]
+      `req.${model.resourceType.toLowerCase()}_history`
     ),
     fqsProvenances.ready && useHistoryUnleash.ready
   );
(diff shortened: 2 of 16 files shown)


src/components/content/resource/history.ts
```diff
@@ -58,38 +50,23 @@ export function useHistory<T extends ResourceTypeString, M extends FHIRModel<Res
...
+          requestContext,
+          patient,
+          valuesToDedupeOn,
+          getHistoryEntry,
+          fqsProvenances.enabled,
+          clientSideFiltersFQS,
+          getFiltersFQS?.(model)
         ),
+      `req.${model.resourceType.toLowerCase()}_history`,
+      ["fqs"]
```
patrickvalle, line 68: Seems like we don't need the `fqs` tag any longer on any of these, as there's only fqs

src/fhir/care-team.ts
```diff
@@ -13,29 +11,10 @@ export function usePatientCareTeam() {
     [],
     "careTeams",
     "req.timing.care_teams",
-    getCareTeamFQS,
-    getCareTeamODS
+    getCareTeamFQS
```
patrickvalle, line 14: Should we move away from using `useFeatureFlaggedQueryWithPatient` and back to `useQueryWithPatient`?

pmadria3: Can't atm as Basic and such use useQueryWithPatient and they are still on ODS. Would be a messy refactor atm. We can remove it when we remove ODS after builder-scoped queries are completed.

Our UPID-based queries also call Basic so it will be a little gnarly atm

src/fhir/allergies.ts
```diff
@@ -43,31 +40,7 @@ async function getAllergyIntoleranceFromFQS(
     const nodes = data.AllergyIntoleranceConnection.edges.map((x) => x.node);
     const results = applyAllergyFilters(nodes, requestContext.builderId);
     if (results.length === 0) {
-      Telemetry.countMetric("req.count.allergies.none", 1, ["fqs"]);
```
sarahmangels, line 46: is this getting removed?

src/fhir/allergies.ts
```diff
@@ -43,31 +40,7 @@ async function getAllergyIntoleranceFromFQS(
...
-    }
-    Telemetry.histogramMetric("req.count.allergies", results.length, ["fqs"]);
```
sarahmangels, line 48: same

src/fhir/allergies.ts
sarahmangels: oh i see -- is removing the "fqs" tag going to mess up any of the datadog dashboards? are there dashboards that are currently relying on that tag? wondering if we should keep it for now

src/fhir/care-team.ts
```diff
@@ -13,29 +11,10 @@ export function usePatientCareTeam() {
     [],
     "careTeams",
     "req.timing.care_teams",
-    getCareTeamFQS,
```
sarahmangels, line 16: cant comment on line 9 but we should stop using `useFeatureFlaggedQueryWithPatient` since we no longer have a feature flag and instead go back to using`useQueryWithPatient`

src/components/content/allergies/helpers/history.tsx
```diff
@@ -13,30 +12,11 @@ export function useAllergiesHistory(allergy: AllergyModel) {
     queryKey: QUERY_KEY_ALLERGY_HISTORY,
     includeVersionHistory: false,
     valuesToDedupeOn,
-    getSearchParams,
     getHistoryEntry,
     clientSideFiltersFQS,
   });
 }
 
-function getSearchParams(allergy: AllergyModel) {
```
sarahmangels, line 22: why removing this?

src/fhir/care-team.ts
sarahmangels: (same comment in lots of other places FYI)

src/fhir/encounters.ts
```diff
@@ -19,7 +19,7 @@ export function usePatientEncounters() {
   return useQueryWithPatient(
```
sarahmangels, line 19: here is an example of how to do `useQueryWithPatient`

sarahmangels (dismissed):
i think we should look for everywhere we are currently using `useFeatureFlaggedQueryWithPatient` and replace it with `useQueryWithPatient` like we had before we had the feature flag

also -- lets double check that removing the `fqs` tag from the telemetry metrics wont break existing DD dashboards. It may just be easier to keep that tag for now if DD stuff is relying on it

src/components/content/allergies/helpers/history.tsx
pmadria3: We no longer need searchparams since it was utilized in the ODS calls

src/fhir/encounters.ts
pmadria3: awesome TY!

pmadria3:
Addressed the comments 

Keeping the function as we will need them shortly in the future for replacing the other queries

sarahmangels (approved):

PR #1096, opened 2023-08-08
This appears to fix table rendering in firefox

thinkbrown:


mrosata (approved):
tested locally, still works in chrome and fixes Firefox

mrosata (approved):

PR #1098, opened 2023-08-08
Cdev 353 new encounters notes component has discoverability functionality of other components

melocule:
CDEV-353

AC:
- First party data distinguished from outside data using check mark icon
- Read/unread and dismiss functionality available
image

- Notification dot in ZAP tab indicating there is unread data (MegaZAP and standalone Zus App)
image

- Zus users cannot mark data as read in prod
image


src/components/content/encounters/helpers/modal-hooks.tsx
```diff
@@ -1,15 +1,23 @@
...
+  RowActions?: RowActionsProp<EncounterModel>;
+  enableDismissAndReadActions?: boolean;
+}) {
   return useResourceDetailsDrawer({
     header: (m) => `${m.periodStart} - ${m.periodEnd}`,
     subHeader: (m) => m.typeDisplay,
     getSourceDocument: true,
     details: encounterData,
-    enableDismissAndReadActions: true,
+    RowActions,
```
melocule, line 19: If we don't plan on adding other actions I can remove this

src/components/content/resource/use-additional-resource-actions.tsx
```diff
@@ -59,6 +61,11 @@ const getDismissAndReadActions =
 
     const readLabel = record.isRead ? t("resourceTable.unread") : t("resourceTable.read");
 
+    // In production we want to avoid non-builder users from inadvertantly marking records as read
+    // In lower environments this is allowed for demos/testing
+    const disableReadButton =
+      requestContext?.env === "production" && requestContext.userType !== "builder";
```
sarahmangels, line 67: `requestContext.userType !== "builder"` because the only other user type right now is 'zus', right? would it be better to say `requestContext.userType == "zus"` ? maybe NBD, but just in case other user types are ever added?

sarahmangels (approved):

PR #1104, opened 2023-08-10
Fix double loading encounters from FQS in both ZAP and Encounters & N…

sarahmangels:
…otes component

### DIFF
--- a/src/fhir/basic.ts
+++ b/src/fhir/basic.ts
@@ -86,19 +86,19 @@ function mapBasics<R extends fhir4.Resource, T extends FHIRModel<R>>(
   queryData: T[],
   basicsData: Basic[]
 ) {
-  const resources2 = queryData;
+  const resources = queryData;
   const basics = basicsData;
   // If basic data came back from the above useBasic call, manually map any basic data to the resources
   // it corresponds to.
   if (basics.length > 0) {
-    resources2.forEach((a, i) => {
+    resources.forEach((a, i) => {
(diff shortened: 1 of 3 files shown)


src/fhir/basic.ts
```diff
@@ -82,38 +82,48 @@ export function useBasic(fqs: FeatureToggle) {
   );
 }
 
+function mapBasics<R extends fhir4.Resource, T extends FHIRModel<R>>(
+  queryData: T[],
+  basicsData: Basic[]
+) {
+  const resources2 = queryData;
```
melocule, line 89: why the 2 here?

melocule (approved):
The fix works! 

But, and idk if this is related or even an issue, in ZAP for MedicationStatementConnection and ConditionConnection i get two queries-- one with edges and one without

e.g. 
image

sarahmangels:
@melocule the doubles for meds and conditions are because one is getting regular resources and one is getting lens resources so that is expected!!

src/fhir/basic.ts
```diff
@@ -82,38 +82,48 @@ export function useBasic(fqs: FeatureToggle) {
   );
 }
 
+function mapBasics<R extends fhir4.Resource, T extends FHIRModel<R>>(
+  queryData: T[],
+  basicsData: Basic[]
+) {
+  const resources = queryData;
+  const basics = basicsData;
```
mrosata, line 90: why not rename the params?

PR #1114, opened 2023-08-14
Cdev 368 remove default filter presets for conditions

melocule:
CDEV-368

image
image
image
image


src/components/content/conditions/helpers/filters.ts
```diff
@@ -1,30 +1,61 @@
 import { faClock } from "@fortawesome/free-regular-svg-icons";
 import { faClipboardCheck, faClipboardList } from "@fortawesome/free-solid-svg-icons";
+import intersectionWith from "lodash/intersectionWith";
```
patrickvalle, line 3: I think we've been wrapping lodash imports w/ `nodash` in **src/utils/nodash/fp.ts,** e.g.

`export { default as intersectionWith } from "lodash/fp/intersectionWith";`

patrickvalle (commented):
I did a quick pass of the AC in the ticket, so I could be wrong here... but is this possible to do in this cut of the ticket:
> Remove “Entered in Error” option - we shouldn’t display these records to users anyways

patrickvalle (approved):
lgtm

PR #1115, opened 2023-08-14
CDEV-384/Analytics capture UPID and component

mrosata:
- `TelemetryProvider` now provides a reference to `Telemetry` class
- Added `AnalyticsProvider` and `useTelemetry` hook
- `useTelemetry` hook provides an abstraction to `Telemetry.trackInteraction()`. The abstraction will include both the UPID and and component name in the analytics sent to /report/analytics.
- Wrapped components that need to have their component name in analytics in the `AnalyticsProvider`
- Using `trackInteraction` in place of `data-zus-telemetry-track-click`
- Added interactions to sorting and filtering


sarahmangels (approved):

src/components/content/document/patient-documents.tsx
```diff
@@ -101,12 +106,11 @@ const RowActions = ({ record, onSuccess, onAddToRecord }: RowActionsProps2) => {
       <button
         type="button"
         className="ctw-btn-primary ctw-ml-1 ctw-capitalize"
-        data-zus-telemetry-click="Add to record"
-        data-testid="add-to-record"
         onClick={async () => {
           const binary = await getBinaryDocument(await getRequestContext(), binaryId);
           onAddToRecord(record, binary);
           onSuccess?.();
+          trackInteraction("btn_add_to_record");
```
patrickvalle, line 113: Should this be more specific to documents vs just "Add to record"?

mrosata: because it's in a component wrapped with
```
<AnalyticsProvider componentName="PatientDocuments">
```
It will send as a metric `btn_add_to_record` with metadata of
```
{
  upid: "abcd-1234...",
  componentName: "PatientDocuments"
}
```
so in theory this button could be reused around the app (though at the moment I think it may not be reused iirc)

mrosata: I shouldn't have removed the `data-testid` though

sarahmangels (approved):

PR #1131, opened 2023-08-22
DEV-307: Encounter NoteType filtering

patrickvalle:
https://zeushealth.atlassian.net/browse/CDEV-307

Added Note Type filtering to Encounters. Additionally, added support to checkbox-type filters so that they can use a `predicate` function for filtering data vs just a property key.

Screenshot 2023-08-23 at 3 12 34 PM


src/components/content/encounters/helpers/filters.tsx
```diff
@@ -1,15 +1,80 @@
+import { faClipboardCheck } from "@fortawesome/free-solid-svg-icons";
 import { FilterChangeEvent, FilterItem } from "@/components/core/filter-bar/filter-bar-types";
 import { EncounterModel } from "@/fhir/models/encounter";
+import { SYSTEM_LOINC } from "@/fhir/system-urls";
+
+export const noteTypeValues = [
```
sarahmangels, line 6: one thing on these, if possible we should try to only show the ones where there would be any data. so crossing thing hard coded list with whatever values are actually in the table. Maybe ask Tai if thats worth the effort - idk how hard that is. But that is how the other filters work

src/components/content/encounters/helpers/filters.tsx
```diff
@@ -1,15 +1,80 @@
...
+  },
+  {
+    name: "History of Present Illness",
+    key: "10164-2",
+  },
+  {
+    name: "Reason for Visit",
+    key: "29299-5",
+  },
+];
```
sarahmangels, line 27: ACs suggest an option for "other"  -- tho im not sure how useful that is? maybe check w Tai

sarahmangels: oh nvm i see u have it below!

sarahmangels (approved):

src/components/content/encounters/helpers/filters.tsx
patrickvalle: ooh good catch, will update

patrickvalle: done

sarahmangels (approved):

PR #1143, opened 2023-08-24
Cdev 364 formatting is squished for notes in tabular format

melocule:
CDEV-364

- Makes the tables in notes responsive and adds some styling to make them look better.
- Also changes the way we display notes, converting the HTML to JSX first. This lets us avoid using `dangerouslySetInnerHTML` and has the added benefit of styles being applied to them (e.g. headers in notes now appear like other headers instead of as plain text)

Full width
image

Somewhat squished? Rows and columns pivot.
image

Ridiculously thin? Rows become scrollable (this is mid-scroll).
image


.github/workflows/storybook.yaml
```diff
@@ -19,5 +19,5 @@ jobs:
       - name: Deploy
         uses: JamesIves/github-pages-deploy-action@3.6.2
         with:
-          branch: gh-pages
```
melocule, line 22: Got a parsing error when trying to run storybook, and in the file I saw an error saying that the required properties "BRANCH" and "FOLDER" were missing. Went away once I put these two in all caps

PR #1149, opened 2023-08-28
HTML to JSX

melocule:
Changes the way we display notes, so that default styling is applied (e.g. headers no longer display as raw text, but in the same way headers in the other components are displayed). 

This new method turns the HTML to JSX first, which has the added benefit of letting us avoid using dangerouslySetInnerHTML, and letting us modifying HTML elements into React components in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@zus-health/ctw-component-library",
-  "version": "1.60.0",
+  "version": "1.60.2",
   "lockfileVersion": 2,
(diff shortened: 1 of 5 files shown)


mrosata (approved):
We also use dangerouslySetInnerHTML here https://github.com/zeus-health/ctw-component-library/blob/67e5834088b6de0bedeb56b6de8054eefeed91d4/src/components/content/CCDA/ccda_viewer/components/Section/Section.tsx#L28-L30

PR #1151, opened 2023-08-29
Replace dangerously set inner html (part 2)

melocule:
Replace dangerously setting inner html with Interweave instead, for CCDA docs.


melocule:
I test-viewed a CCDA doc to make sure it still displays properly

PR #1161, opened 2023-09-01
CTW-1366/Fix width of medication row

mrosata:
- Remove padding from medication row
- CTW resets the browser h3 margin-block-start to 0. To normalize it this PR sets it to `1rem` (just for this component). There is no tailwind equivalent that I could find which is why I used css and not a tailwind class.


luthfib:
Changeset?

luthfib (approved):

PR #1164, opened 2023-09-05
Cdev 383 expand interaction tracking to all components

melocule:
CDEV-383

Adds typing and more consistency in the tracking names. Also tracks some more stuff.


src/SecuredApp.tsx
```diff
@@ -24,7 +26,10 @@ const SecuredAppComponent = ({ AppComponent }: SecuredAppProps) => {
 
       <button
         type="button"
-        onClick={() => logout({ returnTo })}
+        onClick={() => {
+          trackInteraction("logout");
+          logout({ returnTo });
+        }}
```
mrosata, line 32: One thing to check for here is that "logout" events actually make it into Amplitude. This concern may turn out not to be valid, but if the Auth0 logout redirect happens immediately before the browser makes the analytic POST request to ctw with the "logout" event then we might have to change this to await the sending of the analytic event before calling `logout({ returnTo })`.

If you do need to do this, there's a helper `const { analyticsEvent } = useAnalytics` that calls the analytics service directly and can be awaited.

https://github.com/zeus-health/ctw-component-library/blob/6349b423df8a3f9535719817500564e599233a94/src/components/core/providers/analytics/use-analytics.ts#L27-L29
> side note: the helper method `analyticsEvent` doesn't put the componentName and upid on the event, which is probably okay in this situation anyways.

src/components/content/CCDA/ccda_viewer/components/Section/Section.tsx
```diff
@@ -5,16 +5,21 @@ import { useState } from "react";
...
   const [isOpen, setIsOpen] = useState(true);
+  const { trackInteraction } = useAnalytics();
 
   if (!humanReadable) return null;
 
-  const handleClick = () => setIsOpen(!isOpen);
+  const handleClick = () => {
+    setIsOpen(!isOpen);
+    trackInteraction(isOpen ? "collapse_document_section" : "expand_document_section");
+  };
```
mrosata, line 22: We could make the event of clicking the document section a single event where "collapse" and "expand" are metadata on the event.
  
```suggestion
  const handleClick = () => {
    setIsOpen(!isOpen);
    trackInteraction("toggle_document_section", {
      direction: isOpen ? "collapse" : "expand",
    });
  };
```

src/components/content/observations/helpers/trends.tsx
```diff
@@ -53,13 +53,13 @@ export const ObservationTrends = ({ model }: ObservationTrendsProps) => {
                       })}
                       onClick={() => {
                         if (trend.diagnosticReport) {
-                          trackInteraction("open_diagnostic_report");
+                          trackInteraction("open_diagnostic_report", { target: "btn" });
```
mrosata, line 56: nice idea 👍🏼

src/components/content/resource/helpers/history-entry.tsx
```diff
@@ -84,9 +91,7 @@ const DetailSummary = ({
       aria-label="details"
       onClick={() => {
         setIsDetailShown(!isDetailShown);
-        trackInteraction("expand_detail", {
-          direction: isDetailShown ? "collapse" : "expand",
-        });
+        trackInteraction(isDetailShown ? "collapse_history_entry" : "expand_history_entry");
```
mrosata, line 94: This is same as a previous comment. Is there a reason we wouldn't want to group these?
```suggestion
        trackInteraction("toggle_detail", {
          direction: isDetailShown ? "collapse" : "expand",
        });
```

src/components/core/providers/analytics/use-analytics.ts
```diff
@@ -2,6 +2,65 @@ import { useContext, useMemo } from "react";
 import { AnalyticsContext } from "./context";
 import { usePatient } from "@/components/core/providers/patient-provider";
 import { useTelemetry } from "@/components/core/providers/telemetry/use-telemetry";
+import { SortDir } from "@/utils/sort";
+
+// Tracked events indicate the effect the user initiated.
+// They don't indicate details like patient, overarching component, or cause, which belong in metadata.
+type Action =
```
mrosata, line 9: good idea

src/SecuredApp.tsx
melocule: Just realized this button isn't part of the CTW app

mrosata (dismissed):

mrosata (approved):

PR #1200, opened 2023-09-14
fix: only show one date if both dates are the same

bryanjknight:
If an encounter has both a start and end date that are the same, we used to render it as `01/01/2022 - 01/01/2022`. With this PR, it will now only show `01/01/2022`

### DIFF
--- a/src/utils/dates.ts
+++ b/src/utils/dates.ts
@@ -1,6 +1,4 @@
 import { parse, parseISO } from "date-fns";
-// eslint-disable-next-line import/no-named-default
-import { default as LOCALE_US } from "date-fns/locale/en-US";
 
 const DATE_FORMATS = [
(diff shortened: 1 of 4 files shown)


src/utils/dates.ts
pmadria3, line 1: Do you want to move this into `formatters.ts`? We have all our date functionality in there

bryanjknight: maybe, but it's a parser, not a formatter

bryanjknight: also these are non-standard date formats in the CDA, so that's why I want to keep it separate from the fhir logic

pmadria3: That's also a fair point

melocule (approved):

pmadria3 (approved):

PR #1203, opened 2023-09-14
CTW-1393 Search UI improvements

eric-gunther:
- Modified search to only use keyword search, limit to 10 results, and don't search observations. We will continue to tweak this but these settings have yielded sensible results so far during testing.
- Created generic SearchResult component and removed custom components for conditions/documents/etc.
- Modified what data we show for the different search results based on feedback during meeting with design. Will continue to iterate on these.
- Added empty states.
- Updated styling and made it so the heading is the button to open the drawer. Still struggling a bit with getting the text overflow working on the medications heading, but may punt on that in the interest of getting this out faster for people to validate search results.


src/components/content/patient-record-search/helpers/search-result-row.tsx
```diff
@@ -20,24 +22,137 @@ export function SearchResultRow(props: ResourceRowProps) {
...
+    // TODO - this could be improved to highlight all of the spans and not just the first one
+    relevantContent = (
+      <>
+        <span>{content.substring(spans[0].begin - 100, spans[0].begin)}</span>
+        <span className="ctw-font-semibold">{content.substring(spans[0].begin, spans[0].end)}</span>
+        <span>{content.substring(spans[0].end, spans[0].end + 300)}</span>
+      </>
+    );
+  }
+  return relevantContent;
```
mrosata, line 157:
```suggestion
  if (spans.length === 0) {
    return content;
  }
  // TODO - this could be improved to highlight all of the spans and not just the first one
  return (
    <>
      <span>{content.substring(spans[0].begin - 100, spans[0].begin)}</span>
      <span className="ctw-font-semibold">{content.substring(spans[0].begin, spans[0].end)}</span>
      <span>{content.substring(spans[0].end, spans[0].end + 300)}</span>
    </>
  );
```

src/components/content/patient-record-search/helpers/search-result.tsx
```diff
@@ -0,0 +1,48 @@
...
+  text,
+  details,
+}: SearchResultProps<T>) {
+  return (
+    <div className="ctw-patient-record-search-result">
+      <div className="ctw-flex ctw-items-end ctw-justify-between">
+        <span>
+          <button
+            type="button"
+            className="ctw-patient-record-search-result-btn"
```
mrosata, line 29: We can use `mr-1.5` instead of `&nbsp;`
```suggestion
            className="ctw-patient-record-search-result-btn mr-1.5"
```

src/components/content/patient-record-search/helpers/search-result.tsx
```diff
@@ -0,0 +1,48 @@
...
+            onClick={() => openDetails?.(resource)}
+          >
+            {heading}
+          </button>
+          &nbsp;
```
mrosata, line 34:
```suggestion
```

mrosata (approved):

PR #1215, opened 2023-09-18
fix: address clinical note issues

bryanjknight:
Addresses a few issues with clinical notes:
1. Verify the table actually has a `thead` and `tbody` before trying to transpose
2. Show loading status while transposing table
3. Switch from `<table>` to `<div>` for transposed rows. This results in a cleaner experience

### DIFF
--- a/src/components/content/resource/helpers/details-card.tsx
+++ b/src/components/content/resource/helpers/details-card.tsx
@@ -73,6 +73,11 @@ export const recursivelyTransposeTables = (node: ReactNode, curDepth: number): R
 };
 
 const transposeTable = (tbl: ReactElement): ReactElement => {
+  // if there's no children, then return with original tbl
+  if (!tbl.props || !tbl.props.children || !Array.isArray(tbl.props.children)) {
+    return tbl;
+  }
+
   // find the thead, otherwise skip
   const theads = (tbl.props.children as ReactElement[]).filter((child) => child.type === "thead");
   if (theads.length !== 1) {
--- a/src/components/content/resource/helpers/details-card.tsx
+++ b/src/components/content/resource/helpers/details-card.tsx
@@ -2,6 +2,8 @@ import parse from "html-react-parser";
 import React, { ReactElement, ReactNode, useEffect, useState } from "react";
 import { cloneDeep } from "@/utils/nodash";
 
+import "./note-style.scss";
+
 export type DetailEntry = {
   label?: string;
   value: ReactNode;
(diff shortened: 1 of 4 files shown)


src/components/content/resource/helpers/details-card.tsx
```diff
@@ -73,6 +75,11 @@ export const recursivelyTransposeTables = (node: ReactNode, curDepth: number): R
 };
 
 const transposeTable = (tbl: ReactElement): ReactElement => {
+  // if there's no children, then return with original tbl
+  if (!tbl.props || !tbl.props.children || !Array.isArray(tbl.props.children)) {
+    return tbl;
+  }
```
mrosata, line 81: Can probably simplify to
```suggestion
  if (!Array.isArray(tbl.props?.children)) {
    return tbl;
  }
```

src/components/content/resource/helpers/details-card.tsx
```diff
@@ -114,71 +121,45 @@ const transposeTable = (tbl: ReactElement): ReactElement => {
...
-      newRows[transposeIdx] = (
-        <tr key={`transposed-${transposeIdx}`} className={tr.props.className}>
-          {updatedCells}
-        </tr>
+      newRows.push(
+        // eslint-disable-next-line react/no-array-index-key
+        <div key={`cell-${rowIdx}-${cellIdx}`}>
+          <p>
+            <b>{headerRow.props.children}</b>
+          </p>
```
mrosata, line 138: `<b>` should be something like  `.ctw-font-medium` or `.ctw-font-semibold`
```suggestion
          <p className="ctw-font-semibold">{headerRow.props.children}</p>
```

src/components/content/resource/helpers/details-card.tsx
```diff
@@ -114,71 +121,45 @@ const transposeTable = (tbl: ReactElement): ReactElement => {
...
+          <p>{dataCell.props.children}</p>
+        </div>
       );
-
-      transposeIdx += 1;
     });
+
+    if (dataRows.length - 1 > rowIdx) {
+      newRows.push(
```
mrosata, line 145: Just a nit, won't hold up PR, usually using map for iterating children is cleaner than using forEach and pushing to a new array.

src/components/content/resource/helpers/details-card.tsx
```diff
@@ -201,7 +182,8 @@ export const DetailsCard = ({ details, hideEmpty = true, documentButton }: Detai
             // if we want to hide empty rows and the value is falsy and the value
             // is not zero, then hide it
             if (hideEmpty && !value && value !== 0) {
-              return <></>;
+              // eslint-disable-next-line react/no-array-index-key
+              return <div key={idx} />;
```
mrosata, line 185: Thanks for fixing this

mrosata (dismissed):

src/components/content/resource/helpers/details-card.tsx
bryanjknight: yeah, problem here is that it's not one-to-one; I may need to add additional rows for a given row based on whether I need to add a separator line. I think there's probably a cleaner, functional way to do this. I'll poke around some more but if nothing comes to mind I might leave it as is

bryanjknight: Okay, found a way via `reduce`

mrosata (approved):

PR #1218, opened 2023-09-19
CTW-1400 Updates to display in encounters & documents

eric-gunther:
A lot of small, presentation layer, changes here. This was all discussed with product/UX during a session today.

- Updated sample app so you only ever use the new consolidated conditions/medications components (shouldn't be using the old ones anymore)
- Removed "Section Display" from document details drawer. This is listing out all of the LOINC-backed categories in a top level document, but these rarely, if ever, line up with the section level document descriptions. This leads to confusion around what's actually in the document and how it may align with the Encounters & Notes and Search components.
- Updated encounter date column to show a hyphen with "start - end" if it was a multiple day encounter (eg. a hospital visit).
- Cleaned up encounter provider and details columns by adding some padding between elements and showing less data in the `SimpleMoreList`
- Alphabetizing clinical notes in encounter drawer. Ideally these would match the order they appear in the document, but that is not preserved when creating the encounter model, so we might as well alphabetize so it's consistent.
- Updated encounter drawer so that providers/diagnoses are displayed with bullet lists.
- The table transposition logic was causing a lot of performance issues with any drawer using a JSX.Element passed in to the details. Made this optional and only using it for encounter notes.
- Consistently using document.description for the document title unless it's undefined in which case we fall back to the content attachment title. This will reduce misalignment between Encounters & Notes and Search component.
- Updated logic for document encounter date to show hyphen with "start - end" when applicable.
- Added a `MAX_OBJECTS_PER_REQUEST` to limit the amount of data we pull from FQS. This is just the beginning of some performance tuning there.

As a follow-up to this PR I intend to also do the following:
- Update document detail drawer to display clinical notes just like encounter drawer

### DIFF
--- a/src/fhir/document.ts
+++ b/src/fhir/document.ts
@@ -56,7 +56,7 @@ async function getDocumentFromFQS(requestContext: CTWRequestContext, patient: Pa
       {
         upid: patient.UPID,
         cursor: "",
-        first: 50,
+        first: 1000,
         sort: {
           lastUpdated: "DESC",
         },
--- a/src/fhir/document.ts
+++ b/src/fhir/document.ts
@@ -56,7 +56,7 @@ async function getDocumentFromFQS(requestContext: CTWRequestContext, patient: Pa
       {
         upid: patient.UPID,
         cursor: "",
-        first: 1000,
+        first: 50,
         sort: {
           lastUpdated: "DESC",
         },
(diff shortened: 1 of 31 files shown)


src/fhir/document.ts
```diff
@@ -56,7 +56,7 @@ async function getDocumentFromFQS(requestContext: CTWRequestContext, patient: Pa
       {
         upid: patient.UPID,
         cursor: "",
-        first: 1000,
+        first: 50,
```
bryanjknight, line 59: Bit concerned that reducing the number of documents in the FQS request is going to cause issues with user adoption with "missing data". Could we try 500 to start?

bryanjknight (approved):

PR #1222, opened 2023-09-20
CTW-1398 Remove version history from resource history drawer

eric-gunther:
See [this slack thread](https://zushealth.slack.com/archives/C03NWTVCGJV/p1695221829563379) for background on discussion with product/UX.

Essentially we have deemed that it is unnecessary to surface the version history *of a single FHIR resource* in the history section of the details drawer. The primary intent of this section is to highlight what other organizations have participated in care, so being able to show something like the status transitions of a condition (eg. when it went from active to inactive) is not necessary. In an effort to simplify the code and the logic behind this feature, reduce calls to the back-end, and address the bug in CTW-1398 where we're showing "Unknown" as the title for older versions of a condition resource, we are removing this functionality.

### DIFF
--- a/src/components/content/resource/history.ts
+++ b/src/components/content/resource/history.ts
@@ -1,4 +1,3 @@
-import { GraphQLClient } from "graphql-request";
 import { HistoryEntryProps } from "./helpers/history-entry";
 import { CTWRequestContext } from "@/components/core/providers/ctw-context";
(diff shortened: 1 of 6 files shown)


src/components/content/resource/history.ts
pmadria3, line 1: AllergyIntolerance also uses the versionHistory so will need to modify that as well

pmadria3 (approved):

PR #1230, opened 2023-09-21
Resource Drawer Extensibility

melocule:
Preparation for CDEV-472
Allows a function to be passed into the drawer hook that renders a child component, condensing a couple hard-coded function props.


melocule:
Tested on Bruno, renderChild appears to work and the UI looks the same as before.

image

pmadria3 (approved):

PR #1231, opened 2023-09-21
CTW-1388 Enable semantic and generative search results

eric-gunther:
Updating search so that we include the semantic search if the search phrase has more than 2 words. If the user ends with a question mark we'll also include the AI generated response. This behavior is subject to change in the future, just doing this as a short term way of making it available in the app.

Also fixing an issue with how we retrieve documents from FQS by ID.

### DIFF
--- a/src/fhir/document.ts
+++ b/src/fhir/document.ts
@@ -83,7 +83,9 @@ export async function getDocumentsByIdFromFQS(
         cursor: "",
         first: 1000,
         filter: {
-          id: ["in", ids],
+          ids: {
+            anymatch: ids,
+          },
         },
         sort: {
           lastUpdated: "DESC",
--- a/src/services/patient-record-search/patient-record-search.ts
+++ b/src/services/patient-record-search/patient-record-search.ts
@@ -214,7 +214,7 @@ export function usePatientRecordSearch(
           query: searchTerm,
           upid: patient.UPID,
           include: searchOptions,
-          n_results: includeAnswer ? 4 : 10,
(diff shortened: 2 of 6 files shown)


src/fhir/document.ts
```diff
@@ -83,7 +83,9 @@ export async function getDocumentsByIdFromFQS(
         cursor: "",
         first: 1000,
         filter: {
-          id: ["in", ids],
+          ids: {
+            anymatch: ids,
+          },
```
eric-gunther, line 88: Previously we were fetching all documents when searching by ID!

src/services/patient-record-search/patient-record-search.ts
```diff
@@ -196,14 +197,24 @@ export function usePatientRecordSearch(
...
+          searchOptions.push("semantic");
+        }
+
         const body = JSON.stringify({
           query: searchTerm,
           upid: patient.UPID,
-          include: ["keyword"], // , "semantic"],
-          n_results: 10,
+          include: searchOptions,
+          n_results: includeAnswer ? 4 : 10,
```
eric-gunther, line 217: This is a bit of a hack to be kinder to the API when requesting the generative responses, will likely tweak this in the future.

mrosata (approved):

src/fhir/document.ts
mrosata: Why was it getting all documents? Does the "in" operator not test the `id` is in `ids`?

PR #1232, opened 2023-09-22
CTW-1403 Moving search to a tab in the ZAP

eric-gunther:
Moving search out of the ZAP Title bar (which is typically disabled and we may even want to consider removing altogether) and into a dedicated tab. This kind of hides it in the "More" menu but that's OK for the time being while it's feature flagged and experimental. UX is going to work on a longer term plan for this feature while we use this as a way of simply getting it available in the ZAP.

### DIFF
--- a/src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
+++ b/src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
@@ -61,7 +61,6 @@ export type ZusAggregatedProfileProps = {
   forceHorizontalTabs?: boolean;
   includePatientDemographicsForm?: boolean;
   title?: string;
-  includePatientRecordSearch?: boolean;
   hideTitle?: boolean;
   removeBranding?: boolean;
   removeRequestRecords?: boolean;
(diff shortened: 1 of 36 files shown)


src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -108,16 +104,9 @@ const ZusAggregatedProfileComponent = ({
...
-  const [patientRecordSearchIsOpen, setPatientRecordSearchIsOpen] = useState(false);
-  const [zapTitle, setZapTitle] = useState(title);
-  const togglePatientRecordSearchIsOpen = () => {
-    setPatientRecordSearchIsOpen(!patientRecordSearchIsOpen);
-    setZapTitle(patientRecordSearchIsOpen ? title : "Search");
-    trackInteraction("toggle_ai_search", {
-      value: patientRecordSearchIsOpen ? "close" : "open",
-    });
-  };
+  const [zapTitle, _] = useState(title);
```
mrosata, line 107: If we're not using a setter this variable `zapTitle` isn't needed

mrosata (approved):

PR #1252, opened 2023-09-26
CTW-1404 Refactoring diagnostic report queries

eric-gunther:
We are currently making some very expensive calls to retrieve diagnostic reports. Every time we load the ZAP (in order to show the red dot indicating there are unread diagnostic reports) we retrieve up to 1000 observations and up to 1000 diagnostic reports with references to patients/practitioner/observations/medication-requests, and we do that second query TWICE!

This PR refactors our diagnostic reports data retrieval into 2 hooks - `usePatientDiagnosticReports` with options for a limit and whether or not you include observations. We will use this hook with a limit of 20 and not including observations when retrieving data to inform the red dot. Basically this means that if any of the 20 most recently updated resources are unread then show the red dot. This feels like a good-enough answer for "is there any data for me to look at".

The second hook is `usePatientDiagnosticReportsWithTrendData` which takes in a limit as well. This will query both observations and diagnostic reports and is still somewhat expensive, but at least now we're only querying diagnostic report data once. I left in a lot of the logic that re-constructs DiagnosticReportModels with observation trend arrays which themselves contain DiagnosticReportModels as links to their parent diagnostic reports. That's all pretty intensive and could maybe use refactoring in the future, but for now at least the network calls are less intensive.

This PR also refactors out some FQS feature flagging that's no longer necessary and reduces our FQS query limit to 500 from 1000.


src/fhir/diagnostic-report.ts
```diff
@@ -1,141 +1,146 @@
...
 
-  const query = useQueryWithPatient(
-    QUERY_KEY_OTHER_PROVIDER_DIAGNOSTIC_REPORTS,
-    [data?.map((o) => o.id)], // Only use the IDs in our key (fixes issue with ciruclar references).
-    withTimerMetric(
-      diagnosticReportsFetcherFQS("all", data ?? []),
-      "req.timing.all_diagnostic_reports"
-    )
-  );
+  useEffect(() => {
```
eric-gunther, line 45: This `useEffect` is important! Without this we had been calling the diagnostic reports query twice - once on load and then again after the observations query finished.

src/fhir/observations.ts
```diff
@@ -2,29 +2,21 @@ import { ObservationModel, PatientModel } from "./models";
...
-      trendingLabs.enabled && patient.createdAt && patient.createdAt >= "2023-07-19"
-        ? withTimerMetric(fetchObservationsTrendData(loincCodes), "req.timing.all_observations")(
-            requestContext,
-            patient
-          )
-        : []
+      withTimerMetric(fetchObservationsTrendData(loincCodes), "req.timing.all_observations")(
+        requestContext,
+        patient
+      )
```
eric-gunther, line 19: I'm assuming this FQS back-fill is done now.

src/services/fqs/client.ts
```diff
@@ -20,7 +20,7 @@ export interface GenericConnection<T extends ResourceTypeString> {
   edges: GraphqlConnectionNode<ResourceType<T>>[];
 }
 
-export const MAX_OBJECTS_PER_REQUEST = 1000;
+export const MAX_OBJECTS_PER_REQUEST = 500;
```
eric-gunther, line 23: This is pretty unscientific. But we're currently retrieving up to 1000 resources per patient and we're doing it often. This seems like overkill (though some patients do have a ton of resources). Perhaps we can vary this more by resource type?

src/services/fqs/queries/diagnostic-reports-query.ts
```diff
@@ -55,12 +66,6 @@ export const diagnosticReportQuery = gql`
             url
             valueString
           }
-          basedOn {
-            reference
-            resource {
-              ...MedicationRequest
-            }
-          }
```
eric-gunther, line 63: No idea why this was here but I could not come up with a use-case (or any reference in the code) to justify retrieving medication requests along with our diagnostic reports. And this was a very expensive thing to include in this query.

src/services/fqs/queries/diagnostic-reports-query.ts
```diff
@@ -93,18 +98,7 @@ export const diagnosticReportQuery = gql`
...
-            display
-            resource {
-              ...Observation
-            }
-          }
-          contained {
-            resource {
-              ...Observation
-            }
-          }
```
eric-gunther, line 107: These observations are quite expensive to include as well, really only need them when rendering the table, don't need them for the red dots.

src/fhir/observations.ts
bryanjknight: It's sadly isn't :(

bryanjknight (changes requested):

src/fhir/observations.ts
eric-gunther: Added back the hack

src/services/fqs/client.ts
eric-gunther: 100% agree that lens & pagination is the right long term solution. 500 still seems like a pretty reasonably large number while reducing the amount of memory we're using.

src/services/fqs/queries/diagnostic-reports-query.ts
eric-gunther: Still including all of this when you render the full table, it's just now conditional based on an `includeObservations` argument supplied when getting the query for diagnostic reports. This makes it so we don't request them when rendering the red dot.

bryanjknight (approved):

pmadria3 (approved):

PR #1256, opened 2023-09-27
CTW-1404 Updating care team component

eric-gunther:
Updating care team component so that it uses our generic resource table and has sorting options. Cleaned up some of the columns and details drawer, and added de-duping logic by care team provider name. Ran all this by product/UX already.

Also tacking on small fix for CTW-1424, filtering immunizations by completed. This was our only blocker from removing the beta label for immunizations.


src/fhir/immunizations.ts
```diff
@@ -47,3 +49,18 @@ async function getImmunizationFromFQS(requestContext: CTWRequestContext, patient
...
+    immunizations.filter((x) => x.status === "completed"),
+    "occurrence",
+    "desc",
+    true
+  );
+
+  // Bump builder owned allergies to the front, so uniqBy favors them!
+  const builderOwnedFirst = orderBy(sortedByDate, (a) => a.ownedByBuilder(builderId), "desc");
+  return uniqBy(builderOwnedFirst, "uniqueKey");
+};
```
eric-gunther, line 66: Only new logic here is line 57, filtering on completed. Otherwise just moved this out of the filters.tsx file into this file where it seems more appropriate.

src/components/content/care-team/helpers/columns.tsx
```diff
@@ -3,9 +3,9 @@ import { CareTeamPractitionerModel } from "@/fhir/models/careteam-practitioner";
 
 export const patientCareTeamColumns: TableColumn<CareTeamPractitionerModel>[] = [
   {
-    widthPercent: 20,
+    widthPercent: 30,
```
pmadria3, line 6: The widthPercents should be adding to 100

eric-gunther: They do (30+30+20+20) - they weren't before.

pmadria3 (approved):

src/components/content/care-team/helpers/columns.tsx
pmadria3: Ah gotcha I didn't see the last 20

PR #1260, opened 2023-09-29
CTW-1445 First pass on new overview tab

eric-gunther:
Created a new "Overview" tab that will only be visible if you have patient record search enabled. This will serve as our first pass at a summary/overview for the patient while also helping to make search more prominent for beta customers. The content of this tab will likely change a lot in the near future - next up we may surface the patient's subscription info. Discussed/reviewed with product & UX.

https://github.com/zeus-health/ctw-component-library/assets/97455910/2b1f7555-e551-44c1-9679-16cdf082f229

### DIFF
--- a/src/components/content/overview/patient-overview.tsx
+++ b/src/components/content/overview/patient-overview.tsx
@@ -1,0 +1,39 @@
+import { PatientHistoryAction } from "../patient-history/patient-history-action";
+import { PatientRecordSearch } from "../patient-record-search/patient-record-search";
+
+export const Overview = () => (
+  <div className="ctw-flex ctw-w-full ctw-flex-row">
+    <div className="ctw-mx-2 ctw-my-5 ctw-h-fit ctw-basis-2/6 ctw-rounded-lg ctw-border-solid ctw-border-divider-light">
+      <div className="ctw-space-y-2 ctw-px-2 ctw-py-2">
+        <div className="ctw-px-3 ctw-pt-2">
+          <PatientHistoryAction />
+        </div>
+        <div className="ctw-border-0 ctw-border-t ctw-border-solid ctw-border-divider-light ctw-px-3 ctw-pt-3">
+          <div className="ctw-pb-2">
+            <a
+              href="https://share.hsforms.com/10NmAPiu7Qd2OXbxAHnseiwcjai8"
+              className="ctw-text-primary-main"
+              target="_blank"
+              rel="noreferrer"
+            >
+              Share ZAP Feedback
+            </a>
+          </div>
+          <div>
+            <a
+              href="https://docs.zushealth.com/changelog"
+              className="ctw-text-primary-main"
+              target="_blank"
+              rel="noreferrer"
+            >
+              What&apos;s New?
+            </a>
+          </div>
+        </div>
+      </div>
+    </div>
+    <div className="ctw-basis-4/6 ctw-pr-2">
+      <PatientRecordSearch />
+    </div>
+  </div>
+);
--- a/src/components/content/overview/patient-overview.tsx
+++ b/src/components/content/overview/patient-overview.tsx
@@ -1,7 +1,8 @@
 import { PatientHistoryLastRetrievedWithAction } from "../patient-history/patient-history-action";
 import { PatientRecordSearch } from "../patient-record-search/patient-record-search";
+import { withErrorBoundary } from "@/components/core/error-boundary";
 
-export const Overview = () => (
+export const PatientOverviewComponent = () => (
   <div className="ctw-flex ctw-w-full ctw-flex-row">
     <div className="ctw-mx-2 ctw-my-5 ctw-h-fit ctw-basis-2/6 ctw-rounded-lg ctw-border ctw-border-solid ctw-border-divider-light">
       <div className="ctw-space-y-2 ctw-px-2 ctw-py-2">
@@ -37,3 +38,5 @@ export const Overview = () => (
     </div>
   </div>
 );
+
+export const PatientOverview = withErrorBoundary(PatientOverviewComponent, "PatientOverview");
--- a/src/components/content/overview/patient-overview.tsx
+++ b/src/components/content/overview/patient-overview.tsx
@@ -1,18 +1,18 @@
-import { PatientHistoryAction } from "../patient-history/patient-history-action";
+import { PatientHistoryLastRetrievedWithAction } from "../patient-history/patient-history-action";
 import { PatientRecordSearch } from "../patient-record-search/patient-record-search";
 
 export const Overview = () => (
   <div className="ctw-flex ctw-w-full ctw-flex-row">
-    <div className="ctw-mx-2 ctw-my-5 ctw-h-fit ctw-basis-2/6 ctw-rounded-lg ctw-border-solid ctw-border-divider-light">
+    <div className="ctw-mx-2 ctw-my-5 ctw-h-fit ctw-basis-2/6 ctw-rounded-lg ctw-border ctw-border-solid ctw-border-divider-light">
       <div className="ctw-space-y-2 ctw-px-2 ctw-py-2">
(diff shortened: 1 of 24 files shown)


src/components/content/patient-history/patient-history-action.tsx
```diff
@@ -2,12 +2,16 @@ import { PatientHistoryLastRetrieved } from "./patient-history-last-retrieved";
...
-  </div>
+export const PatientHistoryLastRetrievedWithAction = ({
+  hideRequestRecords = false,
+}: PatientHistoryActionProps) => (
+  <>
+    <div>
+      <PatientHistoryLastRetrieved />
+    </div>
+    <div>{!hideRequestRecords && <RequestRecordsButton />}</div>
+  </>
```
eric-gunther, line 16: This is only used elsewhere in the patient conditions outside component which has been deprecated and will be removed completely soon anyway, so changing the layout doesn't have any unintended affects.

src/components/content/patient-history/patient-history-last-retrieved.tsx
```diff
@@ -12,23 +12,22 @@ export const PatientHistoryLastRetrieved = () => {
     return <div />;
   }
 
-  const { lastRetrievedAt } = patientHistory;
+  let { lastRetrievedAt } = patientHistory;
+
+  // if this is a test patient with no last retrieved date then just show today
+  if (!lastRetrievedAt && patient.data?.isTestPatient) {
+    lastRetrievedAt = Date.now();
+  }
```
eric-gunther, line 20: Instead of saying nothing for test patients making it say that records were last retrieved on the current date.

src/components/core/buttons.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 .ctw-btn-clear {
```
eric-gunther, line 13: This class was often fighting with other classes to define font size.

src/utils/query-keys.ts
```diff
@@ -22,6 +22,7 @@ export const QUERY_KEY_PATIENT_HISTORY_DETAILS = "patient_history_details";
 export const QUERY_KEY_PATIENT_HISTORY_LIST = "patient_history_list";
 export const QUERY_KEY_PATIENT_IMMUNIZATIONS = "patient_immunizations";
 export const QUERY_KEY_PATIENT_OBSERVATIONS = "patient_observations";
+export const QUERY_KEY_PATIENT_SUBSCRIPTION = "patient_subscription";
```
eric-gunther, line 25: A bit forward thinking here, going to include this query key in an upcoming PR to get patient subscription information. Had started that in this PR but then thought it was getting unwieldy so pulled it out and looks like I left this const.

src/components/content/overview/patient-overview.tsx
```diff
@@ -0,0 +1,39 @@
+import { PatientHistoryLastRetrievedWithAction } from "../patient-history/patient-history-action";
+import { PatientRecordSearch } from "../patient-record-search/patient-record-search";
+
+export const Overview = () => (
```
WesleyKapow, line 4: **BLOCKER**

1. Should be `PatientOverview`?
2. Should follow the `withErrorBoundary` pattern the other components use. This is to guard the case where this component is used directly and not within the ZAP which has its own error boundary.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
```diff
@@ -50,14 +51,20 @@ import {
...
-export type ZusAggregatedProfileTabs = Record<ZAPResourceName, (props: object) => TabGroupItem>;
+export type ZusAggregatedProfileTabs = Record<ZAPTabName, (props: object) => TabGroupItem>;
 
 export const zusAggregatedProfileTabs: ZusAggregatedProfileTabs = {
+  overview: () => ({
+    key: "overview",
+    display: () => <TabHelper name="overview" beta />,
+    render: () => <Overview />,
+  }),
+
```
WesleyKapow, line 67: May not be worth it, but seems like guarding these tabs with feature flags will be a common pattern. You could refactor to add another key here like `featureToggle: 'ctw-patient-record-search'` and then have a single place that filters out tabs by the toggles. This may require some tweaks to how we use unleash as you wouldn't be able to do `useFeatureToggle(...)` and would instead need something like `unleash.isEnabled(...)` again so that it could be dynamic within the filtering code.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -97,14 +99,13 @@ const ZusAggregatedProfileComponent = ({
...
-  resources = defaultZAPResources,
+  resources = defaultZAPTabs,
   hideTitle = false,
   title = "Outside Records",
   removeBranding = false,
   removeRequestRecords = false,
 }: ZusAggregatedProfileProps) => {
-  const unleash = useUnleashClient();
-  const isSearchEnabled = unleash.isEnabled("ctw-patient-record-search");
+  const isSearchEnabledToggle = useFeatureToggle("ctw-patient-record-search");
```
WesleyKapow, line 108: General refactor/cleanup or driven by something else?

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
```diff
@@ -166,7 +170,7 @@ const ZusAggregatedProfileComponent = ({
           content={tabbedContent}
           forceHorizontalTabs={forceHorizontalTabs}
           topRightContent={
-            <RenderIf condition={!removeRequestRecords}>
+            <RenderIf condition={!removeRequestRecords && isSearchEnabledToggle.enabled === false}>
```
WesleyKapow, line 173: This previously showed up on all tabs right? Now we show it on overview and no where else? Seems reasonable, just making sure I understand the change.

src/components/content/overview/patient-overview.tsx
```diff
@@ -0,0 +1,39 @@
...
+  <div className="ctw-flex ctw-w-full ctw-flex-row">
+    <div className="ctw-mx-2 ctw-my-5 ctw-h-fit ctw-basis-2/6 ctw-rounded-lg ctw-border ctw-border-solid ctw-border-divider-light">
+      <div className="ctw-space-y-2 ctw-px-2 ctw-py-2">
```
WesleyKapow, line 7:
```suggestion
      <div className="ctw-space-y-2 ctw-p-2">
```

WesleyKapow (changes requested):
Looks awesome, few nits/questions and one blocker.

src/components/content/zus-aggregated-profile/zus-aggregated-profile-tabs.tsx
eric-gunther: Yeah this is a fair point, but my gut is that it may be more effort than it's worth as a refactor right now. I can give it a deeper look on my second pass at this feature (adding more info to the overview tab).

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
eric-gunther: I was seeing weird behavior due to the ZAP rendering without the overview (because the flags hadn't come back from Unleash yet) and then again after the flags returned. This wasn't a big deal when search was just getting appended to the end, but having the overview as the first tab made it a problem. So the ZAP is instead showing loading (which is really just a split second, you don't even notice it) when the feature flags are loading. Adding that "ready" check is part of `useFeatureToggle`.

src/components/content/zus-aggregated-profile/zus-aggregated-profile.tsx
eric-gunther: Exactly.

WesleyKapow (approved):

mrosata (approved):

PR #1274, opened 2023-10-03
Add filtering by location to the filters

pmadria3:
CDEV-467

This PR adds the ability to filter by physicalType if there is a value for it.
Screenshot 2023-10-04 at 10 03 05 PM


src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,21 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
     return uniqueLocations.length ? uniqueLocations.join(", ") : undefined;
   }
 
+  get physicalType() {
+    const locations = compact(
+      this.resource.location?.map((l) => {
+        const location = findReference("Location", undefined, undefined, l.location);
```
WesleyKapow, line 126: Is this always going to use the special FQS provided `l.location.resource`? If so, I say use that instead of jumping through `findReference`.

src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,21 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
     return uniqueLocations.length ? uniqueLocations.join(", ") : undefined;
   }
 
+  get physicalType() {
+    const locations = compact(
```
WesleyKapow, line 124: s/locations/physicalTypes

src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,21 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
...
+      this.resource.location?.map((l) => {
+        const location = findReference("Location", undefined, undefined, l.location);
+        return codeableConceptLabel(location?.physicalType);
```
WesleyKapow, line 127: Can you verify this works from ticket, I'd expect some special handling for this.
> Those with a code value of URGENT_CARE should have “Urgent Care” in the display value.  This may have to be hard coded in the short term.

src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,21 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
...
+      })
+    );
+
+    const uniqueLocations = uniq(locations).filter(
+      (l) => l !== "Not Indicated" && l !== "Unknown" && l !== "NoInformation"
+    );
+
+    return uniqueLocations.length ? uniqueLocations.join(", ") : undefined;
```
WesleyKapow, line 135: hmmm if this encounter has several.. will filtering still work correctly? Does selecting "hospital" correctly filter out "rehab hospital"?

WesleyKapow (dismissed):
Add screenshot :D!

src/fhir/models/encounter.ts
pmadria3: It is but the problem is we will have to disable typescript since it doesn't follow the FHIR feature (we use this in other places as well especially for `Organization` types.

src/fhir/models/encounter.ts
pmadria3: I'm not sure if there will be several though - I will ask but I'm not entirely sure how that's possible to have several values for a single encounter

pmadria3: So we only ever get one value - asked product

src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,33 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
...
+        case "Unknown":
+        case "NoInformation":
+        case "No Information":
+        case "Building":
+        case undefined:
+          return "Unknown";
+        default:
+          return l;
+      }
+    });
```
eric-gunther, line 145: This feels a bit hack and like the kinda thing that should be handled by enrichment. Not blocking feedback but I'm wary.

src/fhir/models/encounter.ts
```diff
@@ -120,6 +120,33 @@ export class EncounterModel extends FHIRModel<fhir4.Encounter> {
     return uniqueLocations.length ? uniqueLocations.join(", ") : undefined;
   }
 
+  get physicalTypes() {
```
eric-gunther, line 123: "Physical Types" is confusing term for me here. These are the "location types" of the encounter or "location physical types" of the encounter. You use the term "Location Type" elsewhere so not sure why you wouldn't just use that for this model getter too.

eric-gunther (approved):

src/fhir/models/encounter.ts
pmadria3: So there is actually a type field and a physical type field but I can rename to locationPhysicalType

eric-gunther (approved):

PR #1281, opened 2023-10-04
CTW-1445 Showing subscription info in the patient overview

eric-gunther:
Adding patient subscription information to the overview page.

The most notable, potentially contentious, thing here is how we're deriving the subscription description/bullets from the package metadata. Will bring in DA for an opinion.

Pasted Graphic

### DIFF
--- a/src/components/content/overview/patient-subscription.tsx
+++ b/src/components/content/overview/patient-subscription.tsx
@@ -7,6 +7,26 @@ import {
 } from "@/services/subscriptions/subscriptions";
 import { compact, uniq } from "@/utils/nodash";
 
+const demoPatientSubscription = {
+  patientId: "123",
+  package: {
+    id: "st234",
+    description: "asdfasdf",
+    name: "ZAP Pro",
+    meta: {
+      freshmakerProviders: ["commonwell", "surescripts"],
+      initialProviders: ["commonwell", "surescripts"],
+      recurringProvidersWithInterval: [
+        {
+          intervalDays: 28,
+          provider: "commonwell",
+        },
+      ],
+      subscriptionProviders: ["bamboo", "collective", "quest", "surescripts"],
+    },
+  },
+};
+
 export const PatientSubscriptionDetails = () => {
   const patientSubscription = usePatientSubscription();
   const patient = usePatient();
@@ -19,7 +39,7 @@ export const PatientSubscriptionDetails = () => {
     return <ErrorAlert header="Error loading data" />;
   }
 
-  if (!patientSubscription.data.package) {
+  if (!patientSubscription.data.package && !patient.data.isTestPatient) {
     return (
       <>
         <span className="ctw-font-medium">{patient.data.display}</span> is not enrolled for any
@@ -29,13 +49,15 @@ export const PatientSubscriptionDetails = () => {
     );
   }
 
-  const dataSources = getDataSources(patientSubscription.data);
+  const dataSources = getDataSources(
+    patient.data.isTestPatient ? demoPatientSubscription : patientSubscription.data
+  );
 
   return (
     <>
       <div>
         <span className="ctw-font-medium">{patient.data.display}</span> is enrolled in{" "}
-        <span className="ctw-font-medium">{patientSubscription.data.package.name}</span>
+        <span className="ctw-font-medium">{patientSubscription.data.package?.name}</span>
       </div>
       <div className="ctw-pt-3">This provides access to the following data:</div>
       <ul className="ctw-mx-0 ctw-list-disc ctw-pl-4">
--- a/src/components/content/overview/patient-subscription.tsx
+++ b/src/components/content/overview/patient-subscription.tsx
@@ -50,7 +50,7 @@ export const PatientSubscriptionDetails = () => {
 };
 
 // Take subscription metadata and return a list of translated data sources and the details on their acquisition models
-const getDataSources = (patientSubscription: PatientSubscription) => {
+export const getDataSources = (patientSubscription: PatientSubscription) => {
   const translatedPackages = {
     freshmakerProviders:
       patientSubscription.package?.meta.freshmakerProviders?.map((x) =>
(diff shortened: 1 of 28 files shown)


src/utils/request.ts
```diff
@@ -26,7 +26,7 @@ export type CTWRequestInit = {
 
 export function ctwFetch(input: RequestInfo | URL, init?: CTWRequestInit): Promise<Response> {
   const headers = init?.headers || {};
-  headers["Zus-Request-Source"] = "component-library";
+  // headers["Zus-Request-Source"] = "component-library";
```
luthfib, line 29: Im guessing you don't want to do this anymore?

src/components/content/patient-history/patient-history-action.tsx
```diff
@@ -3,15 +3,22 @@ import { RequestRecordsButton } from "./request-records-button";
...
 export const PatientHistoryLastRetrievedWithAction = ({
   hideRequestRecords = false,
+  includePatientDemographicsForm = false,
 }: PatientHistoryActionProps) => (
   <>
     <div>
       <PatientHistoryLastRetrieved />
     </div>
-    <div>{!hideRequestRecords && <RequestRecordsButton />}</div>
+    {/* TODO - hide demographics! */}
```
luthfib, line 17: Leaving a comment here to ensure we come back to this

src/components/content/overview/patient-subscription.tsx
```diff
@@ -0,0 +1,122 @@
...
+        // @ts-ignore
+        i18next.exists(x) ? i18next.t(x) : x
+      ) ?? [],
+    recurringProvidersWithInterval:
+      patientSubscription.package?.meta.recurringProvidersWithInterval?.map((x) => ({
+        // @ts-ignore
+        provider: i18next.exists(x.provider) ? i18next.t(x.provider) : x.provider,
+        intervalDays: x.intervalDays,
+      })) ?? [],
+  };
```
luthfib, line 97: Curious why we had to have all these `ts-ignore`. Might have to pull down to understand the issue

eric-gunther: Yeah kinda ugly, the call to `i18next.t` causes a Typescript error even though it's within a conditional that checks to see if `x` exists in the translation dictionary.

eric-gunther: I added a comment explaining.

src/services/subscriptions/subscriptions.ts
```diff
@@ -0,0 +1,138 @@
...
+    description: string;
+    meta: PackageMeta;
+  };
+};
+
+export type PackageMeta = {
+  freshmakerProviders?: string[];
+  initialProviders?: string[];
+  recurringProvidersWithInterval?: {
+    intervalDays: number;
```
eric-gunther, line 22: What are the other interval values?

eric-gunther: Need to address this before merging

src/components/content/overview/patient-subscription.tsx
```diff
@@ -0,0 +1,124 @@
+import i18next from "i18next";
+import { Loading } from "@/components/core/loading";
+import { ErrorAlert, usePatient } from "@/index";
+import {
+  PatientSubscription,
+  usePatientSubscription,
+} from "@/services/subscriptions/subscriptions";
+import { compact, uniq } from "@/utils/nodash";
+
+const demoPatientSubscription = {
```
eric-gunther, line 10: Is this too hack? Setting up a mock package so that it looks better for test/demo patients.

src/fhir/diagnostic-report.ts
```diff
@@ -149,7 +149,7 @@ export async function fetchDiagnosticReportsFromFQSById(
     const graphClient = createGraphqlClient(requestContext);
     const { data } = await fqsRequest<DiagnosticReportGraphqlResponse>(
       graphClient,
-      getDiagnosticReportQuery(false),
+      getDiagnosticReportQuery(true),
```
eric-gunther, line 152: Ended up moving this change to its own PR, small fix.

src/components/content/overview/patient-subscription.tsx
luthfib: `isTestPatient` where does this property come from? Is this something the api returns? Generally don't like cases like this. From what I understand this data is particularly hard to mock

src/components/content/overview/patient-subscription.tsx
```diff
@@ -0,0 +1,124 @@
...
+};
+
+// Take subscription metadata and return a list of translated data sources and the details on their acquisition models
+export const getDataSources = (patientSubscription: PatientSubscription) => {
+  const translatedPackages = {
+    freshmakerProviders:
+      patientSubscription.package?.meta.freshmakerProviders?.map((x) =>
+        // All of these ts-ignore's are because the call to i18next.t has a TS error even though there is a check that x exists in the dictionary
+        // @ts-ignore
+        i18next.exists(x) ? i18next.t(x) : x
```
luthfib, line 81: You could define at type called like 

```export type Providers = "commonwell" | "surescripts" | "bamboo" | "collective" | "quest"```


and do `i18next.exists(x) ? i18next.t(x as Providers) : x`

Essentially what is going on it it knows all the types in `resources.en.main` but sees that these ones are strings and thus not matching these types. It's still slightly messy, but at least we could catch other ts errors

src/components/content/overview/patient-subscription.tsx
eric-gunther: It's a getter on the patient model checking for a test tag that's a pretty reliable indicator that the patient is for test/demo. The subscription and patient history data has definitely been challenging to mock and in pretty much all cases it's just empty and makes the demo somewhat confusing. It's certainly easier for us to mock the data close to the UI like this, and it doesn't need to be particularly intelligent (ie. it can always just show the same thing), but I wonder if this is introducing a bad pattern. I'll poke the team about it in Slack.

luthfib (approved):
LGTM.

PR #1282, opened 2023-10-04
CTW-1455 Observations not displaying in diagnostic search results

eric-gunther:


src/fhir/diagnostic-report.ts
```diff
@@ -149,7 +149,7 @@ export async function fetchDiagnosticReportsFromFQSById(
     const graphClient = createGraphqlClient(requestContext);
     const { data } = await fqsRequest<DiagnosticReportGraphqlResponse>(
       graphClient,
-      getDiagnosticReportQuery(false),
+      getDiagnosticReportQuery(true),
```
eric-gunther, line 152: This boolean informs whether or not you return observations in the query which has some performance impact but not significant when querying one diagnostic ID at a time like we do for search results.

luthfib (approved):

PR #1295, opened 2023-10-06
CTW-1445 Hiding subscription info until we get permissions sorted out

eric-gunther:
Care team users don't currently have the necessary permissions to view package metadata, so they get a 401 when we try to display patient subscriptions. Temporarily removing this functionality until we get the permissions sorted out. Also exporting the overview component and the subscription description component so we can use them in standalone.


mrosata (dismissed):
assuming the commented out code is temporary this lgtm

mrosata (approved):

PR #1302, opened 2023-10-07
add lint rules around react query

luthfib:
Add recommended `@tanstack/query` plugin to help ensure we keep up with best react query linting rules. Tested locally with a few patients to ensure queries looked like they were behaving the same.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -2,8 +2,8 @@
  * @type {import('@types/eslint').Linter.BaseConfig}
  */
 module.exports = {
-  extends: ["@kensho-technologies/eslint-config", "prettier"],
-  plugins: ["unused-imports", "sort-exports"],
+  extends: ["@kensho-technologies/eslint-config"],
+  plugins: ["unused-imports", "sort-exports", "@tanstack/query"],
   // Ignore js files as we now have typescript parsing rules.
   // See https://stackoverflow.com/a/65063702 for more.
   ignorePatterns: [
@@ -21,6 +21,9 @@ module.exports = {
     project: ["./tsconfig.json"],
   },
   rules: {
+    "@tanstack/query/exhaustive-deps": "error",
+    "@tanstack/query/prefer-query-object-syntax": "warn",
+    "@tanstack/query/stable-query-client": "error",
     "@typescript-eslint/await-thenable": "error",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-unnecessary-condition": "error",
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -3,7 +3,7 @@
  */
 module.exports = {
   extends: ["@kensho-technologies/eslint-config"],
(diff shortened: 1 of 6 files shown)


src/components/core/providers/patient-provider.tsx
```diff
@@ -155,19 +151,13 @@ export function useQueryWithPatient<T, T2>(
     );
   }
 
-  return useQuery(
-    [queryKey, patientResponse.data?.UPID, ...keys],
-    async () => {
+  return useQuery({ queryKey: [queryKey, patientResponse.data, ...keys], queryFn: async () => {
```
luthfib, line 154: This was done due to this rule https://tanstack.com/query/v4/docs/react/eslint/exhaustive-deps

.eslintrc.cjs
```diff
@@ -2,8 +2,8 @@
  * @type {import('@types/eslint').Linter.BaseConfig}
  */
 module.exports = {
-  extends: ["@kensho-technologies/eslint-config", "prettier"],
-  plugins: ["unused-imports", "sort-exports"],
+  extends: ["@kensho-technologies/eslint-config"],
+  plugins: ["unused-imports", "sort-exports", "@tanstack/query"],
```
mrosata, line 6: Only 1 question, why does "prettier" plugin go away?

src/components/core/providers/patient-provider.tsx
```diff
@@ -155,19 +156,17 @@ export function useQueryWithPatient<T, T2>(
     );
   }
 
-  return useQuery(
-    [queryKey, patientResponse.data?.UPID, ...keys],
-    async () => {
+  return useQuery({
+    queryKey: [queryKey, patientResponse.data, ...keys],
```
eric-gunther, line 160: Is this an intentional change to the query key? Before it was just on UPID, now it's on all data, and since patients can share UPIDs I think this will result in different behavior.

luthfib: I think we'd want this anyways? We could disable the rule here, but the idea is that anytime this data changes we'd want to cause a refetch. If patients share data, I think we'd want that data refetched. But could also just disable it for this instance.

mrosata (approved):

luthfib:
I had auto merge on, didnt mean to approve

src/components/core/providers/patient-provider.tsx
WesleyKapow: Hmmm this is specifically in the `useQueryWithPatient` hook. These queries are UPID based and so I think nothing about the `patientResponse.data` could change that would impact the results EXCEPT for a different upid.

WesleyKapow: Discussed with Luthfi in slack. Leaving this as is (entire patient response) as we currently pass the entire patient model to the `query` callback. This means that queries could, in theory, use any aspect of the patient model (e.g. there could be a query that uses the name, or age, etc.). So, using the entire patient resource in the keys is the safest approach.

PR #1310, opened 2023-10-10
Fix builder selector bug

luthfib:
Fix bug where updating props passed into ctwProvider didnt update the props the callback used.


WesleyKapow (approved):
Wow! I don't fully understand this but it's awesome you were able to figure out a solution to fix this bug!

PR #1313, opened 2023-10-10
Scope to upid patient query

luthfib:
Only re-run query when upid changes.


luthfib:
Closed this pr, due to the behavior actually being correct. Will revisit later if we need to. Essentially we techically use other properties in the model, so it should refresh if other properties change

PR #1314, opened 2023-10-10
Fix Row Actions for Encounters

melocule:
A couple fixes.
- Row actions persist across refreshes in Encounters & Notes. ([Video](https://github.com/zeus-health/ctw-component-library/assets/33408946/ca57f01a-bb7b-4518-90f6-c6ec2e4c92be))
- "Show Dismissed Records" button added to Encounters & Notes table. ([Video](https://github.com/zeus-health/ctw-component-library/assets/33408946/4e09d3bf-efa5-4ed6-b308-2e024a71d4be))

This solution calls to helper functions we didn't use for Encounters, but that we call for the resources.

### DIFF
--- a/src/fhir/models/fhir-model.ts
+++ b/src/fhir/models/fhir-model.ts
@@ -9,7 +9,7 @@ import {
   SYSTEM_ZUS_UNIVERSAL_ID
 } from "../system-urls";
 import { isFHIRDomainResource, ResourceMap, ResourceTypeString } from "../types";
-import { find, isEqual, orderBy, some, startCase } from "@/utils/nodash";
+import { find, orderBy, some, startCase } from "@/utils/nodash";
 
 export abstract class FHIRModel<T extends fhir4.Resource> {
   public resource: T;
--- a/src/fhir/models/fhir-model.ts
+++ b/src/fhir/models/fhir-model.ts
@@ -6,7 +6,7 @@ import {
   SYSTEM_ZUS_OWNER,
   SYSTEM_ZUS_PROFILE_ACTION,
   SYSTEM_ZUS_THIRD_PARTY,
-  SYSTEM_ZUS_UNIVERSAL_ID,
+  SYSTEM_ZUS_UNIVERSAL_ID
 } from "../system-urls";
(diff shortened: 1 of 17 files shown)


eric-gunther (dismissed):

src/fhir/timeline-event.ts
```diff
@@ -14,7 +14,7 @@ export function useTimelineEvents() {
 
   useEffect(() => {
     const models = compact(
-      flatten(concat(queries.map((query) => query.data?.map((m) => new TimelineEventModel(m)))))
+      flatten(concat(queries.map((query) => query.data.map((m) => new TimelineEventModel(m)))))
```
mrosata, line 17: I'm not sure that concat is needed here
```suggestion
      flatten(queries.map((query) => query.data.map((m) => new TimelineEventModel(m))))
```

src/fhir/models/fhir-model.ts
```diff
@@ -72,12 +72,14 @@ export abstract class FHIRModel<T extends fhir4.Resource> {
...
-      system: SYSTEM_ZUS_PROFILE_ACTION,
-      code: "read",
-    });
+    const isRead = some(basic?.code.coding, (coding) =>
+      isEqual(coding, {
+        system: SYSTEM_ZUS_PROFILE_ACTION,
+        code: "read",
+      })
+    );
+    return isRead;
```
WesleyKapow, line 82: This looks identical to what was here before, why the change?

WesleyKapow (approved):
LGTM with one minor comment

src/fhir/models/fhir-model.ts
melocule: Ahh the documentation on the func said the second arg had to be a predicate and didn't mention the 'matches' shorthand, so I found the old version confusing. I can change it back now!

WesleyKapow: Ya we use that shorthand A LOT. You can also verify something isn't working before fixing it (which, like here, sometimes shows you that it actually does work, and there's some misunderstanding at play).

WesleyKapow (approved):
LGTM

Re videos -- easier for reviewer if you have them inline in the description so we don't have to download them.

PR #1316, opened 2023-10-12
CTW-1480 hospitalizations row actions

melocule:
CTW-1480
Add hospitalization table row actions

https://github.com/zeus-health/ctw-component-library/assets/33408946/a838c584-7393-4a6c-bdb4-71d58666ba62

Solution: Needed a new function to get the basics outside of PatientProvider. 

Additionally: Tweaks the styling on button in the drawer that sends you to the patient, based on designer feedback. [CTW-1476](https://zeushealth.atlassian.net/browse/CTW-1476)

TODO: 
- Fetches again once you switch to another page for some reason, need to figure out why
- No query keys, so it refetches upon refresh. May have to do useQuery[...] inside a loop and add a line to ignore the ESLint rule for that.

[CTW-1476]: https://zeushealth.atlassian.net/browse/CTW-1476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/content/patients-adt/patients-adt-alerts-table.tsx
+++ b/src/components/content/patients-adt/patients-adt-alerts-table.tsx
@@ -1,7 +1,6 @@
 import type { TableColumn } from "@/components/core/table/table-helpers";
 import type { PatientModel } from "@/fhir/models/patient";
 import cx from "classnames";
-import { Basic } from "fhir/r4";
 import { useEffect, useState } from "react";
 import { adtFilter, defaultADTFilters } from "./filters";
 import { useADTAlertDetailsDrawer } from "./modal-hooks";
@@ -18,7 +17,7 @@ import { AnalyticsProvider } from "@/components/core/providers/analytics/analyti
 import { CTWRequestContext } from "@/components/core/providers/ctw-context";
 import { useCTW } from "@/components/core/providers/use-ctw";
 import { SimpleMoreList } from "@/components/core/simple-more-list";
-import { fetchBasicsFor } from "@/fhir/basic";
+import { enrichWithBasics } from "@/fhir/basic";
 import { EncounterModel } from "@/fhir/models/encounter";
 import { useFilteredSortedData } from "@/hooks/use-filtered-sorted-data";
 import { createGraphqlClient, fqsRequest } from "@/services/fqs/client";
@@ -48,21 +47,6 @@ export type ADTTableProps = {
   goToPatient: (upid: string) => void;
 } & TableOptionProps<EncounterModel>;
 
-async function enrichBasics(
-  adtEncounters: EncounterModel[],
-  getRequestContext: () => Promise<CTWRequestContext>
-) {
-  const promiseArray: Promise<Basic[]>[] = adtEncounters.map(async (enc) => {
-    const requestContext = await getRequestContext();
-    return fetchBasicsFor(requestContext, enc);
-  });
-  return Promise.all(promiseArray).then((basicsArray) => {
-    adtEncounters.forEach((e, i) => {
-      e.revIncludes = basicsArray[i];
-    });
-  });
-}
-
 function assignEncountersAndNotes(
   adtEncounters: EncounterModel[],
   encounterAndNotesData: RelatedEncounterMap,
@@ -85,13 +69,13 @@ function assignEncountersAndNotes(
             cursor: "",
             first: 1,
             sort: {
-              lastUpdated: "DESC"
+              lastUpdated: "DESC",
             },
             filter: {
               ids: {
-                anymatch: [encAndNote.cwcq_encounter_id]
-              }
-            }
+                anymatch: [encAndNote.cwcq_encounter_id],
+              },
+            },
           });
         },
         { staleTime: RELATED_ENC_STALE_TIME }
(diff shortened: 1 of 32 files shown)


src/hooks/use-filtered-sorted-data.ts
```diff
@@ -11,13 +11,15 @@ export type UseFilteredSortedDataProps<T extends object> = {
   defaultSort?: SortOption<T>;
   defaultFilters?: FilterChangeEvent;
   defaultView?: ViewOption<T>;
+  isLoading?: boolean;
```
WesleyKapow, line 14: At first I thought this made sense but then I was curious why it worked before. It already has the `records` in the dependencies array for the `useEffect` so I'm surprised `isLoading` is needed here. What issue did you run into?

src/components/content/patients-adt/patients-adt-alerts-table.tsx
```diff
@@ -56,11 +57,12 @@ function assignEncountersAndNotes(
       return;
     }
     const encAndNote = encounterAndNotesData.get(e.resource.id ?? "");
+
     if (encAndNote) {
+      const requestContext = await getRequestContext();
```
WesleyKapow, line 62: Why pull this up? Not sure which is better but I suspect keeping it within the query function is good as `getRequestContext` will refresh auth token if needed. So good to always run that code when making a query.

src/components/content/patients-adt/patients-adt-alerts-table.tsx
```diff
@@ -99,6 +102,11 @@ function ADTTableComponent({
   const [currentPage, setCurrentPage] = useState(1);
   const { getRequestContext } = useCTW();
   const { past7days, past30days, past3months } = getDateRangeView<EncounterModel>("periodStart");
+  const { data: encWithBasics, isLoading: isLoadingBasics } = useBasicsOf(
```
WesleyKapow, line 105: nit: better to not create abbreviations like `enc` and instead go with more verbose `encountersWithBasics`. This reduces cognitive load (no "what does enc mean?" thoughts)

src/fhir/basic.ts
```diff
@@ -146,3 +146,42 @@ async function fetchBasic(requestContext: CTWRequestContext) {
...
+    return fetchBasicsFor(requestContext, enc);
+  });
+  return Promise.all(promiseArray).then((basicsArray) => {
+    models.forEach((e, i) => {
+      e.revIncludes = basicsArray[i];
+    });
+  });
+}
+
+export function useBasicsOf<R extends fhir4.Resource, T extends FHIRModel<R>>(
```
WesleyKapow, line 173: nit: favor exported functions at top of file and helpers below. This is totally subjective but I find it easier to see the "interface" a file provides when the exports are at the top so I can quickly scan the available functions.

src/fhir/basic.ts
```diff
@@ -146,3 +146,42 @@ async function fetchBasic(requestContext: CTWRequestContext) {
...
+  models: T[],
+  key: string
+) {
+  const { getRequestContext } = useCTW();
+  return useQueryWithCTW(
+    QUERY_KEY_BASIC,
+    ["hospitalizations", key],
```
WesleyKapow, line 180: **BLOCKER:** This should not have `hospitalizations` in the key.

src/components/content/patients-adt/patients-adt-alerts-table.tsx
```diff
@@ -99,6 +102,11 @@ function ADTTableComponent({
...
+    data,
+    `Page ${currentPage}`
```
WesleyKapow, line 107: Maybe this should provide the key of `Hospitalizations page #`?

src/fhir/basic.ts
```diff
@@ -146,3 +146,42 @@ async function fetchBasic(requestContext: CTWRequestContext) {
...
+    });
+  });
+}
+
+export function useBasicsOf<R extends fhir4.Resource, T extends FHIRModel<R>>(
+  models: T[],
+  key: string
+) {
+  const { getRequestContext } = useCTW();
+  return useQueryWithCTW(
```
WesleyKapow, line 178: This may make more sense if you use this to wrap the part that's doing the fetching, so move it to `fetchBasicsFor`.. then the key would be `QUERY_KEY_BASIC, [resource.id]`.

Also, now looking at it, two thoughts:
1. You are doing N queries (one for each resource instead of a single query that grabs all basic resources).
2. Check with FQS & team here, we want to use FQS **and not ODS/smile**

WesleyKapow (changes requested):
Handful of comments, also I'm confused, where did the enabling of the row actions come in?

src/hooks/use-filtered-sorted-data.ts
melocule: other tables get the main resources with the basics together, but the hospitalizations component is given the main resources by CTW and then I have it assign the basics to a property of those resources. Since `records` is mutated and not replaced, it doesn't re-filter or re-sort with the basics. I added `isLoading` here as a workaround

src/fhir/basic.ts
melocule: 1. Alternative would be doing 1 query for all the basics of all the encounters, and then iterating through the result to map the basics to which encounter they belong to. I thought this would make unnecessary work but is it preferred over making multiple queries?
2. (waiting for backfill)

src/components/content/patients-adt/patients-adt-alerts-table.tsx
melocule: Update: caching wouldn't be very helpful and wasn't working, went back to using just an async function

WesleyKapow (dismissed):

WesleyKapow (approved):

PR #1320, opened 2023-10-16
Sort patient search results

luthfib:
Sort the patient search results ascending. Still have to discuss the odd scenario where search is technically both by first and last name.


WesleyKapow (approved):
LGTM

PR #1321, opened 2023-10-17
Use FQS to get patient list

eric-gunther:
Let's start using FQS to get the patient list. Here's an example query that will work in sandbox. If you want to get the next page then pass in the `endCursor` value to the `after` parameter.
```query Patients {
  PatientConnection (builderID: "2c4ff9df-7f87-4253-b2a7-0e8f4944131e", first: 3, after: "") {
    pageInfo {
      hasNextPage
      startCursor
      endCursor 
    }
    edges {
      node {
        id
        resourceType
      }
    }
  }
}

### CODE REVIEW

**Comment by melocule** [comment]
Working screenshots
Screenshot 2023-10-17 at 12 09 19 PM
Screenshot 2023-10-17 at 12 09 15 PM
Note that for testing purposes, page size in CTW for this page was set to 5 instead of 20.

**Comment by melocule** [comment]
Need to confirm whether this will resolve the underlying issue, or whether using SORT in our FQS query will cause it to take just as long.

File: src/components/content/patients/patients-table.tsx [hunk 1]
```diff
@@ -50,77 +39,33 @@ export function usePatientsList(pageSize: number, pageOffset: number, searchName
...
               handleRowClick={handleRowClick}
               pageSize={pageSize}
               hidePagination
             >
-              <Pagination
-                setCurrentPage={setCurrentPage}
-                total={total}
+              <SimplePagination
+                setCurrentPage={(p) => {
+                  setCursors([...cursors, data?.pageInfo.endCursor ?? ""]);
```
WesleyKapow, line 64: This confuses me a bit, shouldn't we only add to the cursors array when we are on a new page? What happens if I go next, prev, next? I'll have cursors: `[page1, page2, page1]`?

src/components/content/patients/patients-table.tsx
```diff
@@ -50,77 +39,33 @@ export function usePatientsList(pageSize: number, pageOffset: number, searchName
...
-    const debouncedSearch = useCallback(
-      debounce((value) => {
-        setSearchNameValue(value);
-        setCurrentPage(1);
-        setTotal(0);
-        setPatients([]);
-      }, 100),
-      []
-    );
+    const [cursors, setCursors] = useState<string[]>([""]);
```
WesleyKapow, line 42: I like this saving of cursors approach vs saving patient records/results and it'll benefit from tanstack query's caching.

src/components/content/patients/patients-table.tsx
WesleyKapow, line 1: Just high level comment here, so we're dropping search from this component and will add the search to the page itself in CTW.

Did we also want to add `Last Updated` column to this table due to default sort? And is this the only thing we can sort by from FQS? I saw sortParams listed in the query.

WesleyKapow (dismissed):

src/services/fqs/queries/fragments.ts
```diff
@@ -163,6 +170,11 @@ export const fragmentPatient = gql`
       text
       use
     }
+    managingOrganization {
+      resource {
+        ...Organization
```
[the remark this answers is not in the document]
eric-gunther: @melocule this is a good call-out, I do think this is bombing the other queries right now so I'd make this a must-change.

src/components/content/patients/patients-table.tsx
melocule: You're right, changing this to a map

melocule:
New homepage
https://github.com/zeus-health/ctw-component-library/assets/33408946/352623b3-dd38-4775-9a9c-e2ac980ba42a

src/services/fqs/queries/fragments.ts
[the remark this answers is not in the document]
melocule: Changed, is this correct?

src/components/content/patients/patients-table.tsx
melocule: Added, doesn't look great in Sandbox due to the data always having the same last updated dates. Not sure if timestamps would be a good idea?

mrosata (dismissed):

melocule:
Broke at some point, investigating

mrosata (dismissed):
lgtm

eric-gunther:
@melocule can you make the patient search box wider and can you vet the final UX/flow with product/design?

melocule:
> @melocule can you make the patient search box wider and can you vet the final UX/flow with product/design?

Made the search box wider in CTW and got a thumbs up from Tai!
image

melocule:
Feature flag has been added and tested

https://github.com/zeus-health/ctw-component-library/assets/33408946/8c60b7e5-debc-4f5d-9de9-60acca5dd2e6

eric-gunther (commented):
@melocule it doesn't look like this PR does anything to check the new ctw-patient-fqs feature flag, so if we merge this and then version bump CTW everyone will start querying FQS for their patient list. I was expecting this PR to check the feature flag to dictate whether `usePatientsList` hits ODS or FQS. I also see that the feature flag is turned on for everyone by default in prod, I think you likely want to start with it off for everyone (so they are hitting ODS) and then gradually turn on.

mrosata (approved):

PR #1332, opened 2023-10-23
CTW-1522/PatientConditionsAll - Fix "add" button showing on readOnly mode

mrosata:
- Ensure "add" button doesn't show in drawer or row when `readOnly` is true
- Default onlyAllowAddOutsideConditions = true

### DIFF
--- a/.changeset/bright-boxes-study.md
+++ b/.changeset/bright-boxes-study.md
@@ -1,0 +1,5 @@
+---
+"@zus-health/ctw-component-library": major
+---
+
+By default PatientConditionsAll will set onlyAllowAddOutsideConditions to true. Previously this defaulted to false.
--- a/.changeset/bright-boxes-study.md
+++ b/.changeset/bright-boxes-study.md
@@ -3,5 +3,3 @@
 ---
 
 By default PatientConditionsAll will set onlyAllowAddOutsideConditions to true. Previously this defaulted to false.
-By default PatientConditions* will set readOnly to true. Previously this defaulted to false.
-By default PatientMedications* will set readOnly to true. Previously this defaulted to false.
--- a/.changeset/bright-boxes-study.md
+++ b/.changeset/bright-boxes-study.md
@@ -3,3 +3,5 @@
 ---
 
 By default PatientConditionsAll will set onlyAllowAddOutsideConditions to true. Previously this defaulted to false.
+By default PatientConditions* will set readOnly to true. Previously this defaulted to false.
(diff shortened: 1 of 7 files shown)


luthfib (approved):

eric-gunther (changes requested):
I think the default behavior should actually be `readOnly = true` and `onlyAllowAddOutsideConditions = true`. We should provide defaults that are most sensible for an EHR embedding the ZAP and the easiest v0 for them is to have the ZAP be read-only so that they don't have to worry about how to handle users adding/editing/removing resources.

In many cases an EHR will want to set `readOnly = false` but keep `onlyAllowAddOutsideConditions = true`. Conditions is the only resource that has these capabilities right now, but we'll want to follow suit elsewhere. This will allow users to add conditions through the Zus UI but then everything else must be done in their EHR. This is a reasonable workflow that we've implemented with both Canvas and Elation. It also requires the EHR to either consume Zus hooks or rig up some of the custom write-back code to ensure the data makes it back to the EHR. Only if they've implemented these write-backs for all actions (ie. including edit/delete) should they set `onlyAllowAddOutsideConditions = false`.

Those are my recommendations and if you proceed with them please ensure that you update CTW immediately after, if we let this wait til the next version bump then whoever does it is likely to forget about this.

mrosata:
@eric-gunther 
> Conditions is the only resource that has these capabilities right now, but we'll want to follow suit elsewhere.
Medications already has `readOnly` _(noting for next bullet)_

> I think the default behavior should actually be `readOnly = true`
Idk, ideally the EHRs will be `readOnly=false`, it's just that it takes us time to get them that functionality (in some cases we may not be able to). The ideal state does seem like them being able to turn that to false tho. If we do change this to true by default then Meds will have to change as well. Would that required communication of the breaking change out to the EHRs? I believe ConditionsAll `readOnly` would just be a breaking change to ctw.

That being said, I can't think of a reason why `onlyAllowAddOutsideConditions` should even be an option. So far no one has come forward with a valid use case. (which doesn't mean there isn't one, but if there is it's not an obvious one)

> Those are my recommendations and if you proceed with them please ensure that you update CTW immediately after, if we let this wait til the next version bump then whoever does it is likely to forget about this.
That's the plan, putting this in ZAP and also CTW right away to avoid a mistake... unless we remove `onlyAllowAddOutsideConditions` and then only if `readOnly` changed would we need to update CTW. I don't know if we need to update `readOnly` in this PR though, could PL it.

eric-gunther (approved):

PR #1338, opened 2023-10-24
CTW-1478 ADT Table Document Indicator

melocule:
CTW-1478

image

UPDATE: Removes basics instead of doing the below, and disables row actions for table.
~~Also made changes to how I got the related encounters / binary IDs, to ensure they're loaded up before the table loads.~~


src/components/content/patients-adt/patients-adt-alerts-table.tsx
```diff
@@ -30,14 +33,13 @@ const PAGE_SIZE = 10;
...
-  {
-    upid: string;
-    cwcq_encounter_id: string;
-    binary_id: string;
-  }
->;
+type RelatedEncounter = {
+  upid: string;
+  cwcq_encounter_id: string;
+  binary_id: string;
```
luthfib, line 39: Don't use snake case for types

luthfib (changes requested):
Move the columns into a separate file and move the helpers out into a separate file. This file in general is a big too big

src/components/content/patients-adt/patients-adt-alerts-table.tsx
```diff
@@ -230,3 +193,87 @@ const PatientColumn = ({ patient }: PatientColumnProps) => (
...
+  getRequestContext: () => Promise<CTWRequestContext>,
+  data: EncounterModel[],
+  encounterAndNotesData?: RelatedEncounterMap
+) {
+  const encountersWithBasics = await mapBasicsOf(getRequestContext, data);
+  if (encounterAndNotesData) {
+    return mapEncountersAndNotes(encountersWithBasics, encounterAndNotesData, getRequestContext);
+  }
+  return data;
+}
```
luthfib, line 279: A high level comment on what you are trying to achieve would be nice here

luthfib (approved):

PR #1343, opened 2023-10-25
CTW-1497/Support feature flag ctw-patient-history-form

mrosata:
Support feature flag ctw-patient-history-form. This will hide the request records button if this flag is disabled. It is enabled for everyone currently.

Tested this running locally by toggling the `ctw-patient-history-form` feature flag in unleash to be disabled for that builder.

### DIFF
--- a/src/components/content/patient-history/patient-history-action.tsx
+++ b/src/components/content/patient-history/patient-history-action.tsx
@@ -1,5 +1,6 @@
 import { PatientHistoryLastRetrieved } from "./patient-history-last-retrieved";
 import { RequestRecordsButton } from "./request-records-button";
+import { useFeatureToggle } from "@/hooks/use-feature-toggle";
 
 export type PatientHistoryActionProps = {
   hideRequestRecords?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/content/patient-history/patient-history-action.tsx
```diff
@@ -9,15 +10,19 @@ export type PatientHistoryActionProps = {
...
-    </div>
-    <div>
-      {!hideRequestRecords && (
-        <RequestRecordsButton includePatientDemographicsForm={includePatientDemographicsForm} />
-      )}
-    </div>
-  </>
-);
+}: PatientHistoryActionProps) => {
+  const hideRequestRecordsFormToggle = !useFeatureToggle("ctw-patient-history-form").enabled;
```
eric-gunther, line 14: This seems like it hides the request records form when the toggle is enabled, not when it's disabled like the changeset and description suggest.

Think it's very confusing to use a feature flag to hide something, making me further second guess using a feature flag for this. May be a bit clearer if the name of the feature flag is something like `ctw-hide-request-records`.

eric-gunther (changes requested):

src/components/content/patient-history/patient-history-action.tsx
mrosata: The feature flag shows it, the variable hides it, the `!` negates the feature toggle.

eric-gunther: Facepalm, I missed the `!` before `useFeatureToggle`.

mrosata:
> @mrosata in the interest of following with soon incoming PR guidelines, can you put in the description how you tested this and how you will confirm it works after deploying to prod?


@eric-gunther added this to description "Tested this running locally by toggling the ctw-patient-history-form feature flag in unleash to be disabled for that builder."

eric-gunther (approved):
Confirmed that the feature flag is on for all users in production.

Please also update any spots in standalone app where this feature flag will be applicable.

mrosata:
> Confirmed that the feature flag is on for all users in production.
> 
> Please also update any spots in standalone app where this feature flag will be applicable.

@eric-gunther The [PR for that](https://github.com/zeus-health/ctw/pull/903) is already in draft, wanted to merge them in together so the ctw-cl upgrade and ctw changes could go in at the same time.

PR #1348, opened 2023-10-26
Bump dependencies

WesleyKapow:


melocule (approved):
Looks good, do we forgo changesets if it's just version bumps?

PR #1356, opened 2023-11-01
CTW-1531 Support for passing in patient ID by itself

eric-gunther:
Jira Ticket Number: [CTW-1531](https://zeushealth.atlassian.net/browse/CTW-1531)

# What does this PR accomplish?

Adds a `PatientResourceID` prop to the `PatientProvider` so that you can load the ZAP by passing the Zus patient FHIR resource ID.

This PR does not refactor our patient query to run off FQS. There is a separate ticket in progress for that right now, so leaving that out of scope for this specific effort.

# How did you test it?

Tested locally passing `patientResourceID` in lieu of `patientID`/`systemURL`. See video below.

https://github.com/zeus-health/ctw-component-library/assets/97455910/2d90886b-526a-4adb-8fb2-9686703ae440

# How will you know that this is working after deployment?

An EHR customer will be the first ones using this, and will be doing so in sandbox, so I will verify with them once they are ready.

# Any outstanding items or follow ups with Jira tickets?

After this is merged I will update the Readme docs with guidance on this.

[CTW-1531]: https://zeushealth.atlassian.net/browse/CTW-1531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/patient-history.ts
```diff
@@ -12,10 +12,10 @@ export type PatientHistoryResponseError = {
 
 export const schedulePatientHistory = async (
   requestContext: CTWRequestContext,
-  patientIdentifiers: { systemURL: string; patientID: string },
+  patientIdentifier: { systemURL?: string; patientID?: string },
   resultData: { id?: string; npi: string; role: string; name: string }
 ) => {
-  const { systemURL, patientID } = patientIdentifiers;
+  const { systemURL, patientID } = patientIdentifier;
```
WesleyKapow, line 18: Does the below usage for `ehr-data-integration-proxy` make sense if these values are undefined?

WesleyKapow (approved):
LGTM, there's some akido failure though o_O

src/api/patient-history.ts
eric-gunther: Good eye. Had to push a change to make the proxy endpoint work if these are undefined (https://github.com/zeus-health/ehr-data-integration/pull/452). Tested this locally against sandbox and it worked.

src/api/patient-history.ts
```diff
@@ -62,7 +62,7 @@ export const schedulePatientHistory = async (
     return await response.json();
   } catch (e) {
     const err = e as Error;
-    const errorMessage = `Error scheduling patient history job with patientID of ${patientID}: ${err.message}`;
+    const errorMessage = `Error requesting patient history.`;
```
eric-gunther, line 65: No need to show stuff like patient ID to an end user. We don't get particularly helpful stuff back in the error here so just going somewhat generic for now, feels like a slight improvement.

WesleyKapow: Original thinking was so they could then relay the error back to us. But they'd know what patient they are doing this for so I suppose you're right, no need!

WesleyKapow (approved):
LGTM

PR #1359, opened 2023-11-03
CTW-1564 Matched patients component

eric-gunther:
Jira Ticket Number:

Adds a `MatchedPatients` component that shows the demographics for all patient records that match your patient but exist in other builders. This will help users identify things like emails and phone numbers that came from other sources, as well as identify all the different sources from which we got data for a patient.

## How did you test it?

Ran locally. When I put this in standalone I will add a playwright test.

image

Will test manually.

## Any outstanding items or follow ups with Jira tickets?

Needs a follow-up PR to add to Remix and Next.js standalone app.


src/fhir/patient-helper.ts
```diff
@@ -1,12 +1,85 @@
...
+            !p.node.meta?.tag?.some(
+              (t) => t.system === SYSTEM_ZUS_UPI_RECORD_TYPE && t.code === "universal"
+            )
+        )
+        .map((x) => new PatientModel(x.node));
+    } catch (e) {
+      throw new Error(`Failed fetching patients: ${e}`);
+    }
+  };
+}
```
luthfib, line 54: This is slightly weird, in that it works because of where we end up using it, and it the function args arent really clear. We can remove the following line: 

```return async (requestContext: CTWRequestContext, patient: PatientModel) => {``` and the

And instead move it over to above to do:

```withTimerMetric(getPatientsForUPIDFQS, "req.timing.matched_patients")```

They will get passed implicitly, but now you can also add the args to `getPatientsForUPIDFQS()`.

So func definition would look something like:

`getPatientsForUPIDFQS(requestContext: CTWRequestContext, patient: PatientModel)`. 

That way if we use it elsewhere it's clear what args it needs in order to function

luthfib (commented):
Looking good, left some comments

luthfib (approved):

PR #1374, opened 2023-11-13
Ctw 1228/confirmation pattern

luthfib:
Jira Ticket Number: CTW-1228

This pr adds a toast component with three states, success, notify, and error. 

## Video

https://github.com/zeus-health/ctw-component-library/assets/16874444/902a3d65-da5f-4f33-bc38-531a280e3d88

## How did you test it?

Tested it by trying out the calling it in a few areas.  

## Usage in drawer

To use in drawer to get it show on right side you can do something like:

```
 notify({
            type: "success",
            title: "Success",
            body: "Record archived",
            containerId: APP_TOAST_CONTAINER_DRAWER_ID,
            options: drawerToastStyles,
          });
```


src/components/core/providers/theme/theme-provider.tsx
WesleyKapow, line 1: You're piggybacking off of theme provider when I think we should have a `ToastProvider` (like we do for modal and drawers)

src/components/core/providers/theme/theme-provider.tsx
```diff
@@ -1,10 +1,14 @@
...
+import { ToastContainer, Zoom } from "react-toastify";
 import { ThemeContext } from "./context";
 import { IFrameTheme } from "@/components/content/zus-aggregated-profile/zus-aggregated-profile-iframe";
 import i18next, { Locals } from "@/i18n";
 import { DefaultTheme, EmptyTailwindCSSVars, mapToCSSVar, Theme } from "@/styles/tailwind.theme";
 import { merge } from "@/utils/nodash";
 import "../../main.scss";
+import "../../toast.scss";
+
+import "react-toastify/dist/ReactToastify.css";
```
WesleyKapow, line 11: If you switch the order of these, you may not need the `!important` overrides.

src/components/core/toast.scss
```diff
@@ -0,0 +1,39 @@
+.ctw-toast-override {
+  --toastify-color-info: #4b5563 !important;
+  --toastify-z-index: 10001 !important; /* one higher than the drawer */
```
WesleyKapow, line 3: Modal's are even higher at `10100` (from `modal.tsx`). We only use modal for confirm delete prompt and for the CCDA modal.

```suggestion
  --toastify-z-index: 10101 !important; /* one higher than modals which are above drawers */
```

src/components/core/toast.scss
```diff
@@ -0,0 +1,39 @@
...
+  color: #e5e7eb !important;
+}
+
+.Toastify__toast {
+  border-radius: 8px !important;
+  padding: 16px !important;
+}
+
+.Toastify__toast-body {
+  padding: 0px 30px 0px 0px !important;
```
WesleyKapow, line 16: nit, drop units when specifying 0

```suggestion
  padding: 0 30px 0 0 !important;
```

src/components/core/toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+  toast.info(
+    <div>
+      <div className="ctw-font-medium">{title}</div>
+      <div className="ctw-font-normal">{body}</div>
+    </div>,
+    {
+      ...options,
+      type: "info",
+      icon: false,
+      style: { backgroundColor: "#4b5563", ...options?.style },
```
WesleyKapow, line 23: Not a fan of these hardcoded values. Maybe use CSS var instead?

src/components/core/toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+          "ctw-toast-grid-with-icon-body-text": title,
+        })}
+      >
+        {body}
+      </div>
+    </div>,
+    {
+      ...options,
+      icon: false,
+      style: { backgroundColor: "#FEE2E2", zIndex: 99999999, ...options?.style },
```
WesleyKapow, line 43: Why bump zIndex here?

src/components/core/toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+      closeButton: <FontAwesomeIcon icon={faXmark} className="ctw-h-4" color="#991B1B" />,
+    }
+  );
+
+export const notifySuccess = ({ title, body, options }: ToastProps) =>
+  toast.success(
+    <div className="ctw-toast-grid-with-icon">
+      <FontAwesomeIcon icon={faCircleCheck} className="ctw-h-5" color="#10B981" />
+      {title && <div className="ctw-font-medium">{title}</div>}
```
WesleyKapow, line 52: This is the only variant where you guard if title is present...

src/components/core/toast.tsx
WesleyKapow, line 1: Maybe these helpers can have an optional `fromDrawer: boolean` that "does the right thing". Also, interestingly, the ticket says they should be left-aligned within the drawer. We can ask about that if it is hard and go with right-aligned. The figma also makes it look like the toast itself is full-width minus some padding of the drawer, maybe that's the preference?

WesleyKapow (changes requested):

src/components/core/toast.tsx
luthfib: Yeah, perhaps. I did notice that as well. That would mean the toast would need to be aware of the drawer, which then of itself would be quite tricky to calculate.

src/components/core/toast.tsx
luthfib: Wasn't a fan either, good idea.

src/components/core/toast.tsx
WesleyKapow: Drawer has hardcoded width I think.. not responsive? May also be able to put a toast container or provider within the drawer itself.

luthfib: Yeah not responsive. It has a fixed with, I'll see if putting the provider in the drawer makes diff...

src/components/core/variables.scss
```diff
@@ -0,0 +1,4 @@
+:root {
+  --ctw-info-background: #4b5563;
+  --ctw-error-background: #fee2e2;
```
WesleyKapow, line 3: This is error-light and the above could be info-dark. I think you could use `className` prop instead of `style` to set the background color.

src/components/core/toast.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div className="ctw-toast-grid-with-icon">
+      <FontAwesomeIcon icon={faCircleCheck} className="ctw-h-5 ctw-text-success-main" />
+      {title && <div className="ctw-font-medium">{title}</div>}
+      <div className={cx(`ctw-font-normal`, { "ctw-toast-grid-with-icon-body-text": title })}>
+        {body}
+      </div>
+    </div>,
+    {
+      ...options,
+      className: "toast-test",
```
WesleyKapow, line 59: `toast-test`?

src/components/core/toast.tsx
WesleyKapow: Also not guarding in `notifyInfo`. You could modify your use of `ToastProps` to make `title` required. What's the contract we want here? Ticket says primary is required and secondary text is optional.

src/components/core/toast.tsx
```diff
@@ -8,9 +8,39 @@ export type ToastProps = {
...
+  containerId?: string;
+  type: "success" | "info" | "error";
 };
 
-export const notifyInfo = ({ title, body, options }: ToastProps) =>
+export const APP_TOAST_CONTAINER_ID = "ctw-toast-container";
+export const APP_TOAST_CONTAINER_DRAWER_ID = "ctw-toast-container-drawer";
+
+export const drawerToastStyles = { style: { bottom: "40px" } };
+
```
WesleyKapow, line 19: May want to setup a `MyToastProps`, that way `notify` can take that. As it exists right now, users could pass in other things from `ToastProps` (outside of those you pull out) and those props would just get dropped on the floor.

src/components/core/toast.tsx
```diff
@@ -8,9 +8,39 @@ export type ToastProps = {
...
+export const notify = ({
```
WesleyKapow, line 20: May want to add a `notifyFromDrawer` helper to obfuscate the container id part as that is an implementation detail.

WesleyKapow (approved):

WesleyKapow (approved):

PR #1380, opened 2023-11-16
Upgrade typescript

luthfib:

This pr updates Typescript and some other packages that needed to be upgraded in order for typescript 5 to work. 

## How did you test it?

I tested it locally to ensure we could still properly do typescript type checking and building.


melocule (approved):
Looks good, works

PR #1383, opened 2023-11-17
CTW-1462 Refactor Basic queries to use FQS

eric-gunther:
Jira Ticket Number: CTW-1462

We will no longer query ODS to get Basic resources (which we had actually been doing to get _all_ Basic resources for a builder, regardless of patient scope, so pretty unsustainable). Instead our queries to get clinical resources will do reverse references to get Basic resources that reference each resource.

There are a lot of files touched here, but it's mainly due to refactoring some of the query keys constants and the need to pass the query keys down the component hierarchy so that we can invalidate the queries when users read/dismiss records. I'm not feeling great about this query invalidation though as it causes us to refetch conditions/meds/etc every time a user reads or dismisses a record. This is because previously we were just invalidating the general Basic query, but now fetching Basic resources happens as part of each resource's FQS query.

## How did you test it?

Tested locally, see video:

https://github.com/zeus-health/ctw-component-library/assets/97455910/f1265eeb-6ac8-4e83-aea8-7ab079d63ced

Will test manually in sandbox. Will login to prod and verify the queries running against FQS.

## Any outstanding items or follow ups with Jira tickets?

Follow-up PR in CTW repo.


src/fhir/allergies.ts
```diff
@@ -8,15 +7,12 @@ import { AllergyGraphqlResponse, allergyQuery } from "@/services/fqs/queries/all
 import { QUERY_KEY_PATIENT_ALLERGIES } from "@/utils/query-keys";
 import { Telemetry, withTimerMetric } from "@/utils/telemetry";
 
-export function usePatientAllergies(enabled = true) {
-  const patientAllergiesQuery = useQueryWithPatient(
+export function usePatientAllergies() {
+  return useQueryWithPatient(
```
mrosata, line 11: Why are we removing the `enabled=true` here?

eric-gunther: Misread my merge conflicts, added back

src/components/content/hooks/use-toggle-read.ts
```diff
@@ -34,7 +33,7 @@ export function useToggleRead(): UseToggleReadResult {
     }
     setIsLoading(true);
     await toggleRead(model, await getRequestContext());
-    await queryClient.invalidateQueries([QUERY_KEY_BASIC]);
+    await queryClient.invalidateQueries([queryToInvalidate]);
```
luthfib, line 36: Essentially you are doing to an optimization here so that we don't re-run unnecessary queries?

src/fhir/models/fhir-model.ts
```diff
@@ -125,32 +125,27 @@ export abstract class FHIRModel<T extends fhir4.Resource> {
...
 
   // Returns the latest Basic resource that has a code that matches
   // one of profileActions.
-  // This way we get the latest action for either "archive" or "unarchive".
   // In practice, there should only be a single Basic resource, but
   // we need to handle the case where there may be several.
-  getLatestBasicResourceByActions(profileActions: string[]): Basic | undefined {
-    const ordered = orderBy(this.revIncludes, "meta.lastUpdated", "desc") as Resource[];
-    return find(ordered, (resource) => {
-      if (resource.resourceType === "Basic") {
```
luthfib, line 144: Do we not need to this anymore because we fetch basics as part of the resource type so we know that are they basics?

luthfib (approved):
Looks good. I left two questions

luthfib:
Don't forget to add a changeset

src/fhir/models/fhir-model.ts
eric-gunther: Yeah, previously `revIncludes` was `fhir4.Resource`, so it allowed more than `Basic`, but practically only ever included `Basic`. Now the field is named `basics` and is exclusively `Basic` resources.

src/components/content/hooks/use-toggle-read.ts
eric-gunther: There is no `Basic` query anymore, read/dismiss info is returned as `Basic` resources on the FQS queries to get the data for each component, so we need to invalidate those upon read/dismiss now.

luthfib (approved):

PR #1396, opened 2023-11-22
[Bug] Fix Patient Search when patients have the same name

Stephen-Tsen:
Jira Ticket Number: [CTW-1620](https://zeushealth.atlassian.net/browse/CTW-1620)

This PR fixes a bug in patient search when there are multiple patients with the same name. Previously the `ComboBoxOption`'s value was the label, which comprised the patient's full name. When an option was selected, the `ComboBoxField` would use the label to look up the first occurrence in the results, so if there are 2 or more patients with the same full name, it would always return the first patient.

Now the `ComboBoxOption`'s value is the entire option so we can just return the option's value to the `onCustomSelectChange` callback.

## How did you test it?

Manually tested in the demo app.
Manually tested with relative dependancies in standalone.

Manually test in all environments

[CTW-1620]: https://zeushealth.atlassian.net/browse/CTW-1620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/core/form/combobox-field.tsx
+++ b/src/components/core/form/combobox-field.tsx
@@ -67,10 +67,10 @@ export const ComboboxField = <T,>({
   }, [onSearchChange]);
 
   const onSelectChange = (e: unknown) => {
-    const { label, value } = e as ComboxboxFieldOption;
-    setSearchTerm(label);
-    setInputValue(value);
-    onCustomSelectChange?.(value);
+    const option = e as ComboxboxFieldOption;
+    setSearchTerm(option.label);
+    setInputValue(option.value);
+    onCustomSelectChange?.(option);
   };
 
   return (
--- a/src/components/core/form/combobox-field.tsx
+++ b/src/components/core/form/combobox-field.tsx
@@ -15,8 +15,8 @@ export type ComboboxFieldProps<T> = {
   options: ComboxboxFieldOption[];
   isLoading: boolean;
   name: string;
-  defaultValue: T;
-  defaultSearchTerm: string;
+  defaultValue?: T;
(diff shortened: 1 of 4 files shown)


mrosata (dismissed):
makes sense

WesleyKapow:
@Stephen-Tsen this may want to go into CTW repo instead!

src/components/core/form/combobox-field.tsx
```diff
@@ -66,11 +66,11 @@ export const ComboboxField = <T,>({
...
-  const onSelectChange = (eventValue: string) => {
-    const currentItem = options.filter((item) => item.label === eventValue)[0];
-    setInputValue(currentItem.value);
-    setSearchTerm(eventValue);
-    onCustomSelectChange?.(currentItem);
+  const onSelectChange = (e: unknown) => {
+    const option = e as ComboxboxFieldOption;
+    setSearchTerm(option.label);
+    setInputValue(option.value);
+    onCustomSelectChange?.(option);
```
Stephen-Tsen, line 73: Call the callback with the entire option object for backwards compatibility. 

The plan is to apply this same fix to the next app/monorepo, and in there I will change this line to call the callback with just the option's value.

Stephen-Tsen:
> @Stephen-Tsen this may want to go into CTW repo instead!

I'll apply a slightly modified version of this to the monorepo

mrosata (dismissed):

mrosata (approved):

